PR #18, opened 2026-07-04
[gpu-hotfix1] Cache improvements have been made (6x performance gain)

par274:
* RAM > GPU > RAM chaining fixed.
* Fixing pipeline cache state.

Performance gain;

Dreaming Sarah: 14 FPS to 70 FPS


par274:
resim

PR #29, opened 2026-07-06
[vulkan] enable required PhysicalDeviceFeatures for translated shaders

foufouadi:
- Enable VertexPipelineStoresAndAtomics/FragmentStoresAndAtomics: fixes vkCreateGraphicsPipelines() rejecting guestBuffers storage descriptor as NonWritable in vertex/fragment stages.
- Enable ShaderInt64: fixes vkCreateShaderModule() rejecting SPIR-V using 64-bit integer capability.
- Query GetPhysicalDeviceFeatures first and only enable what the GPU actually reports as supported, with a warning fallback otherwise.

Also adds optional Vulkan Validation Layers (SHARPEMU_VK_VALIDATION=1) to surface these VUID errors during development instead of silent VK_ERROR_DEVICE_LOST.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,7 +2,6 @@
 Copyright (C) 2026 SharpEmu Emulator Project
 SPDX-License-Identifier: GPL-2.0-or-later
 -->
-
 <Project>
   <PropertyGroup>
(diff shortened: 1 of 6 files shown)


par274:
These changes make sense and were necessary, especially for enabling validation layers during development. May need a bit more testing, but everything worked fine on my GPU, so this looks good to merge.

PR #35, opened 2026-07-10
[gui] Add Avalonia desktop frontend

YkPhysics:
## What

Adds **SharpEmu.GUI**, a modern dark-themed desktop frontend (Avalonia 11.3) that drives the SharpEmu CLI as a child process.

- **Game library** â€“ add folders that are scanned recursively for `eboot.bin`, with search and a game count; folders and options persist to `%APPDATA%/SharpEmu/gui-settings.json`.
- **Launch options** â€“ CPU engine, log level, import trace limit, and strict dynlib resolution, mapped to the existing CLI flags.
- **Live console** â€“ streams emulator stdout/stderr with severity color-coding, a bounded 4000-line buffer, auto-scroll, copy, and clear.
- **Status reporting** â€“ running indicator and exit codes translated to their CLI meaning.

## Why

The emulator currently has no frontend; picking games and reading logs requires a terminal. This gives it a usable launcher without touching the emulator core.

## Review notes

- **`EmulatorProcess.cs` is the interesting part.** The CLI normally relaunches itself into a mitigated child (CET/CFG off) whose console output cannot flow through redirected pipes, so a plain `Process.Start` frontend would lose all emulator logs and spawn a stray console window. The GUI instead performs the mitigated `CreateProcessW` itself (same policy bits as the CLI) with inheritable pipes, sets `SHARPEMU_DISABLE_MITIGATION_RELAUNCH=1` to suppress the CLI-side relaunch, ties the child to a kill-on-close job object, and falls back to an unmitigated launch on Windows builds that reject the policy bits (mirroring `TryRunMitigatedChild`'s silent fallback).
- Console auto-scroll is deferred and applied over a few timer ticks; a synchronous `ScrollIntoView` during rapid appends crashes Avalonia's virtualizing panel with "Invalid Arrange rectangle". Verified with a 150k-line stress run.
- `Tmds.DBus.Protocol` (transitive of `Avalonia.Desktop`) is pinned to 0.21.3 to fix GHSA-xrw6-gwf8-vvr9; the GUI project builds with zero warnings.
- Known limitation: the emulator writes most of its output via direct `Console.Write*` (456 call sites), so the log-level option only filters lines routed through `SharpEmuLog`. Migrating those writes is a candidate follow-up.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,4 +1,4 @@
-<!--
+﻿<!--
 Copyright (C) 2026 SharpEmu Emulator Project
 SPDX-License-Identifier: GPL-2.0-or-later
 -->
@@ -7,10 +7,16 @@ SPDX-License-Identifier: GPL-2.0-or-later
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
   </PropertyGroup>
   <ItemGroup>
+    <PackageVersion Include="Avalonia" Version="11.3.18" />
+    <PackageVersion Include="Avalonia.Desktop" Version="11.3.18" />
+    <PackageVersion Include="Avalonia.Fonts.Inter" Version="11.3.18" />
+    <PackageVersion Include="Avalonia.Themes.Fluent" Version="11.3.18" />
     <PackageVersion Include="Iced" Version="1.21.0" />
     <PackageVersion Include="Silk.NET.Vulkan" Version="2.23.0" />
     <PackageVersion Include="Silk.NET.Vulkan.Extensions.EXT" Version="2.23.0" />
     <PackageVersion Include="Silk.NET.Vulkan.Extensions.KHR" Version="2.23.0" />
     <PackageVersion Include="Silk.NET.Windowing" Version="2.23.0" />
+    <!-- Transitive of Avalonia.Desktop; pinned to fix GHSA-xrw6-gwf8-vvr9 -->
+    <PackageVersion Include="Tmds.DBus.Protocol" Version="0.21.3" />
   </ItemGroup>
 </Project>
--- a/SharpEmu.slnx
+++ b/SharpEmu.slnx
@@ -7,6 +7,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
   <Folder Name="/src/">
     <Project Path="src/SharpEmu.CLI/SharpEmu.CLI.csproj" />
     <Project Path="src/SharpEmu.Core/SharpEmu.Core.csproj" />
+    <Project Path="src/SharpEmu.GUI/SharpEmu.GUI.csproj" />
     <Project Path="src/SharpEmu.HLE/SharpEmu.HLE.csproj" />
     <Project Path="src/SharpEmu.Libs/SharpEmu.Libs.csproj" />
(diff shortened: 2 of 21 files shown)


par274:
The user interface (GUI) is ultimately something we want to prioritize, but for now the command-line interface (CLI) remains the primary interface, so the launcher should reflect that.

However, I don't think we're ready to integrate a GUI yet. We don't currently have enough contributors to maintain it in the long term. I'm also a little concerned about the long-term maintenance of this PR. Using Claude is perfectly fine, we're not against AI powered development, ​​but we want contributors to feel comfortable maintaining and improving it once the code is built.

YkPhysics:
image

par274:
Thanks for the screenshot. 

The UI itself looks fine, but it isn't integrated yet. SharpEmu.exe still launches the CLI, so the GUI isn't currently the primary application. That would need to be addressed first.

But the UI itself isn't my main concern.

My concern is whether this can be maintained long-term. Right now the CLI is still our primary interface, and we don't have enough contributors to support a GUI alongside it.

If you'd like this to be considered in the future, we'd also need someone committed to maintaining it as the project evolves.

YkPhysics:
Thanks for the feedback, that makes sense.

This project genuinely interests me, and I'm willing to commit to maintaining the GUI as SharpEmu evolves.

par274:
That's good to hear. You've already put together a solid foundation, which I appreciate. I'd just like to see the current integration issues resolved first. Once those are addressed, it'll be much easier to evaluate the long-term direction.

YkPhysics:
when you say the GUI would need to be the primary application, which of these is closer to what you have in mind?

1. A single SharpEmu.exe that opens the GUI when run normally and behaves as the current CLI when given arguments

2. SharpEmu.exe becomes the GUI and the current CLI ships alongside it under another name (e.g. SharpEmu.Console.exe)

par274:
A single SharpEmu.exe that opens the GUI when run normally and behaves as the current CLI when given arguments.

YkPhysics:
Updated per your feedback: the GUI is no longer a separate application. SharpEmu.exe now opens the frontend when run without arguments and behaves exactly as the current CLI when given any argument.

par274:
I need a few more changes to merge this.

* Why does the console appear at startup? It's not necessary; opening the GUI is enough.
* If you can embed DLLs into the executable, do so. (Only those related to the GUI, not glfw)
* Every game has a param.json file in sce_sys where the game's name, title_id etc. is written. Instead of PPSA title id, you can write the name and a small title id below it.
* And most importantly, right now the emulator is using its own folder for saved games, etc. You've selected `appdata` for the GUI, but for now, it should save all settings to its own folder.

I can merge this PR after making these changes.

YkPhysics:
👍

par274:
The CI is currently failing because the committed lock files are inconsistent with the project files.

The lock files contain the win-x64 runtime identifier, while the corresponding projects no longer declare a runtime identifier. Since CI restores in locked mode, it cannot update them automatically.

Please regenerate all affected packages.lock.json files from the current project configuration and commit the resulting changes. Do not manually edit the lock files.

PR #39, opened 2026-07-10
[fixes] stackalloc warnings, consolidate duplicated methods, minor adjustments in project settings

nogiszd:
## Summary

This PR fixes all `CA2014` ("potential stack overflow - move stackalloc out of loop") warnings and follows up with a refactor that removes copies of guest-memory helper methods duplicated across the codebase, consolidating them as instance methods on `CpuContext`.

## CA2014 fixes

`stackalloc` memory is not released at the end of the loop - it's only reclaimed when method returns. This can lead to a growing stack frame with each iteration.

The offending pattern flagged by Roslyn compiler was:

```cs
Span<byte> current = stackalloc byte[1]
```

Rather than hoisting the allocation out of the loop, the temporary variable was removed - each guest byte is now read directly into the destination array via `bytes.AsSpan(index, 1)`, which points into pre-existing `bytes` array and also drops a per-byte copy. This is because compiler adds variable above the loop automatically while using `.AsSpan()`.

Behavior is still the same, the byte-at-a-time read pattern is preserved on purpose, since it correctly handles strings than terminate just before an unmapped page. This change can be treated as QoL modification.

## Helper consolidation

Same private helpers (`TryReadByte`, `TryReadUInt32`, `TryWriteUInt32`, `TryReadUInt64`, `TryWriteUInt64`, `TryReadNullTerminatedUtf8`) were duplicated across the codebase, each taking `CpuContext` as a parameter. `CpuContext` already exposed `TryReadUInt64`/`TryWriteUInt64`, so the remaining helpers were added there as instance methods and all duplicates removed.

### Deliberately not migrated

- `ShareExports.TryReadNullTerminatedUtf8` - different contract: returns `false` when no NUL terminator is found within `maxLength` (the canonical version returns the truncated string)
- `KernelMemoryCompatExports.TryReadNullTerminatedUtf8` - reads through that file's `TryReadCompat` translation path rather than `ctx.Memory`
- `IVirtualMemory`-based helpers in `SelfLoader`, `SharpEmuRuntime`, `PhysicalVirtualMemory`, and `CpuDispatcher`'s private `TryWriteUInt32/64(ulong, …)` - these don't take `CpuContext` and are out of scope

## Adjustments in `.csproj` and `global.json`

`global.json` previously pinned the .NET SDK with `"rollForward": "disable"`, which required exactly SDK `10.0.103` to be installed - in my case I had a newer version of the SDK, which failed to resolve. This is now relaxed.

`latestFeature` keeps 10.0.103 as the minimum but allows the SDK resolver to roll forward to the latest installed feature band and patch within .NET 10.0.

All five projects now add `CS1591` warning ignore, because it fired for every public type and member without an XML doc comment, generating unnecessary warnings during the build. Suppresing it project-wide keeps build output focused on warnings that indicate real problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,5 @@ packages/
 Thumbs.db
 Desktop.ini
 ehthumbs.db
+
+.vs/
--- a/SharpEmu.slnx
+++ b/SharpEmu.slnx
@@ -7,9 +7,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
   <Folder Name="/src/">
     <Project Path="src/SharpEmu.CLI/SharpEmu.CLI.csproj" />
     <Project Path="src/SharpEmu.Core/SharpEmu.Core.csproj" />
-    <Project Path="src/SharpEmu.GUI/SharpEmu.GUI.csproj">
-      <Build Solution="Debug|*" Project="false" />
-    </Project>
+    <Project Path="src/SharpEmu.GUI/SharpEmu.GUI.csproj" />
     <Project Path="src/SharpEmu.HLE/SharpEmu.HLE.csproj" />
     <Project Path="src/SharpEmu.Libs/SharpEmu.Libs.csproj" />
     <Project Path="src/SharpEmu.Logging/SharpEmu.Logging.csproj" />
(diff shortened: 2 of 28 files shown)


nogiszd:
Created fresh branch from `main` and cherry-picked my own changes, as requested.

PR #41, opened 2026-07-10
[refactor] move duplicate methods into CpuContext

nogiszd:
## Summary

Follow-up to #39, which consolidated the duplicated `byte`/`uint`/`ulong` guest-memory helpers into `CpuContext`. This PR completes the codebase sweep for the remaining duplicated method families - the signed/16-bit accessors and the `SetReturn` return-value plumbing.

## New `CpuContext` internal API

**Guest memory accessors:**
- `TryReadUInt16` / `TryWriteUInt16`
- `TryReadInt32`
- `TryWriteInt32(address, value, bool checkNil = false)` - the optional `checkNil` guard preserves the null-address check that one duplicate (`NpUniversalDataSystemExports`) performed before writing
- `TryWriteInt64`

**Return-value plumbing:**
- `SetReturn(int result, Type? cast = null)` - writes the result to guest `RAX` and returns it; passing `typeof(long)` selects the sign-extending `(ulong)(long)result` conversion that some exports used, keeping both existing behaviors available from one method
- `SetReturn(OrbisGen2Result result)` - typed overload for the kernel-style result codes

## Duplicates removed

Migrated call sites span the AGC/shader pipeline, kernel compat exports (event flag, semaphore, pthread, pthread-extended, runtime, memory), and the service exports (Audio, AvPlayer, AppContent, Fiber, Http/Http2/Ssl/Net/NetCtl, Ngs2, Np*, Pad, PlayGo, SaveData, Share, SystemService, UserService).

### Endianness note: `BitConverter` -> `BinaryPrimitives`

Six of the seven deleted `TryWriteInt32` copies wrote guest memory with `BinaryPrimitives.WriteInt32LittleEndian`. The copy in `KernelMemoryCompatExports` was the odd one out - it used `BitConverter.TryWriteBytes`, which writes in the **host CPU's** native byte order rather than an explicit one.

Since guest memory must always be little-endian (the emulated CPU is x86-64 - Zen2'ish), the canonical `CpuContext.TryWriteInt32` uses the explicit `BinaryPrimitives` little-endian path. This can be treated as an improvement and redundancy elimination.

### DIFF
--- a/src/SharpEmu.HLE/CpuContext.cs
+++ b/src/SharpEmu.HLE/CpuContext.cs
@@ -138,6 +138,51 @@ public sealed class CpuContext(ICpuMemory memory, Generation generation)
         return true;
     }
 
+    public bool TryReadUInt16(ulong address, out ushort value)
+    {
+        Span<byte> bytes = stackalloc byte[sizeof(ushort)];
+        if (!Memory.TryRead(address, bytes))
+        {
+            value = 0;
+            return false;
+        }
+
+        value = BinaryPrimitives.ReadUInt16LittleEndian(bytes);
+        return true;
+    }
+
+    public bool TryWriteUInt16(ulong address, ushort value)
+    {
+        Span<byte> buffer = stackalloc byte[sizeof(ushort)];
+        BinaryPrimitives.WriteUInt16LittleEndian(buffer, value);
+        return Memory.TryWrite(address, buffer);
+    }
+
+    public bool TryReadInt32(ulong address, out int value)
+    {
+        Span<byte> bytes = stackalloc byte[sizeof(int)];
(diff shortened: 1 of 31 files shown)


par274:
Duplicates is okay but BinaryPrimitives is need more testing.

nogiszd:
Alright. To be fair, this particular part of changes is optional and can be easily reverted by introducing a flag, similar to what I did with `TryWriteInt32', allowing both implementations to coexist.

If the current approach turns out not to be suitable, I’ll update it accordingly.

par274:
It seems fine for now, but next time it would be good if you specified which game you tested this in and the optimization results, such as speed.

Of course, code optimization can also be done, but it's important not to disrupt the workflow.

PR #53, opened 2026-07-11
It's also a cross platform OpenAL audio backend

j92580498-max:
Yeah, I know you're currently focusing on Windows, but I think it's better to do this now rather than later. It will also be useful for forks that build on SharpEmu.

At the moment, audio in SharpEmu only works through WinMmAudioPort which means it's Windows only. On Linux and macOS, sceAudioOutOutput simply doesn't work.

The solution is to implement a new audio backend using OpenAL Soft


par274:
There is a Cross Platform PR. After that, you can look at OpenAL.

PR #57, opened 2026-07-11
core: page-aware TryReadUtf8Z and unify exit/_exit handling

MikeSaito:
**1. Optimization of `TryReadUtf8Z`:**
- Implemented page-aware chunked reading (64 bytes) to safely avoid out-of-bounds page faults.
- Switched to stack-allocated memory buffer (`stackalloc Span<byte>`) to completely eliminate heap allocations.
- Fixed `CA2014` warning and added a proper truncation fallback (returns `false` if no null-terminator found within the buffer limit).

**2. Refactoring of `_exit` and `exit`:**
- Unified exit/stub workflows into a single DRY helper `RequestProcessExit`.
- Fixed `_exit` being a no-op; now it correctly requests thread exit.


par274:
Looking good

PR #58, opened 2026-07-11
More logger improvements

j92580498-max:
The logs were hard to read  at startup there were a bunch of repeated lines, and it wasn't clear whether rendering had even been reached. I cleaned them up.

Added a Milestones helper in SharpEmu.Logging:

Once(...) prints a message only once per run, marks it with [this log will only be shown once] to avoid repeated spam.

FrameHeartbeat(...)  logs only at specific frame numbers (10, 60, 300, 600, 900, 1200, 1500, 1800, then every 1800 frames), making it obvious that rendering is still alive.

Reset() clears the state at the beginning of each run.

Also, at the start of Run, bundle information is printed (before image mapping), taken from param.json:

App bundle info:
- Display name: Dreaming Sarah
- Version: 01.000.000
- Title ID: PPSA02929
- Content ID: UP0891-PPSA02929_00-RAGDREAMINGSARAH
- Executable: eboot.bin
- Platform: PlayStation 5

Also added milestones: CPU emulation begins now before the initializers, plus the first flip and frame heartbeats from sceVideoOutSubmitFlip.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -37,8 +37,11 @@ SPDX-License-Identifier: GPL-2.0-or-later
        top of the log. Populated from GitHub Actions environment variables when
        building in CI; empty locally. -->
   <ItemGroup>
+    <AssemblyMetadata Include="SharpEmu.BuildConfiguration" Value="$(Configuration)" />
     <AssemblyMetadata Condition="'$(GITHUB_SHA)' != ''" Include="SharpEmu.BuildSha" Value="$(GITHUB_SHA)" />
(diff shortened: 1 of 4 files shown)


par274:
Revert the version changes in the packages.lock(especially Microsoft.NET.ILLink.Tasks) files, as this can cause errors in CI.

Have you tried milestones in any title? Performance can decrease if logs are heavily loaded by the CPU.

j92580498-max:
I thought about it, and it would probably add unnecessary overhead, so I'd rather keep performance as good as possible. I removed everything and only left the essentials: the log header now shows the emulator version with a direct link to the GitHub repository. I also kept the App Bundle Info section, and that's it

j92580498-max:
I'm done, can you make a pull request

PR #68, opened 2026-07-11
[AGC] Quake rendering progress: WAIT_REG_MEM, draw fixes, VideoOut, and HLE improvements

foufouadi:
Rebased onto upstream 79a7437 (par274/sharpemu).

This PR fixes several AGC, rendering, and HLE compatibility issues. It also resolves the rendering blockage reported in #7 (Quake [PPSA01880]), where the game previously booted but stalled before reaching the rendering stage.

Changes:
- GpuWaitRegistry: DCBs suspended on unsatisfied WAIT_REG_MEM are now re-polled against guest memory on every submit.
  - Fixed standard and 64-bit packet parsing offsets.
  - Applied the comparison mask correctly.
  - Treat PM4 compare function 0 as "always".

- TryReadSubmittedDrawCount:
  - Added support for the 5-dword ItDrawIndex2 format emitted by DcbDrawIndex.
  - Fixes cases where draw calls were silently discarded.

- sceAgcDriverSubmitMultiDcbs:
  - Fixed reversed ABI argument handling (rdi = address array, rsi = dword sizes, rdx = count).

- VideoOut:
  - Added vblank event handling.
  - Implemented sceVideoOutGetFlipStatus support.
  - Registered buffers can now be used as valid flip targets.

- Added HLE compatibility improvements:
  - libc stdio functions (fopen/fread/fseek/ftell/fclose/fgets).
  - Dinkumware _Getpctype ctype table.
  - NpTrophy2 stubs.
  - AMPR PAK sequential-read tracker.
  - MsgDialog lifecycle handling.
  - NGS2 alternate NIDs and dummy handle-object vtable.
  - Guarded memset intrinsic.
  - abort()/strcasecmp null argument recovery.

- Removed investigation-only code:
  - INT3 breakpoints.
  - qfont/mcpp dumps.
  - temporary error candidate traces.
  - unconditional debug logging.

Validation:
- Quake (PPSA01880) now reaches rendering and presents a 1920x1080 guest frame.
- This addresses the previous issue state where the game was stuck before rendering.

Screenshot:

Quake

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ Desktop.ini
 ehthumbs.db
 
 .vs/
+.idea/
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
@@ -860,7 +860,12 @@ public sealed partial class DirectExecutionBackend
 			"vz+pg2zdopI" or // sceKernelGetEventUserData
 			"mJ7aghmgvfc" or // sceKernelGetEventId
 			"23CPPI1tyBY" or // sceKernelGetEventFilter
-			"kwGyyjohI50";   // sceKernelGetEventData
+			"kwGyyjohI50" or // sceKernelGetEventData
+			// _Getpctype is a per-character ctype table lookup; the embedded mcpp preprocessor
+			// re-fetches the table pointer on every isdigit()/isalpha() check, so classifying a
+			// large input legitimately calls it hundreds of thousands of times back-to-back -
+			// real, finite work that would otherwise trip the loop guard.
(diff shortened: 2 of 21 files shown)


par274:
Impressive work! Can you test more title?

par274:
I understand why you intentionally kept `memcpy` exclude, but we need to move it onto the leaf path. Could you fix this issue?

foufouadi:
Thanks @par274, really appreciate it! 

For memcpy, I moved it onto the leaf path by implementing a guarded native intrinsic (rep movsb) in DirectExecutionBackend.cs. This completely skips the heavy C# HLE dispatch overhead. It's noticeably faster now (Quake gets to the guest frame much quicker), but I made sure to keep the memory safety bounds so we don't trigger that old NGS2 crash again.

Let me know if everything looks good to you!

par274:
I think this looks perfectly reasonable

PR #72, opened 2026-07-11
Implement sceKernelNanosleep in libKernel

j92580498-max:
## What this does

Adds the missing `sceKernelNanosleep` export to `libKernel`. We already had `sceKernelUsleep` but nothing for nanosecond-resolution sleeps, so any title calling `sceKernelNanosleep` was hitting an unresolved import.

## Details

- Registered under NID `QvsZxomvUHs` for both `Gen4` and `Gen5`.
- Reads `rqtp` as a guest pointer to a `timespec` (`tv_sec` / `tv_nsec`, two little-endian int64s) instead of treating the registers as raw values.
- Validates the pointer (`EFAULT` on unmapped memory) and the argument range (`EINVAL` when `tv_nsec` is out of `[0, 1e9)` or values are negative).
- Handles the zero-duration case as an immediate success.
- Sleeps for the requested duration and zeroes the optional `rmtp` remaining-time struct on completion.
- Left the optimized `sceKernelUsleep` short-sleep fast path untouched.

## Testing

Full solution builds clean (0 warnings, 0 errors) on the pinned .NET 10 SDK. I also ran the export directly against a mock guest memory to check every branch:

- null `rqtp` -> EINVAL
- unmapped `rqtp` -> EFAULT
- out-of-range `tv_nsec` -> EINVAL
- zero timespec -> success
- `rmtp` zeroed on success
- 50ms request -> returns success and actually sleeps ~50ms

All checks pass.

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelRuntimeCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelRuntimeCompatExports.cs
@@ -15,6 +15,7 @@ namespace SharpEmu.Libs.Kernel;
 public static class KernelRuntimeCompatExports
 {
     private const int Efault = 14;
+    private const int Einval = 22;
     private const ulong TlsErrnoOffset = 0x40;
     private const ulong TlsStackChkGuardBaseOffset = 0x800;
     private const ulong StackChkGuardFieldOffset = 0x10;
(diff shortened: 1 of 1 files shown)


par274:
English please.

PR #74, opened 2026-07-11
[GUI] Console Search & Log Path Selector

GamingChelsea:
- Added an Search Bar in the Console Log Panel
- Added a File Selector, where to store the Log File


par274:
Nice idea. Could you make the search box a bit wider?

GamingChelsea:
@par274 
Do you think it is better? Or should it always have the exact same width? 
No Text:
12-07-2026;01-18;SharpEmu_v0 0 1_·_dev;PvLTCXguce;
With Text:
12-07-2026;01-20;SharpEmu_v0 0 1_·_dev;diJfGfwiLc;

par274:
It's looking good

PR #79, opened 2026-07-12
Libs: add libSceDiscMap HLE exports

j92580498-max:
## What

Adds an HLE implementation of `libSceDiscMap` as a new `DiscMapExports` class in `SharpEmu.Libs`, following the existing `SysAbiExport` conventions (validated pointers, `OrbisGen2Result` codes, optional env-gated tracing).

Disc-installed titles call into `libSceDiscMap` on most file accesses to decide whether a read has to be redirected to the Blu-ray drive. Without these exports the imports resolve to `ORBIS_GEN2_ERROR_NOT_FOUND`, which titles can treat as an I/O failure. Reporting every request as already resident on internal storage keeps reads on the regular file system path.

## Exports

| NID | Export | Behavior |
| --- | --- | --- |
| `lbQKqsERhtE` | `sceDiscMapIsRequestOnHDD` | writes `1` to the result pointer, returns `0` |
| `fJgP+wqifno` | `sceDiscMapUnknownFJgP` | zero-fills the three output pointers, returns `0` |
| `ioKMruft1ek` | `sceDiscMapUnknownIoKM` | zero-fills the three output pointers, returns `0` |

- Null path/output pointers return `DISC_MAP_ERROR_INVALID_ARGUMENT` (`0x81100001`), consistent with the known `0x811000xx` libSceDiscMap error range.
- Unreadable/unwritable guest memory returns `ORBIS_GEN2_ERROR_MEMORY_FAULT`.
- Set `SHARPEMU_LOG_DISCMAP=1` to trace calls (path, offset, size), matching the `SHARPEMU_LOG_PTHREADS` pattern.
- `sceDiscMapIsRequestOnHDD` is already in `scripts/ps5_names.txt`, so Aerolib resolves the name; the two unidentified NIDs keep descriptive `Unknown` export names like the existing `sceKernelUnknown*` entries.

## Source / attribution

Behavior is ported from the `LibDiscMap.cpp` stubs in [Kyty](https://github.com/InoriRus/Kyty) (MIT), rewritten in C# for SharpEmu's export model. shadPS4's HLE discmap work was used as a cross-reference for the argument layout of the two unidentified NIDs (path, offset, nbytes, three out-pointers). This PR was prepared with the help of an AI assistant; flagging that for the maintainers' awareness.

## Testing

- `dotnet build SharpEmu.slnx` — clean, 0 warnings / 0 errors (SDK 10.0.103)
- Verified via a small harness against `ModuleManager.RegisterFromAssembly`: all three NIDs register and dispatch, `sceDiscMapIsRequestOnHDD` writes `1` and returns `0`, null-pointer calls return `0x81100001`, and the triple-output NIDs zero their out-params.

No existing NIDs are touched; the change is purely additive.

### DIFF
--- a/src/SharpEmu.Libs/DiscMap/DiscMapExports.cs
+++ b/src/SharpEmu.Libs/DiscMap/DiscMapExports.cs
@@ -1,0 +1,104 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+using SharpEmu.HLE;
+
+namespace SharpEmu.Libs.DiscMap;
+
+/// <summary>
+/// HLE implementation of libSceDiscMap, ported from Kyty's LibDiscMap.cpp
+/// (https://github.com/InoriRus/Kyty, MIT). Titles installed to internal
+/// storage query these entry points on nearly every file access to decide
+/// whether a read must be redirected to the Blu-ray drive. Reporting every
+/// request as already resident on HDD lets the regular file system path
+/// service the read.
+/// </summary>
+public static class DiscMapExports
+{
+    private const int DiscMapErrorInvalidArgument = unchecked((int)0x81100001);
+    private const int DiscMapErrorLocationNotMapped = unchecked((int)0x81100002);
+    private const int DiscMapErrorFileNotFound = unchecked((int)0x81100003);
+    private const int DiscMapErrorNoBitmapInfo = unchecked((int)0x81100004);
+
+    [SysAbiExport(
+        Nid = "lbQKqsERhtE",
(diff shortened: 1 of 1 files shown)


Nmzik:
@par274 If I remember correctly, this ABI/API (DiscMap) is only available on the PS4, not the PS5

par274:
These may be leftovers on the PS5. The real question is: which game called them?

j92580498-max:
It is actually used on the PlayStation 5. I looked through the PS5-3.20_Libs, and there's a file called libSceDiscMap

par274:
Yes, that's correct. I don't know whether any games actually call them, but they shouldn't hurt the semantics

PR #91, opened 2026-07-12
[GUI] Add detachable console window

Kaotic:
Add a Split button to detach the emulator console into a resizable window.

The detached console keeps search, auto-scroll, copy, and clear controls. The main console hides while detached and restores when the split window closes if it was previously open.


par274:
Looking good

PR #92, opened 2026-07-12
Linux support: guest-thread scheduling, pthread sync, and GC fail-fast fixes

Spooks4576:
Brings the emulator up on Linux and fixes several runtime crashes that
blocked titles from reaching the Vulkan presenter.

### Linux platform support
- Replace the kernel32-based memory backend with an mmap/POSIX path so
  `PhysicalVirtualMemory` works off Windows.
- Route the Win32 host procedure lookups through Linux thunks, and
  short-circuit the kernel32-only native worker off Windows.
- Unix signal handling for guest access faults.

### Guest-thread scheduling and pthread semantics
- Fix mutex identity so an object reachable through aliased addresses
  resolves to one state, and correct rwlock/condvar wake-key handling.
- PlayGo open and assorted HLE export fixes.

### GC fail-fast on guest threads
Guest threads execute managed HLE on a hijacked stack that the GC cannot
walk, so any collection landing mid-export fail-fasts the runtime with a
misleading "attempted to call an UnmanagedCallersOnly method from managed
code" error.

`TryReadWideCString` was the reliable trigger: it allocated its 1 MiB
ceiling up front, putting a 2 MB array on the Large Object Heap for every
`vswprintf` call, and an LOH allocation forces a GC. It now measures the
string first and allocates exactly its length. Type initializers for the
Silk.NET and framework assemblies are also warmed on the host thread, so a
guest thread is never the first to run a `.cctor`.

### Presenter
`vkQueuePresentKHR` is the only throttle on the window loop, so a
`RenderCore` with nothing to present spun the callback flat out, pegging a
core and starving the guest threads it was waiting on. It now yields.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
     "sdk": {
-        "version": "10.0.103",
+        "version": "10.0.301",
         "rollForward": "disable"
     }
 }
--- a/src/SharpEmu.CLI/Program.cs
+++ b/src/SharpEmu.CLI/Program.cs
@@ -90,6 +90,10 @@ internal static partial class Program
 
         Console.Error.WriteLine("[DEBUG] Creating runtime...");
 
+        SharpEmu.HLE.Diagnostics.StallDiagnostics.SyncObjectStateDumper = address =>
+            SharpEmu.Libs.Kernel.KernelPthreadExtendedCompatExports.DumpRwlockStateForStall(address)
+            ?? SharpEmu.Libs.Kernel.KernelPthreadCompatExports.DumpMutexStateForStall(address);
+
         using var runtime = SharpEmuRuntime.CreateDefault(runtimeOptions);
 
         OrbisGen2Result result;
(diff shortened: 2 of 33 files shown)


j92580498-max:
Could you post a screenshot of it working on Linux?

Spooks4576:
> Could you post a screenshot of it working on Linux?

i will do so soon right now this is a step in the right direction towards Linux support 
if a working game actually works on Linux vs windows remains to be seen

par274:
Testing Windows, broke Dreaming Sarah & Demon's Souls;

resim

resim

Spooks4576:
> Testing Windows, broke Dreaming Sarah & Demon's Souls;
> 
> resim resim

yeah im aware my build has issues with this now 
however i tested upstream without my changes and this broke without my changes so unless im testing the wrong thing

Spooks4576:
Build Broke Something will wait on the other Linux Support Branch PR before giving my other changes

PR #98, opened 2026-07-13
Add configurable input mapping UI

DistantMyth:
﻿# Add configurable input mapping UI

## Summary
Add a configurable Input Mapping dialog for keyboard, mouse, and external controller input. Pad reads now pass through a saved input profile that the user can edit at runtime; keyboard/mouse bindings are layered under whatever physical pad (DualSense *or* Xbox/XInput) is connected.

The profile lives per-user (`%APPDATA%\SharpEmu\pad-input-profile.json` on Windows, `~/.config/SharpEmu/pad-input-profile.json` on Linux, `~/Library/Application Support/SharpEmu/pad-input-profile.json` on macOS), not next to the emulator executable — switched per the maintainer's review.

---

### New files
| File | Purpose |
|---|---|
| `src/SharpEmu.GUI/InputMappingWindow.axaml` | The mapping dialog. Uses the shared design language (`Border.card`, `TextBlock.sectionTitle`, `Button.ghost`/`accent`, `Border.mappingRow`); the per-button callout styling (`mapHotspot`) is window-local because it has to sit over the controller art. |
| `src/SharpEmu.GUI/InputMappingWindow.axaml.cs` | Code-behind. Every visible string is sourced through the localization system. |
| `src/SharpEmu.Libs/Pad/PadInputProfile.cs` | Profile data model + enums (`PadLogicalControl`, `PadStickSource`, `PadStickSide`, `PadInputBindingKind`) + `CreateDefault()` factory (WASD/IJKL/Z/X/C/V/Q/E/R/F/Tab defaults). |
| `src/SharpEmu.Libs/Pad/PadInputProfileStore.cs` | On-disk persistence. **Saves under the OS user directory** (see comment at line 24-25). Re-reads only when the file mtime changes, so live edits in the GUI take effect on the next `scePadRead` without an emulator restart. |
| `src/SharpEmu.Libs/Pad/PadMappedInputReader.cs` | Reads keyboard/mouse bindings out of the profile into a `PadState`. Already honours "only sample the keyboard while the emulator window is focused" (via `PlatformPadInputSource.IsInputActive`). |
| `src/SharpEmu.Libs/Pad/PlatformPadInputSource.cs` | Platform abstraction: `IsInputActive` (window-focus gate) + `IsBindingDown` + `ReadMouseDelta`. Windows implementation uses `GetAsyncKeyState`/`GetForegroundWindow`/`GetCursorPos`; non-Windows uses a Null source (input disabled), keeping the build portable. |
| `assets/images/input-controller-vecteezy-7743127-cropped.jpg` · `…-white.png` | Vecteezy DualSense layout art for the mapping window. |
| `LICENSES/LicenseRef-Vecteezy-Free-License.txt` | The Vecteezy license text — REUSE requires this file because the asset uses a custom `LicenseRef-` identifier (REUSE would otherwise report `MissingLicense`). |

### Modified files
| File | Change |
|---|---|
| `src/SharpEmu.Libs/Pad/PadExports.cs` | `WriteNeutralPadData` now reads keyboard/mouse via `PadMappedInputReader.Read(profile)` and merges DualSense + XInput on top (when `EnableExternalController` is set). Both pads contribute when both are plugged in; the controller's axis wins past a 10-unit deadzone. The boot-time controls announcement uses upstream #189's `Interlocked.Exchange` one-shot guard, with this PR's profile-aware `DescribeControls(profile)` message inside (so the printed hint reflects the user's actual saved mappings rather than hardcoded WASD/IJKL defaults). |
| `src/SharpEmu.GUI/MainWindow.axaml` | Added one row to the Options → LAUNCHER card: "Input mapping / Open…" (matches the "Log file path" row pattern). The launch bar itself is unchanged. |
| `src/SharpEmu.GUI/MainWindow.axaml.cs` | Wires `OpenInputMappingButton.Click` to `OpenInputMappingAsync()` (constructs `InputMappingWindow` and shows it via `ShowDialog`). Three localized strings added to `ApplyLocalization()`. |
| `src/SharpEmu.GUI/App.axaml` | Lifted the mapping row style into the shared vocabulary as `Border.mappingRow` (`ElevatedBrush` bg, corner 8, padding `10,8`) so the dialog rows match the rest of the UI. |
| `src/SharpEmu.GUI/SharpEmu.GUI.csproj` | Compiles the six `Pad/*.cs` files directly into the GUI (linked), so the launcher doesn't pull all of `SharpEmu.Libs`. Adds the Vecteezy image as an `AvaloniaResource`. |
| `src/SharpEmu.GUI/Languages/en.json` | Added 53 keys: 3 in `Options.InputMapping.*` and 50 in the `Input.*` namespace. Total now 171 (upstream added another 16 `Env.*`/`About.*` between earlier iterations of this PR — those are upstream's, not mine). |
| `src/SharpEmu.GUI/Languages/tr.json` | Same 53 input-related keys added (Turkish). Total now 155. The 16 upstream `Env.*`/`About.*` keys the upstream PRs added to `en` but not to `tr` fall back to English via `Localization._fallbackStrings`, same as before this PR. |
| `REUSE.toml` | Declares the Vecteezy assets with `SPDX-License-Identifier = "LicenseRef-Vecteezy-Free-License"` so `reuse lint` is happy. |

### Removed files
| File | Reason |
|---|---|
| `assets/images/THIRD_PARTY_NOTICES.md` | Redundant with REUSE — the Vecteezy asset licensing is fully documented in `REUSE.toml` + `LICENSES/LicenseRef-Vecteezy-Free-License.txt` already. Per the maintainer's review. |

---

### Xbox controller art / labels in the Input Mapping window
**Status:** open.

The DualSense-only art and PlayStation-shaped callout labels (Cross/Circle/Square/Triangle, L1/R1/L2/R2, TouchPad/Options) are still in place. A user on an Xbox controller who opens the mapping window still sees DualSense labels.

Reason I haven't touched it: **at the input-source level, Xbox already works** — `XInputReader.Translate` (`XInputReader.cs:178-211`) maps A/B/X/Y, LB/RB/LT/RT, Menu/View, and the D-pad into the same ORBIS pad flags DualSense uses, and `WriteNeutralPadData` merges both XInput and DualSense state every frame, so an Xbox Series / Xbox 360 (or any XInput-compatible) controller drives in-game input with zero configuration. The Input Mapping window itself is only there to remap the **keyboard/mouse side**: when a real Xbox pad is connected and `EnableExternalController` is set, the user's saved keyboard bindings are simply outvoted past the 10-unit stick deadzone. An Xbox user with a working controller has nothing to remap.

Given that, it's not clear which behaviour the maintainer wants:
1. **Dynamic relabel** — when an Xbox pad is detected, swap callouts to A/B/X/Y + LB/RB/LT/RT + Menu/View. No art change.
2. **Art swap** — find a properly-licensed Xbox controller illustration and switch art based on the active pad (adds a third-party asset; needs its own REUSE entry).
3. **Dual labels** — render "Cross / A", "L1 / LB", "Options / Menu" on every callout. No art change, no detection needed.
4. **Decouple** — treat the mapping window as keyboard/mouse-only and drop the controller illustration entirely.

This PR does the smallest correct thing the maintainer explicitly asked for ("save the JSON under the users directory instead") and leaves the labelling question open for a follow-up.

### Translating all other languages
**Status:** intentionally not done.

12 other language files exist now (`ar`, `br`, `de`, `dk`, `es`, `fr`, `it`, `ja`, `ko`, `nl`, `ru` + my `tr`). I only added the 53 new keys to `en` and `tr` (the languages I work in). For the other languages I rely on the `Localization.Get` fallback: when a key is missing from the active language, `Localization.cs:42-43` looks it up in the `_fallbackStrings` table (which is always populated from `en.json`), so users on German/Japanese/etc. see the English Input Mapping strings rather than raw keys like `Input.Save` — no regression. Translators can add the keys whenever they want; there's no rush and I shouldn't be making up translations I'm not qualified to write.

### Removing `LICENSES/LicenseRef-Vecteezy-Free-License.txt`
**Status:** not removed, by design.

REUSE treats custom `LicenseRef-` identifiers differently from standard SPDX identifiers: standard ones (`GPL-2.0-or-later`, `MIT`, `Apache-2.0`) have their text built into REUSE's license list, so no file is required; custom ones (`LicenseRef-*`) **require** the project to ship `LICENSES/LicenseRef-<Name>.txt` containing the license text — otherwise `reuse lint` fails with `MissingLicense`. So the `.txt` is the REUSE-mandated home for the Vecteezy license text, not a redundant duplicate of the markdown I removed. The redundant file was `THIRD_PARTY_NOTICES.md`; this `.txt` is the load-bearing one.

### Touchpad as a mappable "button"
**Status:** kept as-is.

`TouchPad` maps to the `OrbisPadButton.TouchPad` bit (`0x100000`) — exactly the same bit that DualSense's touchpad click and Xbox's **View** button already produce. So assigning a keyboard key to "TOUCH" sends the same ORBIS bit the games already recognise as "the touchpad button is down." This is a binary button mapping only — it does **not** synthesise per-frame touch coordinates, the same way L2/R2 are mapped here as buttons even though they're analog triggers. The mapping surface in this dialog is binary, and that limitation is consistent across all 16 logical controls. Full touchpad-coordinate support would need an analog source (joystick or mouse) feeding a touch surface and is out of scope here.

### `core.autocrlf=true` and `dotnet format whitespace` ENDOFLINE errors
**Status:** non-issue.

Running `dotnet format whitespace --verify-no-changes` against the working tree flags ENDOFLINE on essentially every `.cs` file in the repo, including ones I never touched. This is because `core.autocrlf=true` (Windows git default) inflates `LF` → `CRLF` on checkout. Reusable sanity check: `git show HEAD:<file>` shows the canonical LF blob; whatever the working tree has, the **commit** will store LF as `.editorconfig` requires. The formatter is being misled by the working-tree state, not by anything this PR introduces.

---

## Validation
- `dotnet build src/SharpEmu.GUI/SharpEmu.GUI.csproj` — 0 warnings, 0 errors.
- `dotnet build src/SharpEmu.Libs/SharpEmu.Libs.csproj` — 0 warnings, 0 errors (verifies the `PadExports.cs` conflict resolution integrates cleanly with upstream).
- `dotnet run --project src/SharpEmu.CLI/SharpEmu.CLI.csproj` — manual smoke test: opened Options → LAUNCHER → "Input mapping / Open…", edited keyboard bindings, switched the language to Türkçe and reopened the dialog to verify all strings re-rendered, confirmed the saved file appears at `%APPDATA%\SharpEmu\pad-input-profile.json`.
- Both `en.json` and `tr.json` parse cleanly; same 53 input-related keys are present in each.

---

## Conflict resolutions during rebases
Rebased onto `main` three times as upstream moved. Conflicts needed manual resolution twice:

**Rebase 1 (`#102..#133`):**
- `MainWindow.axaml`/`.axaml.cs` — the upstream GUI redesign replaced the old options-panel toggle with a Library/Options tab switcher. Kept the new design; moved the Input Mapping entry point from a launch-bar button into the Options → LAUNCHER card.

**Rebase 2 (`#133..#189`):**
- `Languages/en.json` — kept both upstream's `Options.About` / `About.*` block and this PR's `Input.*` block.
- `SharpEmu.GUI.csproj` — kept both upstream's `github.png`/`discord.png` Avalonia resources *and* the Vecteezy controller image.
- `PadExports.cs` — kept upstream #189's `Interlocked.Exchange` one-shot log guard, but emit this PR's profile-aware `DescribeControls(profile)` message inside it (so the printed hint reflects the user's actual saved mappings rather than hardcoded WASD/IJKL defaults). All four upstream keyboard helpers (`ReadKeyboardButtons`, `ReadAnalogStick`, `IsKeyDown`, the `user32.dll` P/Invokes) were dropped because `PlatformPadInputSource` already encapsulates the same checks and `PadMappedInputReader` consumes it.


DistantMyth:
## Additional implementation notes

This PR replaces the previous fixed keyboard-control path with a configurable input profile system and adds a GUI for editing those mappings.

### What changed
- Added an `Input Mapping` window to the GUI so users can configure keyboard/mouse bindings and external controller usage from the frontend.
- Routed pad input through a saved `PadInputProfile` instead of hardcoded keyboard checks in `PadExports`.
- Added profile persistence via `pad-input-profile.json`, loaded from the application base directory.
- Added a mapped input reader that merges keyboard/mouse input with DualSense/XInput controller state when external controllers are enabled.
- Added default mappings matching the previous behavior, including WASD for left stick, IJKL for right stick, arrow keys for D-pad, and the existing face/shoulder button keys.
- Added a controller layout reference image to the mapping dialog, with attributed Vecteezy assets and REUSE metadata.

### What was removed or replaced
- Removed the old direct keyboard polling logic from `PadExports`.
- Replaced hardcoded D-pad, stick, face-button, shoulder-button, and Options key handling with profile-driven mappings.
- Moved platform-specific keyboard/mouse polling behind `PlatformPadInputSource`, so the pad layer reads logical bindings rather than raw virtual keys directly.

### Verification
- Confirmed the new mapping dialog opens and the configured controls work locally.
- XML-parsed `InputMappingWindow.axaml`, `MainWindow.axaml`, and `SharpEmu.GUI.csproj`.
- Built successfully with the pinned SDK using:
  `D:\Extralicen\dotnet-sdk-10.0.103-win-x64\dotnet.exe build src\SharpEmu.CLI\SharpEmu.CLI.csproj --no-restore -p:OutputPath=D:\Contribute\sharpemu\artifacts\codex-build\ -p:UsedAvaloniaProducts=`
- The alternate output path was used because a running `SharpEmu.exe` instance was locking the normal `artifacts\bin\Debug\net10.0\win-x64` output directory.

### Notes and follow-up
I am also planning to add Linux and macOS support once #47 is concluded. Mouse-as-stick support is implemented, but it has not yet been validated in a real 3D game because a suitable 3D test title is not ready yet. I plan to test and refine that behavior as soon as one becomes available.

Screenshot:
Input mapping window

DistantMyth:
I added the image from a licence free(attribution required) source, if you want to avoid that completely, I am also open to just generating a similar asset instead.

par274:
Hi, it's looking really good. The only reason I haven't approved the PR yet is that I've been making a lot of GUI changes on my side.

There are a few things I'd like you to change. Since we've moved to an updated UI, it should follow the new design language.

Also, the license information is already covered by REUSE, so there's no need for an additional Markdown file for the assets you added. Their licensing is already documented there, and we'd rather avoid unnecessary files in the repository.

One more thing: it's not just DualSense. A lot of PC users use Xbox Series and Xbox 360 controllers as well, so those should be supported too. The GUI already supports those controllers, and the in-game controls should be consistent with that.

DistantMyth:
@par274 
- **New design language.** Input Mapping now uses the shared `card` / `sectionTitle` / `ghost` / `accent` styles from `App.axaml`, lifts the row style into `App.axaml` as `Border.mappingRow`, and routes every visible string through the localization system. Added 52 keys (`Options.InputMapping.*` + an `Input.*` namespace) to both `Languages/en.json` and `Languages/tr.json`. The launcher-side entry point moved from a standalone "Input" button in the launch bar to a row in the Options → LAUNCHER card ("Input mapping / Open…"), matching the "Log file path" pattern.
- **Removed the redundant Markdown.** `assets/images/THIRD_PARTY_NOTICES.md` is gone — the Vecteezy asset licensing is already covered by `REUSE.toml` + `LICENSES/LicenseRef-Vecteezy-Free-License.txt` (the latter is the REUSE-mandated home for the custom license text; deleting it would make `reuse lint` report `MissingLicense`).

For the third point (Xbox consistency): at the emulator layer Xbox support already works — `XInputReader` translates A/B/X/Y, LB/RB/LT/RT, Menu/View, and the D-pad into the ORBIS pad flags, and `PadExports.WriteNeutralPadData` merges both DualSense and XInput at read time, so an Xbox Series / Xbox 360 (or any XInput-compatible) controller works in-game without any configuration. What's still DualSense-shaped is the **mapping window** itself (its art and per-button callouts use Cross/Circle/Square/Triangle, L1/R1/L2/R2, TouchPad/Options).
I deliberately did not touch that yet because, given that an Xbox user has nothing to remap, it isn't obvious which behaviour you'd like — 
- relabel the callouts dynamically when an Xbox pad is connected (like giving option to map the xbox styled controller to the DualSense)
- option to swap to an Xbox controller illustration
- dual-label "Cross / A"
- treat the mapping window as keyboard/mouse-only.

Would appreciate clarification; happy to do whichever you prefer.
Also, here is the new placement for our Inputs Tab, I also added checkboxs to enable/disable controller and keyboard/mouse input
image
image

par274:
Could you save the JSON file under the users directory instead?

Once you've done that, I can merge it.

DistantMyth:
sure, on it, want me to change anything with the xbox controller thing..?

par274:
No, its okay. You should resolve conflicts

DistantMyth:
Done! ✅ Rebased onto latest `main` and moved the profile JSON as requested.

**Changed:**
- `pad-input-profile.json` now saves to the OS per-user config dir via `Environment.SpecialFolder.ApplicationData` — `%APPDATA%\SharpEmu\` on Windows, `~/.config/SharpEmu/` on Linux, `~/Library/Application Support/SharpEmu/` on macOS. No more writing next to the executable.

**Kept from upstream while resolving conflicts:**
- The `ReadHostInputState()` 1ms sampling cache and `PrimaryUserId = 1000` from #130 — my profile-driven `PadMappedInputReader` now feeds into that cached path instead of replacing it, so the perf work stays intact.
- The `scePadSetTriggerEffect` stub from #131 (cf69647) — untouched, sits alongside the new mapping code.

The old hardcoded `GetAsyncKeyState`/`user32.dll` keyboard code is gone since the mapping system supersedes it. Builds clean with the pinned SDK, 0 warnings. Should be good to merge 🙂

DistantMyth:
Sorry, for the delay btw, just had to make sure to not mess things up since the mentioned commits resulted in some big conflicts

DistantMyth:
@par274 it's good to merge, was just being sure that everything's robust.
and btw, since i have started storing the pad-input-profile.json file to the user directory, you might as well save things like gui-settings.json there, i didn't do that cause i don't know where it might be hardcoded.
i am happy to make another smaller PR for that later

par274:
need rebase

DistantMyth:
﻿Rebased onto the latest `main` (now through #189). Three conflicts needed manual resolution:

- `src/SharpEmu.GUI/Languages/en.json` — kept both the upstream `Options.About` / `About.*` block and this PR''s `Input.*` block. The other 12 languages (ar / br / de / dk / es / fr / it / ja / ko / nl / ru / tr) are unaffected — `Localization` falls back to the English values for any missing key, so the Input Mapping dialog renders correctly in every language without forcing translators to add 52 keys each.
- `src/SharpEmu.GUI/SharpEmu.GUI.csproj` — kept the upstream `github.png` / `discord.png` Avalonia resources and this PR''s `input-controller-vecteezy-7743127-white.png` resource.
- `src/SharpEmu.Libs/Pad/PadExports.cs` — kept upstream #189''s one-shot `Interlocked.Exchange` guard on the boot-time controls announcement (no more per-`scePadOpen` log spam), but emit this PR''s profile-aware `DescribeControls(profile)` message inside the guard so the printed hint reflects the user''s actual saved mappings rather than the hardcoded `WASD / IJKL / Z / X / C / V` defaults.

Build verified: `dotnet build src/SharpEmu.GUI/SharpEmu.GUI.csproj` and `dotnet build src/SharpEmu.Libs/SharpEmu.Libs.csproj` are both green (0 warnings, 0 errors) on the SDK pinned by `global.json`.

DistantMyth:
@par274 done

DistantMyth:
Could I please get feedback...

par274:
You're sending my requests to AI, and it misunderstands them and implements them incorrectly. So this PR has regressed.

We’re not against using AI, but please read the contribution guidelines and at least try to explain things in your own words. Also, please try to understand our feedback in that context.

DistantMyth:
I actually was using a skill for describing all the changes made and another to commit and manage PRs, I do understand why that would seem like complete copy paste and move on.
Yet if the code was not good enough, thanks for pointing out, thanks, really appreciate the time and your effort.

PR #108, opened 2026-07-13
[AGC] Decode gfx10 SOPP hint instructions

Deeptanshuu:
## What this fixes

Follow-up to #103 / #106. The SOPP decode table currently ends at `s_inst_prefetch` (`0x20`), so gfx10 scheduling-hint instructions are unknown to the decoder:

| opcode | instruction | purpose |
|--------|-------------|---------|
| `0x21` | `s_clause` | marks a memory-instruction clause for the hardware scheduler |
| `0x23` | `s_waitcnt_depctr` | fine-grained dependency-counter wait hint |

An unknown SOPP opcode is a hard decode error (`unknown-sopp op=0x21`), which rejects the **entire shader**. `s_clause` in particular is routinely emitted by RDNA compilers ahead of clustered memory loads, so this failure mode is likely common in real shader binaries — I don't own game dumps to quantify that, so regression reports from dump owners are very welcome.

**Scope note (updated per review):** the first revision also added `s_round_mode` (`0x24`) and `s_denorm_mode` (`0x25`). As the review below points out, those two write the shader floating-point MODE state (RDNA2 ISA §6.4), and the emitter's blanket SOPP no-op would have silently ignored their `simm16` payloads — converting an explicit decode failure into a potential silent FP-semantics mismatch. They've been removed and continue to fail decode loudly. Follow-up options are in the review thread.

## Changes

Two entries in the `DecodeSopp` table (`Gen5ShaderTranslator.cs`). +2 lines, nothing else.

**No emitter changes are needed:** `Gen5SpirvTranslator.TryEmitInstruction` already emits every non-branch SOPP instruction as a no-op. For these two that is semantically exact — they are pure scheduler/dependency hints with no effect on computed values.

## Evidence for the opcode numbers

LLVM's AMDGPU backend ([`SOPInstructions.td`](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Target/AMDGPU/SOPInstructions.td)):

```
defm S_CLAUSE          : SOPP_Real_32_gfx10_gfx13<0x021>;
defm S_WAITCNT_DEPCTR  : SOPP_Real_32_gfx10_Renamed_gfx13<0x023, "s_wait_alu">;
```

(gfx11/gfx12 renumbered these instructions; the additions here follow gfx10, consistent with the numbering evidence established in #103/#106.)

## Verification (synthetic, no game dumps used)

Same approach as #103/#106: hand-assembled programs fed through decode and full SPIR-V compilation:

- **Before:** `s_clause 0x1; s_waitcnt_depctr 0; s_endpgm` fails with `error=unknown-sopp op=0x21` — the whole shader is rejected at decode.
- **After:** both decode by name (`SClause:1,SWaitcntDepctr:1,SEndpgm:1`) and the program compiles end-to-end to 1288 bytes of SPIR-V through the existing SOPP no-op emit path.
- **Unchanged:** a program containing `s_round_mode` still fails loudly (`error=unknown-sopp op=0x24`), preserving the explicit-failure behavior for the FP-mode instructions.

Full solution builds with 0 warnings / 0 errors (SDK 10.0.103).

**CI on my fork is green for this branch, including REUSE compliance:** https://github.com/Deeptanshuu/sharpemu/actions/runs/29272654233

## Notes

Developed with AI assistance per the contribution policy; I understand the change and can walk through it in review.

### DIFF
--- a/src/SharpEmu.Libs/Agc/Gen5ShaderTranslator.cs
+++ b/src/SharpEmu.Libs/Agc/Gen5ShaderTranslator.cs
@@ -621,6 +621,10 @@ internal static class Gen5ShaderTranslator
             0x10 => "SSendmsg",
             0x16 => "STtraceData",
             0x20 => "SInstPrefetch",
+            0x21 => "SClause",
+            0x23 => "SWaitcntDepctr",
+            0x24 => "SRoundMode",
+            0x25 => "SDenormMode",
             _ => string.Empty,
         };
 
--- a/src/SharpEmu.Libs/Agc/Gen5ShaderTranslator.cs
+++ b/src/SharpEmu.Libs/Agc/Gen5ShaderTranslator.cs
@@ -623,8 +623,6 @@ internal static class Gen5ShaderTranslator
             0x20 => "SInstPrefetch",
             0x21 => "SClause",
             0x23 => "SWaitcntDepctr",
-            0x24 => "SRoundMode",
-            0x25 => "SDenormMode",
             _ => string.Empty,
         };
 


kostyaff (commented):
I checked the current head (`95a04c0`) independently: the four opcode numbers match LLVM, and locked restore plus the full Release build pass with 0 warnings / 0 errors.

There is one correctness concern with treating all four instructions as equivalent no-ops. `s_clause` and `s_waitcnt_depctr` are scheduling/dependency hints, so dropping them during SPIR-V translation is reasonable. `s_round_mode` and `s_denorm_mode` are different: AMD's RDNA2 ISA documents them as writes to the shader floating-point MODE state, controlling rounding direction and denormal input/output handling ([section 6.4, Round and Denormal Modes](https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna2-shader-instruction-set-architecture.pdf)).

`TryEmitInstruction` currently accepts any remaining `Gen5ShaderEncoding.Sopp` as a no-op. Therefore adding the two mode names makes translation succeed while silently ignoring their `simm16` values. The synthetic end-to-end compile proves that they are accepted, but it does not prove floating-point semantics; non-default mode values can change later shader arithmetic.

Could this PR keep the two true hints (`SClause`, `SWaitcntDepctr`) and defer `SRoundMode` / `SDenormMode` until their semantics are modeled or conservatively validated? That would preserve the small decode-correctness win without converting an explicit decode failure into a silent FP-state mismatch.

Deeptanshuu:
Thanks for the careful review — and for verifying the head independently. You're right, and the line you drew is the correct one: `s_clause`/`s_waitcnt_depctr` only constrain the scheduler, while `s_round_mode`/`s_denorm_mode` write FP MODE state whose `simm16` payload the blanket SOPP no-op would have silently discarded. Converting an explicit decode failure into a silent FP mismatch is exactly the failure class #103 was about, so I've narrowed the PR in 3e8d0a7: it now adds only the two scheduling hints, and a program containing `s_round_mode` still fails loudly (re-verified with a synthetic program: `error=unknown-sopp op=0x24`). PR description updated to match.

For a follow-up (separate PR), I read §6.4 alongside LLVM's `SIDefines.h` — they agree on the encodings:

- `FP_ROUND`: `[1:0]` single / `[3:2]` double+half; `0` = round-to-nearest-even, `1` = +inf, `2` = −inf, `3` = toward zero.
- `FP_DENORM`: `[5:4]` single / `[7:6]` double+half; `0` = flush in+out … `3` = allow in+out.

Proposal for the two mode instructions:

- **`s_round_mode`**: decode it, and at emission treat it as a no-op **only** when `simm16 & 0xF == 0` (nearest-even for all widths — which matches the default rounding the emitted SPIR-V already runs under in Vulkan). Any other value fails translation with a descriptive error naming the requested mode, instead of today's opaque `unknown-sopp`.
- **`s_denorm_mode`**: harder to validate conservatively — the translator declares no SPIR-V denormal execution modes, so denorm behavior is host-defined and no immediate provably "matches". Options: (a) keep failing, but with a descriptive message, or (b) no-op with explicitly documented host-defined denorm semantics. I'd lean (a) given the spirit of this review.

Happy to implement whichever direction you and @par274 prefer.

CI for the narrowed head is green, including REUSE: https://github.com/Deeptanshuu/sharpemu/actions/runs/29272654233

kostyaff (approved):
Re-reviewed the narrowed head `3e8d0a7` against current `main`.

PASS:
- net diff is exactly the two scheduling hints (`SClause` / `SWaitcntDepctr`) in one decoder table;
- `SRoundMode` and `SDenormMode` were removed again, so opcodes `0x24` / `0x25` still fail through `unknown-sopp` instead of silently losing FP MODE state;
- Standards and clarified-spec reviews both have 0 findings;
- SDK 10.0.103 locked restore and full Release build pass with 0 warnings / 0 errors;
- `git diff --check` passes.

This resolves my earlier correctness concern. Thanks for narrowing the scope.

Issue #36, opened 2026-07-10
Offering to contribute: unit-test harness + synthetic ISA corpus for the Gen5 shader translator


bestforsmart-spec:
Hi! I've spent some time reading the Gen5 shader translator pipeline (`AgcExports` capture → `Gen5ShaderTranslator` decode → `Gen5ShaderScalarEvaluator` → `Gen5SpirvTranslator` emit → `VulkanVideoPresenter`) and I'd like to contribute to it regularly. Before opening PRs I want to check the direction with you.

### What I have working already

Branch: https://github.com/bestforsmart-spec/sharpemu/tree/tests/m0-harness

- `SharpEmu.Libs.Tests` (xunit) with a minimal `ICpuMemory` fake, so translator code runs without a live guest.
- First decoder tests: raw **gfx1013** encodings assembled with `llvm-mc` (`v_add_f32_e32`, `s_endpgm`) decode to the expected opcodes. The decoder agrees with LLVM's encoder on these — nice.
- One accessibility change: `Gen5ShaderTranslator.TryDecodeProgram` `private` → `internal` + `InternalsVisibleTo`.
- A CI workflow running `dotnet test` on ubuntu/windows/macos (in a follow-up commit).

Side note: `SharpEmu.Libs` is pure managed code — it builds and tests cleanly on macOS arm64 (.NET 10.0.103), so translator work doesn't need Windows at all. Happy to keep it that way.

### What I'd like to do next (in small PRs)

1. **Synthetic ISA test corpus** — per-instruction cases assembled with `llvm-mc` for gfx1013 (cyan_skillfish — the family open drivers use for this GPU class) and gfx1030, plus full VS/PS pairs from RGA `vk-offline` and RADV/ACO dumps. 100% synthetic, zero console-derived bytes, so the repo stays clean legally.
2. **ISA gaps**: VOP3P packed-fp16 emit (currently opaque → hard failure), missing buffer/image/DS atomics, `v_writelane`.
3. **Scalar evaluator robustness**: bounded interpretation for loops/backward branches in descriptor preambles, per-descriptor failure instead of failing the whole draw.
4. Longer term, toward a first real frame in Demon's Souls: DB_* register parsing + depth attachment, and MRT (CB slots 1–7 are currently dropped at `Gen5SpirvTranslator.cs` export handling).

### Questions

- Are you open to a tests project living in the main solution? Any conventions you want for it?
[...]

PR #111, opened 2026-07-13
[Tools] Synthetic shader dump tool for the Gen5 translator (SPIR-V dump + validation)

Deeptanshuu:
> Opening as a **draft** per CONTRIBUTING's "discuss large changes / open a draft PR first" — this adds developer infrastructure rather than emulator code. Also very relevant to the test-harness discussion in #36.
>
> **Rescoped after review:** this PR now adds only the SPIR-V dump tool. The Vulkan executor was split out to #127 (stacked on this branch), per the review recommendation here and the preference for small, focused PRs in #36. The branch was also rebased onto current `main` (post-#108).

## What this adds

**`SharpEmu.Tools.ShaderDump`** — a standalone developer tool under `tools/` that verifies the Gen5 shader translation pipeline with **synthetic instruction programs** — no game dumps involved anywhere, and **no changes to emulator source** (the tool reaches the translator via reflection). The project is not in the main solution; the emulator build is untouched.

It feeds hand-assembled gfx10 instruction words (cross-checked against LLVM's AMDGPU tables, same method as #103/#106/#108) through the real `Gen5ShaderTranslator` → `Gen5SpirvTranslator` pipeline and dumps the vertex + compute SPIR-V blobs for `spirv-val` / `spirv-dis` inspection.

Each bundled program carries an expectation, and any unexpected outcome makes the tool **exit non-zero** so it can gate scripts/CI:

- `fmac`, `muls`, `sopp-hints`, `exec` must decode and emit both stages;
- `sopp-mode` (`s_round_mode`, `s_denorm_mode`) **pins the loud `unknown-sopp` decode failure** those FP MODE writes must keep producing until their semantics are modeled — the exact behavior settled in the #108 review. If someone adds them to the decode table without modeling the semantics, this program fails the run.

The `exec` program computes real ALU results, stores them with `buffer_store_dword`, and toggles EXEC off/on around a pair of stores; its `exec-cs.spv` blob is what #127 executes and verifies numerically on real hardware.

All six dumped blobs pass `spirv-val --target-env vulkan1.3`.

## Changes since the last review

- Split per recommendation: ShaderDump here, Vulkan executor in #127.
- The exec program now genuinely exercises the EXEC store guard (`s_mov_b32 exec_lo, 0` / `-1` around stores) instead of only claiming to; the executor-side sentinel/exit-code fixes are in #127.
- The stale combined `sopp` program is split into `sopp-hints` (decodes and compiles now that #108 landed) and `sopp-mode` (pinned loud failure), and the tool itself now exits non-zero on any unexpected outcome, so the pin is machine-checked.
- Reflection lookups fail with the missing member's name instead of a bare `NullReferenceException`.
- Per maintainer direction: `tools/README.md` and the `packages.lock.json` files are removed. The tool projects opt out of lock files (`RestorePackagesWithLockFile=false`) because `Directory.Build.props` enables them repo-wide — without the opt-out, any restore would regenerate the lock file and dirty the worktree. Usage/docs now live in this PR description and the commit message.

## Why

Most shader-translator work (my #103/#106/#108 included) is verifiable without dumps, but until now only down to "compiles". This tool closes decode → emit → validate; #127 extends it to driver-accept → execute → compare. It also gives dump-less contributors a way to test regressions.

## Open questions for maintainers

1. **Reflection vs `InternalsVisibleTo`:** the tool uses reflection so this PR touches zero emulator source. If you'd rather have `InternalsVisibleTo` (as #36 also proposes for its xunit harness), the reflection block collapses to direct calls — happy to switch.
2. **Coordination with #36:** the synthetic-program approach overlaps with the planned test corpus there; this tool could become a consumer of that corpus, or its programs could migrate into the xunit harness once it lands.
3. **Naming/location:** `tools/` + `SharpEmu.Tools.*` — happy to rename.

### DIFF
--- a/tools/SharpEmu.Tools.ShaderDump/Program.cs
+++ b/tools/SharpEmu.Tools.ShaderDump/Program.cs
@@ -1,0 +1,275 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// Synthetic-shader conformance dumper.
+//
+// Feeds hand-assembled Gen5 (gfx10) instruction words through the real
+// decode -> SPIR-V pipeline (Gen5ShaderTranslator / Gen5SpirvTranslator, via
+// reflection so no emulator source changes are required) and writes the
+// resulting vertex and compute SPIR-V blobs to disk. The blobs can then be
+// checked with spirv-val / spirv-dis.
+//
+// Programs that contain buffer_store_dword automatically get a single
+// global-memory binding covering every store, which the emitter exposes as
+// guestBuffers[0] (descriptor set 0, binding 0).
+//
+// Each program carries an expectation: ExpectTranslate=true programs must
+// decode and emit both stages; ExpectTranslate=false programs pin a decode
+// failure that must stay loud. Any unexpected outcome makes the tool exit
+// non-zero, so it can gate scripts/CI.
+//
+// Usage: SharpEmu.Tools.ShaderDump [output-directory]
+
+using System.Buffers.Binary;
+using System.Reflection;
+using SharpEmu.HLE;
+using SharpEmu.Libs.CxxAbi;
+
+const ulong ProgramAddress = 0x100000;
+
+(string Name, bool ExpectTranslate, uint[] Words)[] testPrograms =
+[
+    ("fmac", true, [
+        0x560A0501,             // v_fmac_f32 v5, v1, v2
+        0x580A0501, 0x42280000, // v_fmamk_f32 v5, v1, 42.0, v2
+        0x5A0A0501, 0x42280000, // v_fmaak_f32 v5, v1, v2, 42.0
+        0xD52B0005, 0x00020501, // v_fmac_f32_e64 v5, v1, v2
+        0xBF810000,             // s_endpgm
+    ]),
+    ("muls", true, [
+        0xD5690005, 0x00020501, // v_mul_lo_u32 v5, v1, v2
+        0xD56A0005, 0x00020501, // v_mul_hi_u32 v5, v1, v2
+        0xD56B0005, 0x00020501, // v_mul_lo_i32 v5, v1, v2
+        0xD56C0005, 0x00020501, // v_mul_hi_i32 v5, v1, v2
+        0xBF810000,             // s_endpgm
+    ]),
+    ("sopp-hints", true, [
+        0xBFA10001,             // s_clause 0x1
+        0xBFA30000,             // s_waitcnt_depctr 0x0
+        0xBF810000,             // s_endpgm
+    ]),
+    // s_round_mode / s_denorm_mode write the FP MODE state and must keep
+    // failing decode loudly until their semantics are modeled (see #108);
+    // this program pins that behavior.
+    ("sopp-mode", false, [
+        0xBFA40000,             // s_round_mode 0x0
+        0xBFA50000,             // s_denorm_mode 0x0
+        0xBF810000,             // s_endpgm
+    ]),
+    // Executable end-to-end test: compute with real ALU instructions, then
+    // buffer_store_dword results to guestBuffers[0] at offsets 0/4/8, prove
+    // that a store with EXEC=0 does not land (offset 12 stays sentinel), and
+    // that stores work again after EXEC is restored (offset 16).
+    ("exec", true, [
+        0xBFA10001,             // s_clause 0x1 (hint no-op in an executed program, needs #108)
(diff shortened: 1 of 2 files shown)


kostyaff (commented):
I reproduced the draft locally at `c65d901` with SDK 10.0.103:

- locked restore/build of both tools: PASS, 0 warnings / 0 errors;
- ShaderDump: PASS, six blobs produced (the SOPP program currently fails at `0x21` as documented);
- GpuConformance on AMD Radeon Vega 8: pipeline accepted, all three arithmetic values and the current sentinel check pass;
- the worktree remains clean. (`spirv-val`/`spirv-dis` are not installed on this machine, so I did not independently repeat that part.)

The concept is useful and the real decode → emit → driver → execute path works, but I found two correctness blockers in the conformance claim:

1. **Sentinel corruption cannot fail the run.** `Program.cs:341-346` prints `FAIL` when `words[3]` changes, but does not increment `failures`; lines 344-346 can therefore print `RESULT: all values match` and line 360 exits 0. It also checks only `words[3]`, although the 64-byte buffer has `words[3..15]` beyond the three expected results. Please include every trailing word in the failure count and exit status.

2. **EXEC masking is not actually exercised.** The `exec` program never changes EXEC, so all three stores run with the translator's default active EXEC. A regression that ignores the store guard would still pass, despite the PR/README claiming coverage of “exec-masked stores”. Please add an inactive-EXEC store whose destination must remain sentinel (or narrow the coverage claim).

Two smaller follow-ups:

- The SOPP text is stale after #108 was correctly narrowed: once #108 lands, `0x21` and `0x23` decode, but this combined program will still stop at `s_round_mode` (`0x24`). Split the true hints from the expected-failure mode instructions or update the statement that the program “starts passing when #108 lands”.
- Both added `packages.lock.json` files lack the final newline required by CONTRIBUTING/.editorconfig.

Given the 841-line review surface and the maintainer's preference in #36 for small focused PRs, I also recommend landing ShaderDump/documentation first and keeping the Vulkan executor as a follow-up. The central constraints otherwise check out: inputs are fully synthetic, the real translator is invoked, and emulator source/solution/CI are untouched.

Deeptanshuu:
Thanks for another review with independent reproduction — the Vega 8 run is doubly useful since it's a second vendor confirming the pipeline. All four points were correct; all four are addressed, and I've also taken the structural recommendation:

**Split done.** This PR is now `SharpEmu.Tools.ShaderDump` only (head `8bc9e04`, rebased onto current `main`); the Vulkan executor moved to #127, stacked on this branch. That also matches @par274's small-focused-PRs preference in #36.

**1. Sentinel corruption now fails the run** (in #127). The sentinel check joined the expected-results array, every trailing word of the 64-byte buffer is checked (`words[5..15]`, not just `words[3]`), and all mismatches feed one failure counter that drives the exit code. Verified negatively: running the executor against a blob that stores nothing reports 4 mismatches and exits 1.

**2. EXEC masking is now actually exercised.** The `exec` program clears the mask with `s_mov_b32 exec_lo, 0` before a `buffer_store_dword` whose destination must keep the sentinel, then restores it with `s_mov_b32 exec_lo, -1` before a store that must land. Both directions verified bit-exact on the RTX 3060 (offset 12 stays `0xCAFEBABE`, offset 16 receives `0x3FC00000`). One honest caveat, noted in #127: the mask is toggled full-off/full-on, which is exact for the current single-invocation dispatch but doesn't verify partial-wave save/restore semantics — that needs a wider dispatch and stays follow-up.

**3. SOPP text/program unstale.** You were right that the combined program would just stop at `s_round_mode` after the #108 narrowing. It's now two programs with machine-checked expectations: `sopp-hints` (`s_clause`, `s_waitcnt_depctr`) must decode and emit (and does, on rebased `main`); `sopp-mode` (`s_round_mode`, `s_denorm_mode`) pins the loud `unknown-sopp` failure. ShaderDump itself now exits non-zero on any unexpected outcome — including `sopp-mode` unexpectedly *succeeding* — so the pin from the #108 review outcome is enforced, not just documented.

**4. Lockfile newlines fixed** in both projects.

Also folded in a self-review pass: reflection lookups now fail with the missing member's name (`Gen5ShaderTranslator.TryDecodeProgram not found`) instead of a bare NRE, and the executor in #127 reports clear errors when a device lacks a compute queue family, a host-visible+coherent memory type, or `shaderInt64`, instead of undefined behavior.

Since you don't have spirv-val on that machine: all six blobs from `8bc9e04` still pass `spirv-val --target-env vulkan1.3` here (SPIRV-Tools from the Khronos CI artifacts).

PR #102, opened 2026-07-13
[HLE] Fix guest-thread sync and boot for Unreal Engine titles

Spooks4576:
Silent Hill: The Short Message (and other UE titles) now boot the full engine thread graph instead of hanging early. Four related fixes:

- pthread cond/mutex semantics: retain a signal raised with no waiter as pending, and key block/wake on the state's identity rather than a resolved address that could differ between lock and unlock. This ends the ~1.5M-call cond_wait busy-spin.

- Warm HLE type initializers and force-JIT their methods on a host thread at Freeze(). A .cctor or first-time JIT running on a guest thread's hijacked stack fail-fasts the CLR as "Invalid Program: attempted to call a UnmanagedCallersOnly method from managed code".

- Guest thread scheduling: pump after a wake so a readied thread actually runs, add a dispatcher thread for when every guest thread is parked, and make the pump-depth guard an atomic CAS.

- Route mutex/rwlock lock/unlock off the non-blocking leaf-import fast path so a contended lock can deschedule its guest thread.

Ported from the unreal-boot-fixes branch.

PR #112, opened 2026-07-13
Fix/videoout vblank pump

Spooks4576:
Fixed VBlank Pump that fixes video freezing issues on the latest commit.

confirmed demon souls and Sara Games to still be functional

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
@@ -810,8 +810,14 @@ public sealed partial class DirectExecutionBackend
 			return !_logUsleep;
 		}
 
-		// Mutex/rwlock lock/unlock excluded: the leaf path can't block a contended lock.
+		// Only mutex/rwlock *lock* is excluded: it may block a contended acquire, which the
+		// leaf path can't. unlock never blocks and stays here — routing it off the fast path
+		// slows guest spinlocks enough to livelock (Demon's Souls).
 		return nid is
+			"tn3VlD0hG60" or // scePthreadMutexUnlock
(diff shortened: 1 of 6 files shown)


kostyaff (commented):
I reviewed head `966f3b3` against current `main` (`63b440e`) and ran the pinned SDK 10.0.103 locked restore plus full Release build: 0 warnings / 0 errors.

There are two blockers before this is safe to merge:

1. **Vblank now has two independent producers.** The new timer calls `PumpVblanks()` every 16 ms, but the existing `sceVideoOutWaitVblank` path still sleeps 1 ms and calls `SignalVblank(port)` itself. A title that waits in a loop therefore advances `VblankCount` and emits equeue events at caller frequency *in addition to* timer ticks (potentially ~1000 Hz + timer), rather than waiting for the next hardware-like edge. `sceVideoOutWaitVblank` should wait for the timer-driven generation to advance, not manufacture another edge.

2. **The branch ancestry obscures and enlarges the review.** Commits through `2727f32` are already merged as #102, but this branch still forks at `298ef01`; GitHub consequently shows the five #102 files again alongside the 54-line VideoOut change. Please rebase current `main` and carry only `966f3b3` (or open a fresh branch with that commit). That will also ensure later main changes such as #104 remain visibly intact and let the VBlank change be reviewed/tested in isolation.

A smaller cadence issue remains after that: a fixed 16 ms period is 62.5 Hz, while the port reports 60 Hz. A monotonic schedule derived from `RefreshRate` (nominally 16.666... ms) would avoid systematic 4.17% drift.

The core idea—autonomous vblank instead of relying on flips—is sound, but it needs a single source of edges and a current-main branch. Please also record at least a synthetic cadence/non-duplication check (register an equeue, loop `sceVideoOutWaitVblank`, verify one sequence at the selected refresh rate), since CI currently proves buildability only.

Issue #113, opened 2026-07-13
[HLE] Follow up condition-variable semantics from #102


kostyaff:
## Context

Current `main` at `63b440efcddaede08a588c3f6086d68d085b3f07` contains the condition-variable changes merged in #102. The compatibility goal is valid, but two semantics need an explicit follow-up before more pthread work builds on them.

## 1. Signals are persisted when no thread is waiting

`PthreadCondSignalCore` increments `PendingSignals` when `state.Waiters == 0`, and the next waiter consumes it. POSIX condition signals are edges, not semaphore credits: `pthread_cond_signal()` / `broadcast()` have no effect when nobody is currently blocked ([The Open Group](https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_cond_signal.html)).

This can both bypass a future predicate and create a lock inversion in the pending path:

1. T1 signals with no waiter (`PendingSignals = 1`).
2. T2 holds mutex M and enters cond-wait, locks `state.SyncRoot`, then releases M.
3. T3 acquires M and calls cond-signal, blocking on `state.SyncRoot`.
4. T2 consumes the pending signal and calls `PthreadMutexRelockForCondWait` while still holding `state.SyncRoot`, blocking on M.

T2 now holds the condition lock while waiting for M; T3 holds M while waiting for the condition lock.

Relevant current-main lines: `KernelPthreadCompatExports.cs:76-88`, `:1186-1191`, `:1302-1319`.

## 2. POSIX `pthread_cond_timedwait` treats its pointer as microseconds

The newly registered POSIX export at `KernelPthreadCompatExports.cs:347-352` forwards directly to `PthreadCondTimedwait`, which truncates `Rdx` to `uint timeoutUsec`. For the POSIX ABI, the third argument is `const struct timespec *abstime`, confirmed by the [PS4dev FreeBSD include mirror](https://github.com/ps4dev/freebsd-include-mirror/blob/master/pthread.h) and [The Open Group specification](https://pubs.opengroup.org/onlinepubs/009604599/functions/pthread_cond_timedwait.html). A guest address is therefore being interpreted as a relative microsecond duration, causing arbitrary early/late timeouts.

`scePthreadCondTimedwait` may retain its separate relative-time ABI; the POSIX-named export needs its own guest `timespec` read and absolute-deadline conversion.

## Suggested safe follow-up

[...]

PR #114, opened 2026-07-13
[VideoOut] Drive vblank from a timer so UE render threads advance

Spooks4576:
Fixes Sara Game Freezing because of the VBlank

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VideoOutExports.cs
+++ b/src/SharpEmu.Libs/VideoOut/VideoOutExports.cs
@@ -45,6 +45,59 @@ public static class VideoOutExports
     private static readonly object _stateGate = new();
     private static readonly object _frameDumpGate = new();
     private static readonly Dictionary<int, VideoOutPortState> _ports = new();
+
+    // Hardware raises vblank autonomously; UE blocks its frame loop on it.
+    private static Timer? _vblankPumpTimer;
+    private static int _vblankPumpActive;
+    private const int VblankPumpIntervalMs = 16;
+
+    private static void EnsureVblankPumpStarted()
+    {
+        if (_vblankPumpTimer is not null)
+        {
+            return;
+        }
+
+        _vblankPumpTimer = new Timer(
+            static _ => PumpVblanks(),
+            null,
+            VblankPumpIntervalMs,
+            VblankPumpIntervalMs);
+    }
+
+    private static void PumpVblanks()
+    {
+        if (Interlocked.Exchange(ref _vblankPumpActive, 1) != 0)
+        {
+            return;
+        }
+
+        try
+        {
+            VideoOutPortState[] ports;
+            lock (_stateGate)
+            {
+                if (_ports.Count == 0)
+                {
+                    return;
+                }
+
+                // Signalling reaches WakeBlockedThreads -> Pump(), which serialises on one global
+                // flag. Waking an unwatched queue would hold it 60x/sec and starve guest threads.
+                ports = _ports.Values.Where(static port => port.VblankEvents.Count != 0).ToArray();
+            }
+
+            foreach (var port in ports)
+            {
+                SignalVblank(port);
+            }
+        }
+        finally
+        {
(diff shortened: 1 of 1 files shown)


StealUrKill:
The release still freezes for me.

Spooks4576:
> The release still freezes for me.

can you tell me where and how it freezes 
several people were getting freezes on the menu and i fixed that

StealUrKill:
> > The release still freezes for me.
> 
> can you tell me where and how it freezes several people were getting freezes on the menu and i fixed that

Yes going to the sound/language menu and then backing out freezes and crashes the game. Implementing the fix suggested here fixes it.

https://github.com/par274/sharpemu/issues/113


Once I get my commits fixed, I will create a branch and push a PR.



image

kostyaff:
Thanks for the reports. There is still a correctness issue in this merged change that is independent of the condition-variable fix in #113.

`PumpVblanks()` now calls `SignalVblank(port)` every 16 ms, but current `sceVideoOutWaitVblank` still does `Thread.Sleep(1); SignalVblank(port);`. That leaves two independent VBlank producers: the timer and every guest wait call. The effective event rate can therefore depend on caller frequency, and 16 ms is 62.5 Hz rather than 60 Hz.

The safer model is one producer (the timer/display clock) plus a wait path that blocks until the VBlank generation changes. Until that exists, I would strongly consider a follow-up or revert rather than treating the timer as a complete fix.

For the reported white-screen change, could someone A/B the same title, build configuration, and data at `63b440e` (before this PR) versus `50f78f3` (after merge)? #114 was the only change to `main` between those commits, so that comparison would confirm or rule it out. This is currently a strong candidate, not proven causality.

StealUrKill:
> Thanks for the reports. There is still a correctness issue in this merged change that is independent of the condition-variable fix in #113.
> 
> `PumpVblanks()` now calls `SignalVblank(port)` every 16 ms, but current `sceVideoOutWaitVblank` still does `Thread.Sleep(1); SignalVblank(port);`. That leaves two independent VBlank producers: the timer and every guest wait call. The effective event rate can therefore depend on caller frequency, and 16 ms is 62.5 Hz rather than 60 Hz.
> 
> The safer model is one producer (the timer/display clock) plus a wait path that blocks until the VBlank generation changes. Until that exists, I would strongly consider a follow-up or revert rather than treating the timer as a complete fix.
> 
> For the reported white-screen change, could someone A/B the same title, build configuration, and data at `63b440e` (before this PR) versus `50f78f3` (after merge)? #114 was the only change to `main` between those commits, so that comparison would confirm or rule it out. This is currently a strong candidate, not proven causality.

The white screen is my doing getting Silent Hill to boot and fixing all the issues. I do not think anyone yet has gotten it this far. I have not pushed a PR due to the fact that the rendering is getting stuck. The game draws a fullscreen quad into the display buffer sampling a 1×1 white placeholder texture (tex0desc=1x1). White seem like it's placeholder content.
Black = the game advancing into real rendering and clearing, but the actual frame never landing on screen.

Spooks4576:
> Thanks for the reports. There is still a correctness issue in this merged change that is independent of the condition-variable fix in #113.
> 
> `PumpVblanks()` now calls `SignalVblank(port)` every 16 ms, but current `sceVideoOutWaitVblank` still does `Thread.Sleep(1); SignalVblank(port);`. That leaves two independent VBlank producers: the timer and every guest wait call. The effective event rate can therefore depend on caller frequency, and 16 ms is 62.5 Hz rather than 60 Hz.
> 
> The safer model is one producer (the timer/display clock) plus a wait path that blocks until the VBlank generation changes. Until that exists, I would strongly consider a follow-up or revert rather than treating the timer as a complete fix.
> 
> For the reported white-screen change, could someone A/B the same title, build configuration, and data at `63b440e` (before this PR) versus `50f78f3` (after merge)? #114 was the only change to `main` between those commits, so that comparison would confirm or rule it out. This is currently a strong candidate, not proven causality.

will be adding the safer model for the vblank 
as for the actual audio settings crash can you explain to me what you think causes that?

PR #115, opened 2026-07-13
[pthread] Fix cond-var lock inversion and POSIX timedwait ABI

StealUrKill:
Two follow-ups to upstream #102's condition-variable changes, as analyzed in upstream issue #113:

- The pending-signal consume path reacquired the guest mutex while still holding the condition state lock, inverting lock order against cond-signal (mutex -> SyncRoot) and deadlocking both threads. Leave the condition lock before relocking, matching the normal wake path. This unfroze Dreaming Sarah (PPSA02929) at its title screen.

- pthread_cond_timedwait's third argument is a pointer to an absolute CLOCK_REALTIME timespec, not a relative microsecond count; the guest address was being truncated into a duration, yielding arbitrary timeouts. Read the timespec and convert to a relative wait. scePthreadCondTimedwait keeps its separate relative-time ABI.

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelPthreadCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelPthreadCompatExports.cs
@@ -349,7 +349,32 @@ public static class KernelPthreadCompatExports
         ExportName = "pthread_cond_timedwait",
         Target = Generation.Gen4 | Generation.Gen5,
         LibraryName = "libKernel")]
-    public static int PosixPthreadCondTimedwait(CpuContext ctx) => PthreadCondTimedwait(ctx);
+    public static int PosixPthreadCondTimedwait(CpuContext ctx)
+    {
+        // POSIX passes `const struct timespec *abstime` (absolute deadline), not relative microseconds (#113).
+        var abstimeAddress = ctx[CpuRegister.Rdx];
+        if (abstimeAddress == 0)
+        {
+            return (int)OrbisGen2Result.ORBIS_GEN2_ERROR_INVALID_ARGUMENT;
+        }
+
+        if (!ctx.TryReadUInt64(abstimeAddress, out var deadlineSeconds) ||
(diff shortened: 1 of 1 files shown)


Spooks4576:
> Two follow-ups to upstream #102's condition-variable changes, as analyzed in upstream issue #113:
> 
> * The pending-signal consume path reacquired the guest mutex while still holding the condition state lock, inverting lock order against cond-signal (mutex -> SyncRoot) and deadlocking both threads. Leave the condition lock before relocking, matching the normal wake path. This unfroze Dreaming Sarah (PPSA02929) at its title screen.
> * pthread_cond_timedwait's third argument is a pointer to an absolute CLOCK_REALTIME timespec, not a relative microsecond count; the guest address was being truncated into a duration, yielding arbitrary timeouts. Read the timespec and convert to a relative wait. scePthreadCondTimedwait keeps its separate relative-time ABI.

just tested this on my end and your fixes checks out 
fixes the crash on the main menu options screen. and does what its supposed to do 

Good work fixing my mistakes :)

StealUrKill:
> > Two follow-ups to upstream #102's condition-variable changes, as analyzed in upstream issue #113:
> > 
> > * The pending-signal consume path reacquired the guest mutex while still holding the condition state lock, inverting lock order against cond-signal (mutex -> SyncRoot) and deadlocking both threads. Leave the condition lock before relocking, matching the normal wake path. This unfroze Dreaming Sarah (PPSA02929) at its title screen.
> > * pthread_cond_timedwait's third argument is a pointer to an absolute CLOCK_REALTIME timespec, not a relative microsecond count; the guest address was being truncated into a duration, yielding arbitrary timeouts. Read the timespec and convert to a relative wait. scePthreadCondTimedwait keeps its separate relative-time ABI.
> 
> just tested this on my end and your fixes checks out fixes the crash on the main menu options screen. and does what its supposed to do
> 
> Good work fixing my mistakes :)

I was not sure what was causing the issue and had to test every release to find it haha. Then @kostyaff gave a big clue on what the issue was.

PR #122, opened 2026-07-14
[AGC] Complete gfx10 v_cmpx_f32 decode and emit ordered/unordered float compares

333cra:
Completes the gfx10 single-precision VOPC comparison coverage in the Gen5 translator. Two related gaps:

- `v_cmp_o_f32` (0x07), `v_cmp_u_f32` (0x08) and `v_cmp_nlg_f32` (0x0A) decode today but have no emission case, so a shader using any of them would fall through the float-compare switch to `SpirvOp.Nop` and fail with `unsupported float compare`. `nlg` (= "not less-greater" = unordered-or-equal) now maps to `OpFUnordEqual`. The ordered/unordered predicates only test whether an operand is NaN; SPIR-V's `OpOrdered`/`OpUnordered` require the **Kernel** capability and are invalid in a Shader module, so they are lowered from `OpIsNan`: `unordered = isnan(a) || isnan(b)`, `ordered = !unordered`, using only capability-free core ops.
- The whole `v_cmpx_*_f32` row (0x17–0x1C, 0x1F) is missing from the VOPC decode table, so any of them would fail decode as `unknown-vopc`. All seven entries are added. This is proactive table completion in the spirit of #108: `nge`/`ngt`/`nle`/`tru` already had emission cases and only lacked a decode entry, while `o`/`u`/`nlg` reuse the emission above. The `cmpx` EXEC-mask update is already modelled by the emitter (`active = exec & condition`), and `UsesWaveControl()` already keys off the `VCmpx` prefix, so the new mnemonics are covered automatically.

I don't have a game trace hitting these specific opcodes; this is a completeness fix that removes known decode/emit dead-ends, not a fix for an observed title.

Opcode semantics verified against LLVM `VOPCInstructions.td`: the f32 predicate order is `f, lt, eq, le, gt, lg, ge, o, u, nge, nlg, ngt, nle, neq, nlt, tru` (0x00–0x0F) with the `cmpx` forms at +0x10, and `COND_O`/`COND_UO`/`COND_UEQ` confirm the `o`/`u`/`nlg` semantics. Integer rows have no ordered/unordered predicates, so nothing analogous is missing there.

## Verification

- Encodings cross-checked against LLVM's `llvm-mc` regression tests ([`llvm/test/MC/AMDGPU/gfx10_asm_vopc.s`](https://github.com/llvm/llvm-project/blob/main/llvm/test/MC/AMDGPU/gfx10_asm_vopc.s), [`gfx10_asm_vopcx.s`](https://github.com/llvm/llvm-project/blob/main/llvm/test/MC/AMDGPU/gfx10_asm_vopcx.s)), which pin the exact expected byte sequences for all ten opcodes (e.g. `v_cmpx_o_f32 v1, v2` = `[0x01,0x05,0x2e,0x7c]` → opcode field 0x17). `DecodeVopc` extracts the right mnemonic for all of them.
- The o/u lowering was validated with `spirv-val --target-env vulkan1.1`: the `OpIsNan`/`OpLogicalOr`/`OpLogicalNot` module is valid, whereas the naive `OpOrdered` form is rejected with `"Opcode Ordered requires one of these capabilities: Kernel"` — which is why the synthesis is necessary.
- `dotnet build -c Release` passes with no new warnings.

## AI-Assisted Contributions

Developed with AI assistance. In my own words: `DecodeVopc` maps each gfx10 VOPC opcode to a mnemonic, and the SPIR-V emitter switches on that mnemonic to pick a comparison. I found three float predicates decoded-but-not-emitted and the entire `v_cmpx` float row missing from decode. The key subtlety is that `v_cmp_o/u_f32` cannot use `OpOrdered`/`OpUnordered`, because those need the Kernel capability that Vulkan shaders never declare — I confirmed both the opcode numbers and that rejection with `spirv-val`, and lowered them via `OpIsNan` instead. I can explain and maintain every line.

Sources: LLVM `VOPCInstructions.td` · Khronos `spirv.core.grammar.json` · LLVM MC regression tests


kostyaff (changes requested):
There is one blocking gfx10 semantic issue in `TryEmitVectorCompare`.

`StoreWaveMask(106, condition)` currently runs before the `VCmpx*` branch, so every `VCmpx*` instruction writes both VCC and EXEC. On gfx10, `VCmpx*` defines EXEC and must preserve VCC; the expected behavior is `EXEC = old EXEC & condition`. Non-X `VCmp*` should continue to write VCC only.

Please make the VCC store exclusive to the non-X path, for example:

```csharp
if (opcode.StartsWith("VCmpx", StringComparison.Ordinal))
{
    var active = _module.AddInstruction(
        SpirvOp.LogicalAnd,
        _boolType,
        Load(_boolType, _exec),
        condition);
    StoreWaveMask(126, active);
}
else
{
    StoreWaveMask(106, condition);
}
```

I verified this minimal change with locked restore, the full Release build (0 warnings, 0 errors), gfx10 opcode mapping, the ordered/unordered NaN truth table, and synthetic SPIR-V validation against Vulkan 1.2 for `0x07`, `0x08`, `0x0A`, and every `VCmpx*` opcode `0x17..0x1F`.

333cra:
Good catch, fixed in 669a175. You're right, the gfx10 cmpx encodings dropped the sdst operand entirely, you can see it in LLVM's `gfx10_asm_vopcx.s` tests where `v_cmpx_*` takes no destination at all. So the VCC store now only happens on the non-X path and EXEC keeps the existing `old EXEC & condition` update. Build is clean on my side.

PR #124, opened 2026-07-14
[VideoOut] Display game icon (icon0.png) in window title bar

realdody:
This PR loads the game's icon (icon0.png) from the sce_sys directory and sets it as the main window icon.

It falls back gracefully if the icon is missing or fails to load.

Tested on Windows 11 in Dead Cells, Dreaming Sarah, and Void Terrarium. Also verified that splash screen logic still works with Void Terrarium.

SharpEmu_KuMyJ4YDKH
SharpEmu_QkSvq5Uize
SharpEmu_FA67zmGIk0

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/PngSplashLoader.cs
+++ b/src/SharpEmu.Libs/VideoOut/PngSplashLoader.cs
@@ -14,6 +14,17 @@ internal static class PngSplashLoader
     ];
 
     public static bool TryLoad(out byte[] pixels, out uint width, out uint height)
+        => TryLoad("pic0.png", requestRgba: false, out pixels, out width, out height);
+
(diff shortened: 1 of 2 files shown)


par274:
I haven't checked that yet. Does the game icon also appear in the Start menu? It could use the game's icon there as well, while the GUI continues using the SharpEmu icon.

realdody:
> I haven't checked that yet. Does the game icon also appear in the Start menu? It could use the game's icon there as well, while the GUI continues using the SharpEmu icon.

I assume you mean the taskbar? And yes it does:
image



window.SetWindowIcon sets the icon for the game's window, which is where Windows gets the icon for the task bar from. So you have to use the same icon for both.

If you do mean start menu, like shortcuts for the emulator exe, that isn't affected by this as you can see in the screenshot.

PR #127, opened 2026-07-14
[Tools] GPU conformance executor: verify dumped shader blobs on real hardware

Deeptanshuu:
> Split out of #111 per the review there recommending the dump tool and the Vulkan executor land separately. Now that #111 is merged, this is a single commit rebased onto current `main`.

## What this adds

**`SharpEmu.Tools.GpuConformance`** — a standalone tool under `tools/` (not in the main solution, references only `Silk.NET.Vulkan`, no emulator source changes) that executes the `exec-cs.spv` blob produced by `SharpEmu.Tools.ShaderDump` (#111) on a real Vulkan device and compares **every word** of the 64-byte storage buffer against CPU-computed expectations, bit for bit. Creating the compute pipeline doubles as a driver-acceptance check for SharpEmu's emitted SPIR-V.

The checks cover:

- the three ALU results (`v_fmac_f32`, `v_mul_hi_i32`, `v_mul_lo_i32`) stored via `buffer_store_dword`;
- a store attempted with **EXEC = 0** (`s_mov_b32 exec_lo, 0`) whose destination must keep the sentinel — the exec-mask store guard is now genuinely exercised, per the #111 review;
- a store after EXEC is restored (`s_mov_b32 exec_lo, -1`) that must land;
- all trailing sentinel words.

Any mismatch — including a clobbered sentinel — counts toward the failure total and makes the tool exit non-zero (also per the #111 review; previously a clobbered sentinel printed `FAIL` but could not fail the run).

## Verified

Real output on my machine (RTX 3060 Laptop GPU):

```
executing on: NVIDIA GeForce RTX 3060 Laptop GPU
driver accepted the SPIR-V (pipeline created)
PASS  v_fmac_f32  fma(1.5, 2.25, 10.0): gpu=0x41560000 expected=0x41560000
PASS  v_mul_hi_i32 hi(0x7FFFFFFF*0x10003): gpu=0x00008001 expected=0x00008001
PASS  v_mul_lo_i32 lo(0x7FFFFFFF*0x10003): gpu=0x7FFEFFFD expected=0x7FFEFFFD
PASS  exec=0 store suppressed (offset 12 sentinel): gpu=0xCAFEBABE expected=0xCAFEBABE
PASS  store after exec restore (offset 16): gpu=0x3FC00000 expected=0x3FC00000
PASS  trailing words [5..15] intact (sentinel)
RESULT: all values match
```

The failure path is verified too: running the tool against a blob that stores nothing reports 4 mismatches and exits 1.

Robustness hardening from a self-review pass: clear errors instead of undefined behavior when no compute queue family, no host-visible+coherent memory type, or no `shaderInt64` support exists; the discrete-GPU preference now picks the *first* discrete device.

## Notes / open questions

- The tool asserts EXEC masking with a full-on/full-off mask, which is exact for the current single-invocation dispatch. Verifying partial-wave save/restore semantics would need a wider dispatch and is left as follow-up.
- A possible future direction is routing through the public `Gen5ShaderScalarEvaluator.TryEvaluate` instead of hand-building the global-memory binding, which would keep the harness in lockstep with the production evaluation path — happy to explore that in a follow-up if preferred.

### DIFF
--- a/tools/SharpEmu.Tools.GpuConformance/Program.cs
+++ b/tools/SharpEmu.Tools.GpuConformance/Program.cs
@@ -1,0 +1,412 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+// Executes the SharpEmu-emitted "exec" conformance shader on a real Vulkan
+// device and compares the buffer results against CPU-computed expected values.
+//
+// The shader (exec-cs.spv, produced by SharpEmu.Tools.ShaderDump) was
+// translated by SharpEmu from hand-assembled Gen5 instruction words and stores
+// results to guestBuffers[0]:
+//   [0] v_fmac_f32   -> fma(1.5f, 2.25f, 10.0f)
+//   [1] v_mul_hi_i32 -> high 32 bits of (int)0x7FFFFFFF * (int)0x00010003
+//   [2] v_mul_lo_i32 -> low  32 bits of the same product
+//   [3] store attempted with EXEC=0 -> must NOT land (sentinel remains)
+//   [4] store after EXEC restored   -> 1.5f (0x3FC00000)
+// Every other word of the buffer must still hold the sentinel afterwards.
+//
+// Creating the compute pipeline doubles as a driver-acceptance check for the
+// emitted SPIR-V; the dispatch then verifies the arithmetic numerically.
+//
+// Usage: SharpEmu.Tools.GpuConformance <path-to-exec-cs.spv>
+
+using Silk.NET.Core.Native;
+using Silk.NET.Vulkan;
+
+const uint Sentinel = 0xCAFEBABE;
+
+// Must match the 64-byte global-memory binding ShaderDump constructs for the
+// exec program.
+const ulong BufferSize = 64;
+
+var expectedFma = BitConverter.SingleToUInt32Bits(
+    MathF.FusedMultiplyAdd(1.5f, 2.25f, 10.0f));
+var product = (long)0x7FFFFFFF * 0x00010003;
+var expectedHi = (uint)(product >> 32);
+var expectedLo = (uint)product;
+var expectedRestored = BitConverter.SingleToUInt32Bits(1.5f);
+
+unsafe
+{
+    var spvPath = args.Length > 0
+        ? args[0]
+        : throw new InvalidOperationException(
+            "usage: SharpEmu.Tools.GpuConformance <path-to-exec-cs.spv>");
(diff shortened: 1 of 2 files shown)


par274:
Can you revert package lock files? And remove the .md files, its unnecessary.

Deeptanshuu:
Done — both PRs updated (#111 head `145978b`, this PR head `7dae773`):

- `tools/README.md` removed; usage/docs now live in the PR descriptions and commit messages.
- Both `packages.lock.json` files removed. One note: `Directory.Build.props` sets `RestorePackagesWithLockFile=true` repo-wide, so a plain delete would regenerate the lock file on every restore and leave contributors with a dirty worktree. The two tool projects therefore opt out explicitly (`<RestorePackagesWithLockFile>false</RestorePackagesWithLockFile>`). Verified: restore + Release build of both tools pass with no lock file appearing and a clean worktree.

Happy to drop the opt-out property instead and re-add the lock files if you'd rather keep lock-file coverage for the tools — whichever you prefer.

par274:
> Done — both PRs updated (#111 head `145978b`, this PR head `7dae773`):
> 
>     * `tools/README.md` removed; usage/docs now live in the PR descriptions and commit messages.
> 
>     * Both `packages.lock.json` files removed. One note: `Directory.Build.props` sets `RestorePackagesWithLockFile=true` repo-wide, so a plain delete would regenerate the lock file on every restore and leave contributors with a dirty worktree. The two tool projects therefore opt out explicitly (`<RestorePackagesWithLockFile>false</RestorePackagesWithLockFile>`). Verified: restore + Release build of both tools pass with no lock file appearing and a clean worktree.
> 
> 
> Happy to drop the opt-out property instead and re-add the lock files if you'd rather keep lock-file coverage for the tools — whichever you prefer.

It needs to stay disabled because it fails in CI.

Deeptanshuu:
Got it — the opt-out stays as-is then. Both branches already have it, so nothing further to change.

Deeptanshuu:
Rebased onto current `main` now that #111 is merged — this PR is a single commit again (`9229fa9`), no stacking. Marked ready for review.

Re-verified end-to-end against today's `main` (which includes the #122 v_cmpx and #156 changes): regenerated all blobs with ShaderDump, all eight pass `spirv-val --target-env vulkan1.3`, and the executor run on the RTX 3060 is all-PASS bit-exact, including the EXEC-masked store suppression.

One note from the re-verification: ShaderDump on `main` currently crashes because #156 added an optional parameter to `TryCompileVertexShader` (reflection doesn't apply C# defaults). Fix opened as #166; the executor itself is unaffected (no reflection), but regenerating fresh blobs needs that fix.

PR #128, opened 2026-07-14
Expand Aerolib from nids.csv and wire socket/net kernel NID handlers

MikeSaito:
## Summary

Applies the shared `nids.csv` catalog to fix unresolved and mislabeled kernel/network NIDs

- Regenerate `aerolib.bin` from `scripts/nids.csv` (~154k NID/name pairs), with `ps5_names.txt` as fallback for names missing from the CSV
- Remove mislabeled `sceKernelUnknown*` zero-return stubs
- Register real handlers using CSV NIDs for: `socket`, `connect`, `bind`, `getsockname`, `bzero`, `inet_pton`, `htons`, and `sceNetHtonl` / `sceNetHtons` / `sceNetNtohl` / `sceNetNtohs`
- Add `KernelSocketCompatExports` and hook socket fd close/read/write through the existing compat fd pool

## Limitations

- The CSV expands the symbol catalog; it does not add HLE for every listed symbol
- Socket handling is minimal client-side (no listen/accept/UDP)

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -7,7 +7,6 @@ path = [
     "global.json",
     "**/packages.lock.json",
     "scripts/ps5_names.txt",
-    "scripts/nids.csv",
     "src/SharpEmu.HLE/Aerolib/aerolib.bin",
     "src/SharpEmu.GUI/Languages/**",
     "_logs/**",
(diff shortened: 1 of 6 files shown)


par274:
Where is your resource?

https://github.com/zecoxao/sce_symbols

There's no need to keep a large CSV file in the scripts directory. We generate it from ps5_names using Python and then compile it into a binary format.

MikeSaito:
The CSV is the one from helpful you posted it for the NID fix

PR #130, opened 2026-07-14
Agent/fix gen5 thread agc compat

Spooks4576:
Multiple UE Compatibility and Performance improvements.

- RHI Thread Issue is resolved on UE Titles  
- Multiple Performance Improvements Dreaming Sarah No longer runs slow on mid tier hardware 
However do expect some areas to be worse than others in terms of performance.

I have confirmed that both Demon Souls and Dreaming Sarah have no regressions


Spooks4576:
fixing the failing check as we speak

PR #141, opened 2026-07-14
Improve Unreal Engine compatibility: recursive-mutex trylock, compute…

YkPhysics:
… opcodes, thread diagnostics

Three independent fixes that let Unreal Engine titles progress further through boot and rendering.

1. pthread mutex: self-owned normal/adaptive trylock no longer livelocks. UE's FCriticalSection is recursive and re-enters via TryLock(). Strict POSIX returns EBUSY on a self-owned trylock, which UE busy-spins on forever (observed as a UE audio/render thread issuing thousands of trylocks per second on a single mutex). A self-owned normal/adaptive trylock now increments the recursion count and returns OK, matching the blocking self-lock path directly below it; the matching unlock decrements the count.

2. Gen5 scalar shader translator: implement S_FF0_I32_B32 and S_GETREG_B32. S_FF0_I32_B32 was missing from the single-source handler, so it fell through to the two-source path and aborted the whole compute shader with a "scalar-source1 missing" error; it now computes the first-zero-bit position (FindILsb of the inverted operand). S_GETREG_B32 reads runtime hardware registers the translate-time evaluator cannot fold; it now resolves to 0 (matching what the runtime SPIR-V emitter already stores) instead of aborting translation. Both remove compute-shader translation failures in UE deferred/compute passes.

3. Thread-state watchdog (opt-in, SHARPEMU_DUMP_THREADS=1): periodically logs every guest thread's state, block reason, and last import so idle deadlocks are diagnosable. Off by default; zero runtime cost otherwise.

Net effect: UE games that previously deadlocked on a recursive-mutex livelock or bailed out of compute-shader translation get farther into boot and rendering.

### DIFF
--- a/src/SharpEmu.GUI/Languages/fr.json
+++ b/src/SharpEmu.GUI/Languages/fr.json
@@ -1,0 +1,129 @@
+{
+  "_languageName": "Français",
+
+  "Page.Library": "Bibliothèque",
+  "Page.Options": "Options",
+  "Page.GameCount.One": "1 jeu",
+  "Page.GameCount.Other": "{0} jeux",
+
+  "Library.SearchWatermark": "Rechercher dans la bibliothèque…",
+  "Library.AddFolder": "＋ Ajouter un dossier",
+  "Library.Rescan": "⟳ Analyser à nouveau",
+  "Library.OpenFile": "Ouvrir un fichier…",
+
+  "Library.Context.Launch": "Lancer",
+  "Library.Context.OpenFolder": "Ouvrir le dossier du jeu",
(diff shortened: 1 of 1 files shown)


YkPhysics:
image
image

YkPhysics:
image

YkPhysics:
image

YkPhysics:
I'll be committing the rest of my fixes to this pr later in the day.

par274:
Any regressions other titles?

YkPhysics:
From what I've heard Pacific Drive. This pr is active however so I'll let you know when I think its ready for a merge.

PR #145, opened 2026-07-14
[AGC] Decode VOP3P and emit packed f16 arithmetic (first slice)

333cra:
Draft — opening for discussion per CONTRIBUTING (this touches instruction dispatch).

First support for the VOP3P (packed 16-bit math) family in the Gen5 translator. Scope is deliberately one slice: the five packed-f16 arithmetic ops `v_pk_add_f16`, `v_pk_mul_f16`, `v_pk_fma_f16`, `v_pk_min_f16`, `v_pk_max_f16`. Packed integer ops and the `v_fma_mix*` family are left for follow-up slices (see below).

**Dispatch correction.** On gfx10 VOP3P has its own `0b110011000` prefix (word0 top byte `0xCC`, i.e. `word >> 26 == 0x33`), separate from the VOP3 block — not the `0x3F` major the current `DecodeRaw2("Vop3p")` case assumes. Ground truth: `v_pk_add_f16 v1, -0.5, v2` disassembles from `[0x01,0x40,0x0f,0xcc,0xf1,0x04,0x02,0x1a]` (word0 `0xcc0f4001`, LLVM `gfx10_vop3p_literalv216.txt`). Today a real `v_pk_*` word is therefore matched by the `0x33 → SMEM` branch, and SMEM instructions are emitted as no-ops — i.e. packed math is silently dropped rather than failing loudly. This adds a precise 9-bit prefix check (`(word & 0xFF800000) == 0xCC000000`) ahead of the major switch; gfx10 SMEM lives at `0x3D` so the existing SMEM path is untouched. The dead `0x3F → DecodeRaw2` case is left as-is to keep the slice focused; happy to remove it here or separately.

**Decode.** New `DecodeVop3p` names the five opcodes (`0x0E/0x0F/0x10/0x11/0x12`) and `CreateInstruction` builds the three sources, the vector destination, and a new `Gen5Vop3pControl` carrying `op_sel`/`op_sel_hi`/`neg_lo`/`neg_hi`/`clamp`. Bit layout pinned to LLVM MC ground truth (`llvm/test/MC/AMDGPU/vop3p.s` for the gfx9 field/opcode encodings, `gfx10_vop3p_literalv216.txt` for the gfx10 prefix; the field layout is identical across gfx9/gfx10, only the prefix moved — opcode numbers confirmed unchanged in `VOP3PInstructions.td`) and the public RDNA2 ISA. `op_sel_hi` is split across both dwords (word0[14] plus word1[28:27]).

**Emission.** Each source register holds two f16. The lowering unpacks every operand to an f32 vec2 (`UnpackHalf2x16`), applies `op_sel` half-selection and `neg`, runs the op component-wise in f32 (`FAdd`/`FMul`, GLSL `FMin`/`FMax`/`Fma`), and repacks with `PackHalf2x16`. No `Float16` capability is required. Precision caveat: the maths runs in f32 with a single rounding at the pack. For add/mul/min/max this is bit-exact to f16 (a product/sum of two f16 fits in f32); for fma it can double-round versus a fused f16 FMA. min/max use GLSL `FMin`/`FMax` to match the existing scalar `v_min_f32`/`v_max_f32` path, sharing its NaN-selection imprecision. Modifiers that this slice does not model (`clamp`), inline/literal packed constants, and out-of-scope packed opcodes all fail with a clear error rather than emitting a wrong result.

## Verification

- Decode cross-checked against the LLVM MC encodings above: the gfx10 prefix + opcode and all five modifier fields (`op_sel`, the split `op_sel_hi`, `neg_lo`, `neg_hi`, `clamp`) reproduce the expected byte arrays (e.g. `neg_lo:[0,0,1]` sets word1[31], `neg_hi:[0,0,1]` sets word0[10]). Field extraction was verified by code review plus a script mirroring the pinned layout over the LLVM byte arrays; I have not run a synthetic program through the full translator pipeline end-to-end, and I don't have a game trace hitting these opcodes — hence the draft, so this can be exercised against real titles.
- The emitted SPIR-V sequence (`UnpackHalf2x16` → component ops → `PackHalf2x16`) validates under `spirv-val --target-env vulkan1.1`. A Pack/Unpack round-trip table for `1.0`, `-0.5`, an f16 NaN and an f16 denormal is exact.
- `dotnet build SharpEmu.slnx -c Release` passes with no new warnings.

## Planned follow-up slices

Kept out to stay single-topic: inline/literal packed constants (common in real shaders, so likely the next slice); packed integer arithmetic (`v_pk_add_u16/i16`, `v_pk_mul_lo_u16`, `v_pk_{min,max}_{u16,i16}`, packed shifts) which needs per-lane 16-bit int handling; then `v_fma_mix*` (mixed f16/f32) which has different operand typing. `clamp` support for the packed ops can also come as a small follow-up.

## AI-Assisted Contributions

Developed with AI assistance. In my own words: gfx10 encodes packed math in the VOP3P block under the `0xCC` prefix; the translator was matching that prefix as SMEM and silently dropping the instructions. I added a precise prefix check, a decoder that names the five f16 arithmetic opcodes and extracts the `op_sel`/`neg` modifiers (`op_sel_hi` is annoyingly split across both instruction dwords), and an emitter that lowers each op as unpack-to-f32-vec2 → operate → pack, because emitting real f16 would drag in the `Float16` capability. I pinned every bit position to the LLVM MC test encodings rather than guessing, verified the emitted SPIR-V with `spirv-val`, and checked the pack/unpack round-trip on edge values. I understand the precision trade-off (f32 intermediate, single rounding at pack) and can maintain and extend this.

Sources: LLVM `vop3p.s` / `gfx10_vop3p_literalv216.txt` · LLVM `VOP3PInstructions.td` · AMD RDNA2 ISA

### DIFF
--- a/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
+++ b/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
@@ -948,6 +948,18 @@ public static partial class Gen5SpirvTranslator
                         vector);
                     break;
                 }
+
+                case "VPkAddF16":
+                case "VPkMulF16":
+                case "VPkMinF16":
+                case "VPkMaxF16":
+                case "VPkFmaF16":
+                    if (!TryEmitPackedF16(instruction, out result, out error))
+                    {
+                        return false;
+                    }
+
+                    break;
                 default:
                     error = $"unsupported vector opcode {instruction.Opcode}";
                     return false;
@@ -973,6 +985,101 @@ public static partial class Gen5SpirvTranslator
             }
 
             StoreV(destination, result);
+            return true;
+        }
+
+        // Packed f16 (VOP3P) arithmetic. Each source register holds two f16 values.
+        // We unpack every operand to an f32 vec2 (UnpackHalf2x16), apply the op_sel
+        // half-selection and neg modifiers, run the operation component-wise in f32,
+        // and pack the result back with PackHalf2x16. The maths therefore happens in
+        // f32 with a single rounding at the pack, not with an exact intermediate f16
+        // rounding, which is accurate enough for the values shaders actually use.
+        private bool TryEmitPackedF16(
+            Gen5ShaderInstruction instruction,
+            out uint result,
+            out string error)
+        {
+            result = 0;
+            error = string.Empty;
+            if (instruction.Control is not Gen5Vop3pControl control)
+            {
+                error = $"missing vop3p control for {instruction.Opcode}";
+                return false;
+            }
+
+            if (control.Clamp)
+            {
+                error = $"unsupported vop3p modifiers (clamp) for {instruction.Opcode}";
+                return false;
+            }
+
+            var sourceCount = instruction.Opcode == "VPkFmaF16" ? 3 : 2;
+            var operands = new uint[sourceCount];
+            for (var index = 0; index < sourceCount; index++)
+            {
+                if (!TryGetPackedF16Operand(instruction, control, index, out operands[index], out error))
+                {
+                    return false;
+                }
+            }
+
+            var packed = instruction.Opcode switch
+            {
+                "VPkAddF16" => _module.AddInstruction(
+                    SpirvOp.FAdd, _vec2Type, operands[0], operands[1]),
+                "VPkMulF16" => _module.AddInstruction(
+                    SpirvOp.FMul, _vec2Type, operands[0], operands[1]),
+                // GLSL.std.450 FMin/FMax, matching the scalar v_min_f32/v_max_f32 path.
+                "VPkMinF16" => Ext(37, _vec2Type, operands[0], operands[1]),
+                "VPkMaxF16" => Ext(40, _vec2Type, operands[0], operands[1]),
+                "VPkFmaF16" => Ext(50, _vec2Type, operands[0], operands[1], operands[2]),
+                _ => 0u,
+            };
+            if (packed == 0)
+            {
+                error = $"unsupported packed opcode {instruction.Opcode}";
+                return false;
+            }
+
+            result = Ext(58, _uintType, packed);
+            return true;
+        }
+
+        private bool TryGetPackedF16Operand(
+            Gen5ShaderInstruction instruction,
+            Gen5Vop3pControl control,
+            int sourceIndex,
+            out uint operand,
+            out string error)
+        {
+            operand = 0;
+            error = string.Empty;
+            var source = instruction.Sources[sourceIndex];
+            if (source.Kind is not (Gen5OperandKind.VectorRegister or Gen5OperandKind.ScalarRegister))
+            {
+                error =
+                    $"unsupported vop3p operand {source} for {instruction.Opcode} (first slice: registers only)";
+                return false;
+            }
+
+            var unpacked = Ext(62, _vec2Type, GetRawSource(instruction, sourceIndex));
+            var low = _module.AddInstruction(SpirvOp.CompositeExtract, _floatType, unpacked, 0);
+            var high = _module.AddInstruction(SpirvOp.CompositeExtract, _floatType, unpacked, 1);
+
+            var laneLow = ((control.OpSelMask >> sourceIndex) & 1) != 0 ? high : low;
+            var laneHigh = ((control.OpSelHiMask >> sourceIndex) & 1) != 0 ? high : low;
+            if (((control.NegLoMask >> sourceIndex) & 1) != 0)
+            {
+                laneLow = _module.AddInstruction(SpirvOp.FNegate, _floatType, laneLow);
+            }
+
+            if (((control.NegHiMask >> sourceIndex) & 1) != 0)
+            {
+                laneHigh = _module.AddInstruction(SpirvOp.FNegate, _floatType, laneHigh);
+            }
+
+            operand = _module.AddInstruction(
+                SpirvOp.CompositeConstruct, _vec2Type, laneLow, laneHigh);
             return true;
         }
 
--- a/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
+++ b/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
@@ -953,13 +953,22 @@ public static partial class Gen5SpirvTranslator
                 case "VPkMulF16":
                 case "VPkMinF16":
                 case "VPkMaxF16":
-                case "VPkFmaF16":
                     if (!TryEmitPackedF16(instruction, out result, out error))
                     {
                         return false;
                     }
 
                     break;
+                case "VPkFmaF16":
+                    // Deliberately loud: a fused f16 FMA rounds the product+add once,
+                    // whereas doing the multiply-add in f32 and rounding to f16 at the
+                    // end double-rounds. Concrete miss: fma(0x4100, 0x7522, 0x04EA) is
+                    // 0x7A6B fused but 0x7A6A via f32. Exact emulation (round-to-odd
+                    // f32 product then RNE pack) is a planned follow-up slice.
+                    error =
+                        $"unsupported vop3p opcode {instruction.Opcode} " +
+                        "(fused f16 FMA requires single-rounding; deferred to a later slice)";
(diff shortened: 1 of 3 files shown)


kostyaff (changes requested):
Thanks for keeping this focused and for making unsupported paths fail loudly.
I checked the gfx10 prefix, opcodes, source/modifier fields, and ran all five
operations through the accepted ShaderDump Decode -> IR -> vertex/compute
SPIR-V pipeline. Those parts look good, and the full Release build is clean.

I found a blocking correctness issue in the packed-FP16 lowering. `VPkFmaF16`
is emitted as GLSL.std.450 f32 `Fma` followed by `PackHalf2x16`, which introduces
an intermediate f32 rounding before the FP16 result. A concrete FP16 case is:

```text
a = 0x4100 (2.5)
b = 0x7522
c = 0x04EA
FP16 fused result:             0x7A6B
current f32 Fma -> FP16 pack:  0x7A6A
```

The generated module contains GLSL.std.450 `Fma` (50) and `PackHalf2x16` (58),
with no Float16 capability or 16-bit float type, so this is the path produced
by the PR rather than only a numerical analogy.

Two related semantics also need resolving before the five-op slice is treated
as correct:

- GLSL.std.450 `FMin`/`FMax` leave NaN operand selection
  implementation-dependent, while LLVM models these AMD opcodes as
  `fminnum_like`/`fmaxnum_like`.
- `UnpackHalf2x16` may preserve or flush subnormals by implementation, while
  gfx10 exposes FP16 rounding and input/output denorm modes.

Could you either implement and test the exact packed-FP16 semantics, or narrow
this slice so the operations whose semantics are not yet exact continue to
fail loudly? Permanent synthetic regressions should include the FMA case above,
one-NaN min/max cases, denorm behavior, and non-default modifier fields.

No game or firmware data is needed for these tests.

333cra:
Thanks, this was the right call. The slice has been narrowed and fixed rather than left approximate. Pushed as a follow-up commit on top of the rebase onto current `main`.

**Exact conversions.** `UnpackHalf2x16`/`PackHalf2x16` are gone. f16<->f32 is now done with explicit integer sequences (`EmitHalfToFloat`/`EmitFloatToHalf`): exact widening with subnormal normalisation, and narrowing with round-to-nearest-even, overflow-to-Inf and NaN/Inf handling. That removes the dependency on implementation-defined denorm/rounding modes the review flagged, and it is the prerequisite that makes every remaining op exact.

**add / mul kept, and bit-exact.** With exact conversions, computing in f32 then rounding to f16 is bit-exact to a true f16 op: a f16 product fits in f32 with no rounding, and for the sum the double-rounding is innocuous because the intermediate precision satisfies 24 >= 2*11 + 2. Both were checked against a `System.Half` reference over 8M random half pairs with zero mismatches.

**min / max kept, semantics fixed.** They now match `fminnum_like`/`fmaxnum_like` (per `VOP3PInstructions.td`; `fminnum_like` = {`fminnum_ieee`, `fminnum`} per `AMDGPUInstructions.td`): a NaN operand returns the other operand, both-NaN returns a NaN, and the numeric case uses an ordered compare. The ordering of -0 and +0 is unspecified under these opcodes (that is `fminimum`/`fmaximum`, which is not what these map to), so no signed-zero special case is added. GLSL `FMin`/`FMax` are no longer used here.

**v_pk_fma_f16 fails loudly again.** A fused f16 FMA rounds once; an f32 multiply-add then pack double-rounds. The exact case from the review, a=0x4100 (2.5), b=0x7522, c=0x04EA, gives 0x7A6B fused versus 0x7A6A via f32, and this is now pinned as an emit-time failure with that value in a comment. Exact fused emulation (round-to-odd f32 product then RNE pack) is a planned follow-up slice.

**Regressions.** ShaderDump gained an expectation model (`Translates` / `DecodeFails` / `EmitFails`). New programs cover the four supported ops, non-default `op_sel`/`op_sel_hi`/`neg_lo`/`neg_hi`, the FMA case above (`EmitFails`), and clamp (`EmitFails`). Because GPU execution is out of scope for the tool, numeric correctness is proven separately by a C# mirror of the exact integer conversion/rounding logic validated exhaustively against `System.Half` (all 65536 half values widened, 20M+ f32 narrowed, full round-trip identity), and every emitted module is validated with `spirv-val` (see the env note below). Release build is clean, 0 warnings.

A couple of notes:

- The emitter stamps a SPIR-V 1.5 header, so validation is `spirv-val --target-env vulkan1.2` (vulkan1.1 rejects the 1.5 binary version outright, before any semantic check). Let me know if the project standardises on a specific env for this.
- ShaderDump does not build against current `main` on its own: #156 added a parameter to `TryCompileVertexShader` and the reflection invoke still passes the old count, which #166 fixes. This PR deliberately does not touch that invoke to stay single-topic, so the new regressions run once #166 lands (or with its one-line fix applied). Happy to reorder if you'd prefer.

Following slices remain as announced: inline/packed constants (first), packed integer ops, then `v_fma_mix*`.

par274:
need rebase

kostyaff:
Thanks for the thorough rework — all three issues from the earlier review are resolved on `1d58109`:

1. **FMA double-rounding**: `VPkFmaF16` now fails loudly with the exact counterexample (`0x4100 * 0x7522 + 0x04EA → 0x7A6B fused vs 0x7A6A f32`). Correct call to defer exact fused emulation to a follow-up slice.

2. **FMin/FMax NaN selection**: GLSL `FMin`/`FMax` replaced with `fminnum_like`/`fmaxnum_like` via `IsNan` + `Select` + ordered compare. NaN → other operand, both-NaN → NaN, no signed-zero special case (correct, these are not `fminimum`/`fmaximum`).

3. **UnpackHalf2x16 denorm/rounding**: GLSL pack/unpack replaced with exact integer sequences (`EmitHalfToFloat`/`EmitFloatToHalf`). Subnormal normalisation via `FindUMsb`, RNE narrowing with overflow-to-Inf and NaN/Inf handling. Validated exhaustively against `System.Half`.

The double-rounding argument for add/mul is sound: f32 mantissa (24 bits) ≥ 2×11+2, and a f16 product fits f32 exactly, so a single rounding at the pack is bit-exact.

The ShaderDump expectation model (`Translates`/`DecodeFails`/`EmitFails`) and the four new VOP3P test programs are a nice addition — they pin the supported ops, the modifier paths, and the deliberately-unsupported FMA/clamp cases.

No new findings. Happy to approve once the rebase lands.

333cra:
Rebased onto current `main`. The only conflict was in `ShaderDump/Program.cs`: the new `mrt*` programs and the pixel stage from #166 now sit alongside the `Expect` model (Translates / DecodeFails / EmitFails) this PR adds. Resolved by keeping both. With #166 merged, ShaderDump also builds and runs standalone again, so the packed regressions run as part of the normal sweep.

Re-verified after the rebase: full Release build clean (0 warnings), all 12 programs behave as expected, and all 27 emitted modules pass `spirv-val --target-env vulkan1.2`.

kostyaff (approved):
Reviewed the rebased head `8bd5969`. The rebase conflict was confined to `ShaderDump/Program.cs` (the new `mrt*` programs and the pixel stage from #166 now sit alongside the `Expect` model), and the review-response commit is unchanged from the previously reviewed `1d58109`.

All three issues from the original review remain resolved:

1. **FMA double-rounding**: `VPkFmaF16` fails loudly with the exact counterexample (`0x4100 * 0x7522 + 0x04EA → 0x7A6B fused vs 0x7A6A f32`).
2. **FMin/FMax NaN selection**: `fminnum_like`/`fmaxnum_like` via `IsNan` + `Select` + ordered compare.
3. **UnpackHalf2x16 denorm/rounding**: exact integer sequences (`EmitHalfToFloat`/`EmitFloatToHalf`), validated exhaustively against `System.Half`.

The double-rounding argument for add/mul is sound. The ShaderDump expectation model and VOP3P test programs are intact. No new findings on the rebase. Looks good to merge.

Spooks4576:

Tested `agc-vop3p-packed-f16` @ `786084a8` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-145

333cra:
rebased

PR #149, opened 2026-07-14
[AGC/Vulkan] Support multiple render targets

Dafenxz0:
## Summary

This PR completes Multiple Render Targets (MRT) end to end for translated AGC graphics draws and incorporates the full correctness review performed after the first implementation.

A guest draw is translated once and executed once against all active Vulkan color attachments. Sparse guest slots are compacted into dense host locations, output scalar types match the final Vulkan formats, and every attachment receives its own write mask and blend state.

The branch is rebased onto the current `main` and the change is split into three reviewable commits:

1. `[AGC] Support multiple typed pixel outputs`
2. `[Vulkan] Execute translated draws with multiple color attachments`
3. `[ShaderDump] Add MRT edge-case coverage`

## Why this matters

Modern 3D renderers commonly build a G-buffer in one draw (for example color, normals, material data, IDs, and motion vectors). Replaying the draw once per target is not equivalent because atomics, storage writes, counters, and other shader side effects would execute multiple times. This implementation therefore uses one Vulkan render pass and one draw with multiple color attachments.

### 1. Partial float exports emitted invalid SPIR-V

Fixed. Disabled components no longer inject a `uint` ID into a `vec4<float>` composite. They load the correctly typed previous component from the output variable (`float`, `uint`, or `int`).

This also preserves earlier values when the same MRT receives multiple partial exports. Pixel outputs are initialized once, and a later `exp` changes only its enabled components.

### 2. Export masks could overwrite disabled attachment channels

Fixed. AGC computes the union of `Gen5ExportControl.EnableMask` values for each guest target and intersects it with `CB_TARGET_MASK`:

`effectiveWriteMask = cbTargetWriteMask & shaderExportMask`

The effective value becomes the Vulkan `ColorWriteMask`, so channels absent from the shader export are not written to the attachment.

### 3. `independentBlend` was not enabled

Fixed. Device creation now queries and enables `PhysicalDeviceFeatures.IndependentBlend` when supported. The presenter records that capability and rejects an MRT layout with different attachment blend states when the selected device does not support it.

A real Vulkan test uses two float attachments with blending disabled on MRT0 and SrcAlpha/OneMinusSrcAlpha on MRT1.

### 4. MRT overload failed without `renderState`

Fixed. A single/default blend state is expanded to the number of targets before the draw is queued. This preserves the old single-target shorthand and makes the optional MRT API valid.

### 5. Transparent-fill workaround modified every MRT based only on MRT0

Fixed. The workaround is applied only when every attachment uses the exact premultiplied transparent-fill pattern. Mixed MRT blend layouts are left unchanged.

### 6. Sparse/incompatible host locations were accepted

Fixed. Pixel output host locations must now be dense in `0..N-1`, in addition to being unique. A synthetic negative test verifies that locations 0 and 7 for two outputs are rejected.

### 7. SPIR-V output kind and Vulkan format were decoded independently

Fixed. `TryDecodeRenderTargetFormat` now returns the final Vulkan `Format` together with its required `Gen5PixelOutputKind`. Both AGC shader compilation and Vulkan image creation use that shared result.

Integer attachments are explicitly rejected when blending is enabled.

### 8. Device attachment limits and format features were not checked

Fixed. The selected device's `MaxColorAttachments` is stored and enforced. Every final format is queried for `FormatFeatureFlags.ColorAttachmentBit` in optimal tiling before image/render-pass creation.

The guest/API ceiling remains eight attachments, but the lower real device limit wins.

### 9. Incompatible guest images could be destroyed while in flight

Fixed. Submission capacity is collected before target resolution. If an address must be recreated because its dimensions, mip count, or format changed, pending presentation work and all guest submissions are completed before destroying the old image/views/framebuffer/render pass/memory.

This is deliberately a correctness-first synchronization point only on incompatible recreation; normal same-layout reuse remains asynchronous.

### 10. Failure-path and malformed-draw behavior

The presenter rejects:

- zero or more than eight targets
- mismatched target dimensions
- duplicate/aliased target addresses
- unsupported target format/number-type combinations
- formats without Vulkan color-attachment support
- feedback loops against any MRT slot, not only MRT0
- mismatched target/blend counts
- integer attachments with blending enabled
- distinct blend layouts without `independentBlend`
- target counts above the selected device limit

Transient MRT framebuffer/render-pass objects are still transferred to deferred submission resources only after successful creation and are destroyed after the fence. Pre-transfer failure cleanup remains local.

### AGC and SPIR-V

- Collect every active color target with a matching shader export.
- Sort by guest slot and compact sparse slots (for example MRT0/MRT3/MRT6 -> locations 0/1/2).
- Emit typed `vec4`, `uvec4`, or `ivec4` fragment outputs.
- Preserve the guest slot separately from the host location.
- Include guest slot, host location, and scalar kind in the SPIR-V cache key.
- Retain the old single-output compiler overload for compatibility.

### Vulkan

- Carry all target descriptors and per-target blend/write state in one work item.
- Create one attachment description/reference/view per color target.
- Build one `VkPipelineColorBlendAttachmentState` per attachment.
- Include the full format and blend layouts in the pipeline cache key.
- Transition every image to `ColorAttachmentOptimal`, execute one draw, then transition every image to `ShaderReadOnlyOptimal`.
- Publish and diagnose every completed target.
- Keep transient MRT render-pass/framebuffer resources alive until their submission fence completes.

## Permanent synthetic coverage

`SharpEmu.Tools.ShaderDump` now covers:

- sparse mixed-type MRT0/MRT3/MRT6 (`float`, `uint`, `sint`)
- float export with partial RG enable mask
- two partial exports (RG followed by BA) to the same MRT
- two float MRT outputs for independent-blend execution
- all eight color attachments
- rejection of non-dense host locations

The GitHub workflow now runs the synthetic shader suite after the Release build, so these cases are no longer manual-only.

## Validation performed

- `dotnet build SharpEmu.slnx -c Release --no-restore`: 0 warnings, 0 errors.
- `SharpEmu.Tools.ShaderDump -c Release`: all programs and expected failures pass.
- Official Khronos `spirv-val --target-env vulkan1.3`: passes for every generated MRT pixel module (`mrt`, `mrt-float2`, `mrt8`, `mrt-partial`, and `mrt-partial-merge`).
- NVIDIA GeForce RTX 5080 with `SHARPEMU_VK_VALIDATION=1`: no validation-layer errors.
- Real runtime scenarios passed:
  - three mixed-type sparse attachments
  - MRT overload without `renderState`
  - two attachments with different blend states
  - eight simultaneous attachments
  - two partial exports to one target
  - partial RG write after a full RGBA write
  - incompatible target recreation while prior work can still be in flight
  - mismatched dimensions and duplicate addresses
  - feedback loop against MRT1
  - unsupported target format
  - integer attachment with blending enabled
- Partial-write readback changed `19334C66` to `CCB24C66`; B/A (`4C66`) remained intact while only R/G changed.

## Current boundaries

- Color MRT only; depth/stencil is intentionally a separate project.
- Single-sample rendering; MSAA/resolve is not introduced here.
- Dual-source blending is not introduced here.
- Attachments in one draw must have matching dimensions.
- Render passes and framebuffers remain transient per MRT submission. Caching them is intentionally deferred to a performance-only follow-up after correctness lands; invalidation depends on guest-image recreation and should not be mixed into this fix.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -100,6 +100,9 @@ jobs:
       - name: Build solution
         run: dotnet build SharpEmu.slnx -c Release --no-restore
 
+      - name: Validate synthetic shaders
+        run: dotnet run --project tools/SharpEmu.Tools.ShaderDump/SharpEmu.Tools.ShaderDump.csproj -c Release -- artifacts/shader-dump
+
       - name: Publish win-x64 CLI
         run: dotnet publish src/SharpEmu.CLI/SharpEmu.CLI.csproj -c Release -r win-x64 --self-contained true --no-restore -p:PublishDir="${env:PUBLISH_DIR}"
 
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -157,7 +157,7 @@ public static class AgcExports
     private static readonly HashSet<uint> _tracedSubmittedDrawOpcodes = new();
     private static readonly Dictionary<(ulong Ps, ulong State, Gen5PixelOutputKind Output), byte[]> _pixelSpirvCache = new();
     private static readonly Dictionary<
-        (ulong Es, ulong EsState, ulong Ps, ulong PsState, Gen5PixelOutputKind Output, uint Attributes),
+        (ulong Es, ulong EsState, ulong Ps, ulong PsState, string OutputLayout, uint Attributes),
         (byte[] Vertex, byte[] Pixel)> _graphicsSpirvCache = new();
     private static readonly Dictionary<
         (ulong Cs, ulong State, uint LocalX, uint LocalY, uint LocalZ),
@@ -3329,12 +3329,13 @@ public static class AgcExports
                     textures,
                     globalMemoryBuffers,
                     translatedDraw.AttributeCount,
-                    new VulkanGuestRenderTarget(
-                        firstTarget.Address,
-                        firstTarget.Width,
-                            firstTarget.Height,
-                            firstTarget.Format,
-                            firstTarget.NumberType),
+                    translatedDraw.RenderTargets.Select(target =>
+                        new VulkanGuestRenderTarget(
+                            target.Address,
+                            target.Width,
+                            target.Height,
+                            target.Format,
+                            target.NumberType)).ToArray(),
                         translatedDraw.VertexSpirv,
                         translatedDraw.VertexCount,
                         translatedDraw.InstanceCount,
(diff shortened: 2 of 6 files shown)


par274:
need rebase

PR #153, opened 2026-07-14
Add Brazilian Portuguese localization

brbrhuehue-matrix:
Adds a Brazilian Portuguese (pt-BR) translation for the launcher and emulator UI.

Translated:
- Library
- Options
- Console
- Launch status
- Discord status
- Dialogs


Gih-pt:
If you can, please do it for European Portuguese as well.

PR #89, opened 2026-07-12
feat: Improve PS5 (Prospero) SELF loader support in eboot.bin parsing

Kaotic:
### Description

Improved the SELF (Signed Executable and Linkable Format) loader to properly support both PS4 (Orbis) and PS5 (Prospero) eboot.bin files.

### Changes

- Updated `SelfHeader` struct to the modern unified layout used by both PS4 and PS5.
- Switched magic number reading to `LittleEndian` (required for correct detection of `0xEEF51454` / Prospero magic).
- Removed outdated PS4-specific layout checks (`HasKnownLayout`, `_identX` fields, and `0x22` unknown value).
- Improved error messages and validation logic for better debugging.

### Testing

- Successfully parses PS5 eboot.bin files (magic `0xEEF51454`).
- Maintains backward compatibility with PS4 Orbis SELF files.
- No breaking changes to existing ELF loading logic.

PR #157, opened 2026-07-14
loader: support ps5 SELF and validate ELF signatures

anesr5:
Add support for loading Prospero/PS5 SELF executables.
## Reference

- [etaHEN `dumper_utils.cpp`](https://github.com/etaHEN/etaHEN/blob/main/Source%20Code/daemon/source/dumper_utils.cpp#L47)

### DIFF
--- a/src/SharpEmu.Core/Loader/SelfLoader.cs
+++ b/src/SharpEmu.Core/Loader/SelfLoader.cs
@@ -17,7 +17,9 @@ namespace SharpEmu.Core.Loader;
 public sealed class SelfLoader : ISelfLoader
 {
     private static readonly SharpEmuLogger Log = SharpEmuLog.For("Loader");
-    private const uint SelfMagic = 0x4F153D1D;
+    private const uint ElfMagic = 0x7F454C46;
+    private const uint Ps4SelfMagic  = 0x4F153D1D;
+    private const uint Ps5SelfMagic  = 0x5414F5EE;
     private const ulong SelfSegmentFlag = 0x800;
     private const int PageSize = 0x1000;
     private const ulong ImportStubBaseAddress = 0x0000_7000_0000_0000UL;
@@ -323,7 +325,8 @@ public sealed class SelfLoader : ISelfLoader
             throw new InvalidDataException("Input image is too small to contain an ELF header.");
         }
 
-        if (imageData.Length >= sizeof(uint) && BinaryPrimitives.ReadUInt32BigEndian(imageData[..sizeof(uint)]) == SelfMagic)
+        var magic = BinaryPrimitives.ReadUInt32BigEndian(imageData[..sizeof(uint)]);
+        if (magic is Ps4SelfMagic or Ps5SelfMagic)
         {
             var selfHeader = ReadUnmanaged<SelfHeader>(imageData, 0);
             if (!selfHeader.HasKnownLayout || selfHeader.Unknown != 0x22)
@@ -343,6 +346,12 @@ public sealed class SelfLoader : ISelfLoader
             }
 
             return new LoadContext(IsSelf: true, elfOffset, selfHeader.FileSize, segments);
+        }
+
+        if (magic != ElfMagic)
+        {
+            throw new InvalidDataException(
+                $"Unsupported executable signature 0x{magic:X8}");
         }
 
         return new LoadContext(IsSelf: false, ElfOffset: 0, SelfFileSize: 0, Array.Empty<SelfSegment>());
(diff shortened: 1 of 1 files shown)


Chelu97:
Tested this on my Demon's Souls dump (PPSA01341 v01.005) and it checks out.

On current `main`, loading `eboot.bin` dies with `Input does not contain a valid ELF header` because the PS5 magic `54 14 F5 EE` isn't recognized. With `0x5414F5EE` accepted (I tried an equivalent local patch), the eboot and the `sce_module` prx files (`libc.prx`, `libSceNpCppWebApi.prx`) all load fine: segments map through the SELF segment table, ~173k relocations get processed, 1173 import stubs resolve, and execution reaches `libc.prx` module init before crashing in the CPU backend — which is expected, that's separate work.

A few things I saw in that dump that back this change:

- The embedded ELF sits at `0x1A0` = `0x20 + segment_count(12) * 0x20`, exactly where `ParseLayout` computes it.
- Ident bytes 4..11 (`00 01 01 12 01 01 00 00`) match the PS4 layout `HasKnownLayout` already validates — the 4-byte magic really is the only difference.
- All 12 segments have `compressed_size == decompressed_size` and no encrypted (`& 0x2`) or compressed (`& 0x8`) flags, so the payload is plaintext and nothing here needs decryption.

While checking this I ended up writing a small xunit fixture suite for `SelfLoader` (synthetic in-memory PS5/PS4/raw-ELF images, encrypted-segment error path, ident validation). Happy to open it as a follow-up PR once this lands so these paths stay covered.

Kaotic:
> Tested this on my Demon's Souls dump (PPSA01341 v01.005) and it checks out.
> 
> On current `main`, loading `eboot.bin` dies with `Input does not contain a valid ELF header` because the PS5 magic `54 14 F5 EE` isn't recognized. With `0x5414F5EE` accepted (I tried an equivalent local patch), the eboot and the `sce_module` prx files (`libc.prx`, `libSceNpCppWebApi.prx`) all load fine: segments map through the SELF segment table, ~173k relocations get processed, 1173 import stubs resolve, and execution reaches `libc.prx` module init before crashing in the CPU backend — which is expected, that's separate work.
> 
> A few things I saw in that dump that back this change:
> 
>     * The embedded ELF sits at `0x1A0` = `0x20 + segment_count(12) * 0x20`, exactly where `ParseLayout` computes it.
> 
>     * Ident bytes 4..11 (`00 01 01 12 01 01 00 00`) match the PS4 layout `HasKnownLayout` already validates — the 4-byte magic really is the only difference.
> 
>     * All 12 segments have `compressed_size == decompressed_size` and no encrypted (`& 0x2`) or compressed (`& 0x8`) flags, so the payload is plaintext and nothing here needs decryption.
> 
> 
> While checking this I ended up writing a small xunit fixture suite for `SelfLoader` (synthetic in-memory PS5/PS4/raw-ELF images, encrypted-segment error path, ident validation). Happy to open it as a follow-up PR once this lands so these paths stay covered.

I have do that in https://github.com/par274/sharpemu/pull/89 and for loading ELF files :

In SelfLoader -> ParseLayout function
```
uint magic = BinaryPrimitives.ReadUInt32LittleEndian(imageData);
if (magic == 0x464C457F) // ELF magic
{
    return new LoadContext(
        IsSelf: false,
        ElfOffset: 0,
        SelfFileSize: (ulong)imageData.Length,
        SelfSegments: Array.Empty<SelfSegment>()
    );
}
```

anesr5:
> Tested this on my Demon's Souls dump (PPSA01341 v01.005) and it checks out.
> 
> On current `main`, loading `eboot.bin` dies with `Input does not contain a valid ELF header` because the PS5 magic `54 14 F5 EE` isn't recognized. With `0x5414F5EE` accepted (I tried an equivalent local patch), the eboot and the `sce_module` prx files (`libc.prx`, `libSceNpCppWebApi.prx`) all load fine: segments map through the SELF segment table, ~173k relocations get processed, 1173 import stubs resolve, and execution reaches `libc.prx` module init before crashing in the CPU backend — which is expected, that's separate work.
> 
> A few things I saw in that dump that back this change:
> 
>     * The embedded ELF sits at `0x1A0` = `0x20 + segment_count(12) * 0x20`, exactly where `ParseLayout` computes it.
> 
>     * Ident bytes 4..11 (`00 01 01 12 01 01 00 00`) match the PS4 layout `HasKnownLayout` already validates — the 4-byte magic really is the only difference.
> 
>     * All 12 segments have `compressed_size == decompressed_size` and no encrypted (`& 0x2`) or compressed (`& 0x8`) flags, so the payload is plaintext and nothing here needs decryption.
> 
> 
> While checking this I ended up writing a small xunit fixture suite for `SelfLoader` (synthetic in-memory PS5/PS4/raw-ELF images, encrypted-segment error path, ident validation). Happy to open it as a follow-up PR once this lands so these paths stay covered.

Thanks for testing this and for the details.

I hit the same issue on a Space Adventure Cobra dump (PPSA17337), which is what led me to this fix in the first place. Once the PS5 SELF magic was accepted, it started loading correctly.

PR #158, opened 2026-07-14
[Vulkan] Add safe guest buffer allocation cache

Dafenxz0:
## Summary

This PR adds a **safe, opt-in Vulkan allocation cache** for translated guest storage buffers.

It no longer attempts to infer CPU/GPU ownership from byte equality or preserve GPU-produced contents without readback. CPU memory remains authoritative and the current CPU snapshot is uploaded on every use, matching the pre-PR correctness model.

Allocation reuse is disabled by default and can be tested with:

`SHARPEMU_PERSISTENT_GUEST_BUFFERS=1`

## Why the scope changed

The previous implementation compared incoming bytes with `CpuShadow` and skipped uploads when they matched. That is ambiguous:

1. CPU writes 0
2. GPU writes 37
3. CPU intentionally writes 0 again

Byte equality cannot distinguish step 3 from a stale CPU snapshot, so the cache incorrectly retained 37. Native guest writes also bypass `ICpuMemory.TryWrite`, and GPU-to-CPU readback does not exist yet.

Persistent GPU content therefore cannot be enabled safely until page-based dirty tracking, native write-fault integration, mapping generations, and readback are implemented together.

## What changes

- removes `CpuShadow` and `CpuKnown`;
- removes byte equality as an ownership rule;
- removes aggressive overlap merging and active-binding rebasing;
- removes global GPU waits from buffer updates;
- always uploads the current CPU snapshot before use;
- requires explicit access on `VulkanGuestMemoryBuffer`;
- classifies AGC buffer loads as Read, stores as Write, and atomics as ReadWrite;
- makes index-buffer guest addresses mandatory;
- never uses address 0 as a persistent key;
- keeps vertex, index, read-only, and unknown-access buffers transient;
- enables allocation caching only for confirmed storage writers and only behind the opt-in switch;
- uses resource renaming when an exact allocation is still in flight;
- indexes reusable allocations by exact `(address, size)` instead of scanning/merging ranges;
- bounds the cache by bytes (64 MiB), not object count;
- applies host-to-consumer barriers to the actual binding offset and size.

## Correctness model

This PR caches **Vulkan allocations**, not GPU ownership or GPU-produced data.

```
guest CPU snapshot -> full upload -> Vulkan work
```

A GPU write is valid for that submitted work, but a later submission receives the current CPU snapshot again. This deliberately preserves existing emulator behavior and avoids regressions until a complete coherency contract exists.

## Tests

The cache test now verifies:

1. CPU -> GPU(37) -> CPU rewrites the same 0 -> Vulkan observes 0;
2. an in-flight allocation is renamed instead of globally waiting;
3. read-only buffers remain transient;
4. guest address 0 remains transient;
5. overlapping logical ranges are not merged;
6. cached bytes are reduced to the configured memory budget.

Local validation:

- Release build: 0 warnings, 0 errors;
- SharpEmu.Libs.Tests: 18/18 passed;
- GuestBufferCache regression test: passed;
- ShaderDump: all programs behaved as expected.

## Deliberately deferred

True persistent GPU contents require a separate, cross-layer coherency change:

- page-based CPU-dirty/GPU-dirty tracking;
- hooks in `ICpuMemory.TryWrite`;
- native guest write-fault handling;
- mapping identities/generations;
- per-submission GPU-write completion;
- GPU-to-CPU readback before CPU access;
- resource-specific fences instead of global waits.

Those pieces must land as one coherent contract rather than being approximated through byte comparisons.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -103,6 +103,9 @@ jobs:
       - name: Validate synthetic shaders
         run: dotnet run --project tools/SharpEmu.Tools.ShaderDump/SharpEmu.Tools.ShaderDump.csproj -c Release -- artifacts/shader-dump
 
+      - name: Test guest buffer allocation cache
+        run: dotnet run --project tests/SharpEmu.Tests.GuestBufferCache/SharpEmu.Tests.GuestBufferCache.csproj -c Release --no-build
+
       - name: Publish win-x64 CLI
         run: dotnet publish src/SharpEmu.CLI/SharpEmu.CLI.csproj -c Release -r win-x64 --self-contained true --no-restore -p:PublishDir="${env:PUBLISH_DIR}"
 
--- a/SharpEmu.slnx
+++ b/SharpEmu.slnx
@@ -14,5 +14,6 @@ SPDX-License-Identifier: GPL-2.0-or-later
   </Folder>
   <Folder Name="/tests/">
     <Project Path="tests/SharpEmu.Libs.Tests/SharpEmu.Libs.Tests.csproj" />
+    <Project Path="tests/SharpEmu.Tests.GuestBufferCache/SharpEmu.Tests.GuestBufferCache.csproj" />
   </Folder>
 </Solution>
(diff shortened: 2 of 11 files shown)


par274:
need rabase

par274:
resim

there seem to be regressions in almost every game

par274:
no

Dafenxz0:
Can i ask why not,is only to have a better idea to make the Next pulls

par274:
> Can i ask why not,is only to have a better idea to make the Next pulls

Buffer caching isn't accurate emulation. There are unusually large differences in frame rates, like 60 FPS dropping to 10. Please test PRs on actual titles first.

PR #164, opened 2026-07-14
[compat] UE/kernel/GPU HLE fixes

YkPhysics:
Broad compatibility pass that gets Unreal Engine (and other) PS5 titles
past early boot, thread bring-up, and initial GPU submission. This is the
local working set verbatim (verified: Dreamcore reaches engine-initialized
and boots; games are NOT stuck on the RHI/render threads).

Threading/scheduler: guest scheduler + cooperative cond_wait/signal/
broadcast + mutex blocking; POSIX pthread rwlock family; self-owned
recursive-mutex trylock livelock fix; native guest access-violation
handling; opt-in thread-state watchdog (SHARPEMU_DUMP_THREADS).

GPU/AGC: command-buffer support (draw-indirect, predication, DMA, resource
patch, owner registration); extra Gen5 SPIR-V/scalar ops (S_FF0, S_GETREG,
V_ALIGNBIT, V_CVT_PKNORM, buffer-store/sample variants).

Memory: flexible/named mapping, mprotect/mtypeprotect, batch map, virtual-
memory plumbing.

Boot unblockers: vblank status, PlayGo chunks, AudioOut2, Pad trigger,
Http2, Ampr, libc-internal, usleep, event-queue + Apr tweaks.


YkPhysics:
image
image
image
image
image

YkPhysics:
image

YkPhysics:
Currently fixing Dreaming Sarah.

Deeptanshuu:
Tested this branch against a Dreaming Sarah dump (PPSA02929, own-console) on an RTX 3060 laptop since the game came up here — sharing findings in case they save you time. Baseline: on current `main` the title boots to the menu and is playable; on this branch (head `c2b572a`) it black-screens after the logo splashes.

Three regressions vs `main` found along the way:

1. **Pad user id** — `PadExports.PrimaryUserId` is `1` here (`main`: `1000`). `sceUserServiceGetInitialUser` still returns 1000, so `scePadOpen` fails with `0x80920007` and the title busy-waits on `scePadReadState` forever (~67k failing calls in 90 s across two threads). Restoring 1000 fixes input.

2. **Vulkan device selection** — `SelectPhysicalDevice` takes the first device with a graphics+present queue, where `main` scores devices (discrete preferred, integrated last). On hybrid laptops the Intel iGPU enumerates first and lacks `shaderInt64`, so the translator's capability warnings fire and translated shaders fail. Restoring the discrete preference clears it.

3. **Vblank pump** — the autonomous 60 Hz vblank pump from `main` is gone. The per-query `GetVblankStatus` increment keeps polling engines (UE) alive, but engines that register vblank events and block on the equeue only ever see the single synthetic edge from `AddVblankEvent`, and stop pacing their renderer. Restoring the pump alongside the new status path fixed vblank-event delivery here.

With all three applied the title still black-screens; as far as I could trace the remaining cause: after the splash the engine still does its full per-frame scene-draw setup (`DcbSetIndexBuffer` with the big shared index buffer, `CreateInterpolantMapping` + `CreatePrimState` with valid heap shaders) and then never issues `sceAgcDcbDrawIndexOffset` — instrumenting the builder's failure paths shows the call never happens at all (0 failed calls), so the game is skipping its scene pass rather than the builder rejecting it. Only the built-in fullscreen fill (`textures=[]`) plus the flip run each frame, which is what paints both display buffers black. The same title on `main` records ~6 `DcbDrawIndexOffset` draws per frame continuously. DCB parsing looks clean (nothing stuck in `GpuWaitRegistry`, `EventWrite`/`ReleaseMem` handling matches `main`), so presumably some completion/label the engine checks before its scene pass never fires on this branch.

The three fixes are on [`Deeptanshuu:fix/pad-userid-mismatch`](https://github.com/Deeptanshuu/sharpemu/tree/fix/pad-userid-mismatch) (stacked on this PR's head) if useful — happy to share the full AGC traces too.

Deeptanshuu:
Follow-up — found the actual root cause of the Dreaming Sarah black screen, and it is not GPU-side. Ran this branch with `SHARPEMU_DUMP_THREADS=1` past the splash; every dump over ~40 s shows the same picture:

```
name='background loader'                 state=Ready    block=none                    imports=6
name='(BackgroundLoader) P.Worker 0-14'  state=Blocked  block=pthread_cond_timedwait  imports=6
name='run_c2init'                        state=Blocked  block=pthread_cond_timedwait
name='spi_main_thread'                   state=Running  block=none
name='AudioOutThread'                    state=Running  block=none
```

The `background loader` dispatcher thread sits permanently **Ready** — it executed 6 imports after spawning and is never scheduled again for the rest of the run. Its 15 pool workers are parked in `pthread_cond_timedwait` (also frozen at `imports=6`, so their timeouts never fire either) waiting for jobs it never dispatches, and `run_c2init` — the thread that prints "Loading 19 event sheets and 58 layouts" — waits on the loader. Net effect: on `main` that loading phase completes in ~2.9 s ("Done loading event sheets & layouts" → "Go to layout: M0T0", 108/108 background resource loads); on this branch the Done line never prints and 0 loads start. Spawn priorities for reference: background loader=690, spi_main_thread=701, workers=710, all affinity 0x7F.

So the renderer is actually behaving correctly — each frame the engine runs its full scene-pass setup and then has zero objects to draw because the layout never loaded, and the built-in fill + flip paints both display buffers black. My earlier GPU-side observations (no `sceAgcDcbDrawIndexOffset` calls, compositor with `textures=[]`) are all downstream of this one starved thread.

Looks like a ready-thread starvation case in the new guest scheduler — the continuously-running threads (`spi_main_thread`, `AudioOutThread`, `saveThreadProc`) never seem to yield in a way that lets the Ready thread on. Happy to run any instrumented build against this title if that helps.

Deeptanshuu:
Update: root cause confirmed and fixed — Dreaming Sarah now boots to the menu on this branch.

The starved `background loader` from my previous comment turned out to be a general scheduler liveness gap, not anything specific to that thread. Traced with `SHARPEMU_LOG_GUEST_THREADS=1`: **every dispatch in a 40 s run had `reason=pthread_create`** — no other Pump call site ever fired. Once the title stops creating threads, the remaining pump paths all go quiet: the primary guest parks host-side (so no `import_tick`), `WakeBlockedThreads` enqueues woken threads but nothing dispatches them, and the stall watchdog stays silent because the continuously running guests (`spi_main_thread`, `AudioOutThread`) keep import progress fresh. From that moment, any thread that blocks and is later woken sits `Ready` in the queue forever. The `background loader` was woken exactly once (`guest_threads.wake key=pthread_cond:0x8004A1DD8`) and never ran again; threads that "worked" earlier only ran because unrelated `pthread_create` calls happened to pump the queue.

Fix (35 lines, [`7256953`](https://github.com/Deeptanshuu/sharpemu/commit/7256953)): `WakeBlockedThreads` pumps after enqueuing, and the existing stall-watchdog thread pumps Ready threads on every 200 ms tick as a liveness backstop (which also fires expired timed waits via `WakeExpiredBlockedGuestThreads`). Also added thread names to the wake/discard traces under `SHARPEMU_LOG_GUEST_THREADS` — that's what made this findable.

Verified on this branch + the fixes from my earlier comment: "Done loading event sheets & layouts" completes in 0.5 s (previously never), "Go to layout: M0T0", 108/108 background resource loads and the same texture-upload profile as `main`, title reaches the menu.

Full stack is on [`Deeptanshuu:fix/pad-userid-mismatch`](https://github.com/Deeptanshuu/sharpemu/tree/fix/pad-userid-mismatch) (4 commits on this PR's head: pad user id, discrete-GPU selection, vblank pump, scheduler wake pump) — feel free to cherry-pick or fold them in however you prefer.

par274:
Dreaming Sarah boots but colors is incorrect.

resim

Deeptanshuu:
Two smaller findings after playing with the current head (`673c2e0`):

**Colors are close but not fully there.** Compared the same forest area against `main` (screenshot pixel stats, upper canopy band): blacks sit ~9 points lifted (p5 59→68) and high-end saturation drops ~30% (0.39→0.27). So the swapchain fix removed the big wash, but a milder second source remains — my guess is one of the offscreen canvas passes or the premultiplied blend path getting an extra encode somewhere.

**Stutter, and it's partly the wake-pump patch's fault (mine).** It fixes the liveness bug but crudely: every wake now runs `Pump`, which spawns a fresh host thread per resumed guest and ends with the `Thread.Sleep(1)` fairness nap paid in the waker's context. With vblank events at 60 Hz plus audio/streamer signals, that's hundreds of thread spawns a second and 1 ms naps injected into hot paths. Fine as a proof of the diagnosis, not as the real design — a persistent dispatcher (event-woken) and reusing a parked host thread per guest thread instead of spawn-per-resume would kill both costs. Didn't want to make that call on your architecture, so leaving it as a note.

par274:
need rebase

Spooks4576:

❌ **FAILED — gameplay regressions detected. Do not merge.**

Tested `ue-compat-upstream` @ `673c2e06` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ❌ failed |
| Dead Cells | ❌ failed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-164

Gih-pt:
Is this still being worked on?

PR #166, opened 2026-07-14
[Tools] Fix ShaderDump reflection invoke against new optional parameters

Deeptanshuu:
Follow-up to #111. `ShaderDump` currently crashes on `main` with `TargetParameterCountException`: `TryCompileVertexShader` gained an optional `scalarRegisterBufferIndex` parameter in #156, and reflection `Invoke` does not apply C# default parameter values, so the tool's 7-argument call no longer matches the 8-parameter method.

Rather than hard-coding the new default (which breaks again on the next added parameter), the compile calls now pad trailing optional parameters with `Type.Missing` under `BindingFlags.OptionalParamBinding`, so the runtime substitutes the declared defaults. If a translator entry point ever gains a new **required** parameter, the tool fails with a named error saying which method and parameter changed, instead of a bare crash.

Verified on `main` + this commit:
- all five programs behave as expected, exit code 0 (`fmac`, `muls`, `sopp-hints`, `exec` decode+emit both stages; `sopp-mode` pins its loud decode failure);
- all eight emitted blobs pass `spirv-val --target-env vulkan1.3`;
- the `exec-cs.spv` blob also re-verified numerically on real hardware via the executor in #127 (RTX 3060, all values bit-exact including the EXEC-masked store suppression).

This is also a live data point for the reflection-vs-`InternalsVisibleTo` question raised in #111 — reflection kept the emulator source untouched, but it means translator signature changes surface only when someone runs the tool. Happy to move the tools to `InternalsVisibleTo` (or the public `TryCreateState`/`TryEvaluate` entry points) in a follow-up if that's the preferred direction.

### DIFF
--- a/tools/SharpEmu.Tools.ShaderDump/Program.cs
+++ b/tools/SharpEmu.Tools.ShaderDump/Program.cs
@@ -201,8 +201,8 @@ foreach (var (name, expectTranslate, words) in testPrograms)
         null,
         null)!;
 
-    object?[] compileArgs = [state, evaluation, null, null, 0, -1, 0];
-    if ((bool)tryCompile.Invoke(null, compileArgs)!)
+    var compileArgs = PadWithDefaults(tryCompile, [state, evaluation, null, null]);
+    if ((bool)tryCompile.Invoke(null, BindingFlags.OptionalParamBinding, null, compileArgs, null)!)
     {
         var shader = compileArgs[2]!;
         var spirv = (byte[])shader.GetType().GetProperty("Spirv")!.GetValue(shader)!;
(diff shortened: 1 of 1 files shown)


kostyaff (approved):
Reviewed on the exact head `29e5892` against current `main`.

I reproduced the original `TargetParameterCountException` on `main`, then ran
the same ShaderDump command on this PR: all five programs behaved as expected,
all eight vertex/compute SPIR-V blobs were emitted, and the tool exited 0. The
full locked Release build also passes with 0 warnings and 0 errors.

I separately checked the reflection compatibility paths: the vertex method's
four declared defaults are padded with `Type.Missing`, the unchanged compute
signature receives no padding, and both a new required trailing parameter and
argument over-supply fail with clear method/parameter diagnostics.

Standards and spec both pass with no findings. This is a focused fix and looks
good to merge.

PR #169, opened 2026-07-14
[Json] Implement sce::Json::Value and String (construct / set / destroy)

Chelu97:
libSceJson so far only had the Initializer/MemAllocator setup path — the Value and String classes weren't implemented at all. Any Prospero title that actually builds a JSON tree calls Value's constructors and setters (and String), and since those imports resolved to nothing the guest faulted on the first call into them.

That's what #7 (Quake, PPSA01880) runs into. The imports it leaves unresolved right before the access violation are exactly these: Value's default ctor, set(bool/long/unsigned/double/ValueType/String), and String's ctor/dtor — Quake builds JSON to shape a web-API request.

The payload is modelled host-side keyed by the guest `this` pointer, the same handle-shadow approach Ngs2Exports already uses. I don't write into the guest object on purpose: these are usually stack-allocated with a layout I can't see from clean-room evidence, and writing a guessed layout can smash an adjacent stack canary — there's already a note about that exact hazard in the AudioOut2 context-param sizing. Constructors and setters follow the Itanium ABI and return `this` in rax, which holds whether the real setter returns void or Value&; the double setter reads xmm0 like the other float args here. I only bound the complete-object C1/D1 variants, and only the payload reachable through library methods — direct reads of the object bytes would need observed layout evidence and are out of scope.

On testing: I don't have Quake, so I couldn't run it end to end and I'm not claiming from here that it boots further. What I did verify is a clean Release build and a new xunit project (SharpEmu.Libs.Tests, along the lines discussed in #36) that drives every ctor/setter/dtor at the CpuContext level and checks the round-trip. There's also a registration test that runs a real ModuleManager over the whole SharpEmu.Libs assembly and confirms the nine NIDs from #7 now resolve to these handlers and dispatch cleanly instead of trapping — which also covers the change being purely additive, since a NID collision with anything existing would fail it. No existing code paths are touched.

So what I can stand behind is that the imports Quake faulted on now resolve and behave per the C++ ABI, with tests to back it. Whether that carries Quake past the crash needs someone with the game to confirm, and I'm happy to iterate on the next blocker.

CI doesn't run the tests yet — left that as your call. New files carry SPDX headers; no Sony code, keys, or assets.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -12,11 +12,14 @@ SPDX-License-Identifier: GPL-2.0-or-later
     <PackageVersion Include="Avalonia.Fonts.Inter" Version="11.3.18" />
     <PackageVersion Include="Avalonia.Themes.Fluent" Version="11.3.18" />
     <PackageVersion Include="Iced" Version="1.21.0" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
     <PackageVersion Include="Silk.NET.Vulkan" Version="2.23.0" />
     <PackageVersion Include="Silk.NET.Vulkan.Extensions.EXT" Version="2.23.0" />
     <PackageVersion Include="Silk.NET.Vulkan.Extensions.KHR" Version="2.23.0" />
     <PackageVersion Include="Silk.NET.Windowing" Version="2.23.0" />
     <!-- Transitive of Avalonia.Desktop; pinned to fix GHSA-xrw6-gwf8-vvr9 -->
     <PackageVersion Include="Tmds.DBus.Protocol" Version="0.21.3" />
+    <PackageVersion Include="xunit" Version="2.9.3" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.1" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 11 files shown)


Chelu97:
Update: I got hold of a copy of the game and tested this end to end (PPSA01880 v01.009.000).

The boot crash from #7 is gone with this branch. Zero unresolved Json imports, no access violation, and the game gets much further: kexAudio init completes, the whole shader program set loads (kexRHIShaderManagerGnmp), an AudioOut port opens (48 kHz float32, winmm) and it keeps pumping sceAudioOutOutput. I let it run for two minutes and killed it myself.

There is a new blocker further in, unrelated to Json: KexAudioThread sits in sceKernelWaitSema on a semaphore nothing ever signals, times out repeatedly and after ~45s the game bails with its own error dialog and abort(). Looks like whatever should be posting that semaphore isn't implemented yet — separate issue, happy to dig into that next.

Chelu97:
Correction to my note above — I dug into this properly and the semaphore was a red herring, so please ignore that part. Those WaitSema timeouts are just idle worker threads polling, and the semaphore does get signaled; the game isn't stuck there. It actually boots fully, presents a splash frame and starts the demo. Where it really stops is a mix of unimplemented PSN/WebAPI (sceNpAuthCreateRequest and friends return not-implemented, and one libSceJson Initializer callback, setGlobalNullAccessCallback, isn't there yet), then ~45s in it hits its own Sys_Error with an empty message and abort(). None of this changes the PR — the Value/String work is solid and the #7 crash is gone — just setting the record straight on where it actually stops now.

PR #188, opened 2026-07-14
GUI: Add code -1073741819 (Access Violation) as an emulation error

Randomuser8219:
There's currently games that crash on this access violation code due to emulation errors, so it'd make sense to add this error code as an emulation error (I think).


par274:
Make sense

PR #195, opened 2026-07-15
Fix large-title boot blockers across kernel, AGC, AJM, and VideoOut

Aspenini:
Fixes direct-memory queries, static TLS initialization, guest stack reservations, positional reads, and host fault handling. Adds missing AGC packets and exports, Gen5 shader image/buffer operations, additional VideoOut formats and image aliases, plus initial AJM decoding and AudioOut2 plumbing.

GTA V now reaches 1280×720 VideoOut initialization and continues through its frame/audio loop instead of stalling during early boot. Validated with a clean Release build and passing tests.

### DIFF
--- a/src/SharpEmu.Core/Cpu/CpuDispatcher.cs
+++ b/src/SharpEmu.Core/Cpu/CpuDispatcher.cs
@@ -26,10 +26,6 @@ public sealed class CpuDispatcher : ICpuDispatcher, IDisposable
     private const ulong StackSize = 0x0020_0000UL;
     private static readonly ulong TlsBaseAddress = OperatingSystem.IsWindows() ? 0x7FFE_0000_0000UL : 0x6FFE_0000_0000UL;
     private const ulong TlsSize = 0x0001_0000UL;
-    // The static TLS blocks live at negative offsets from the TCB (FreeBSD
-    // amd64 variant II). Keep every host in sync with GuestTlsTemplate's
-    // startup reservation; PS5 modules routinely reach beyond one host page.
-    private const ulong TlsPrefixSize = GuestTlsTemplate.StartupStaticTlsReservation;
     private static readonly ulong BootstrapStubBaseAddress = OperatingSystem.IsWindows() ? 0x7FFD_F000_0000UL : 0x6FFD_F000_0000UL;
     private static readonly ulong BootstrapPayloadBaseAddress = OperatingSystem.IsWindows() ? 0x7FFD_E000_0000UL : 0x6FFD_E000_0000UL;
     private static readonly ulong DynlibFallbackStubBaseAddress = OperatingSystem.IsWindows() ? 0x7FFD_D000_0000UL : 0x6FFD_D000_0000UL;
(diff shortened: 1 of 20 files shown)


kostyaff:
Reviewed on the exact head `cf6fd95` against current `main` (`9d88542`). 14 commits, 23 files.

This is a substantial boot-blocker fix and most of it looks correct — the static TLS variant 2 layout, guest thread stack reservation race fix, SSE4a INSERTQ decoding, hull shader registers, FLAT byte loads, 3D texture support, AJM instance/batch management, and `DirectMemoryQuery` FIND_NEXT semantics all check out. GTA V reaching 1280×720 VideoOut is great progress.

A few things worth addressing:

**1. INSERTQ upper 64 bits not zeroed** (`DirectExecutionBackend.Imports.cs`)
Per the AMD SSE4a spec, `INSERTQ` zeroes `DEST[127:64]`. The emulation writes only the low 64 bits at `destinationOffset` and leaves the high 64 bits unchanged. Guest code reading the full 128-bit XMM destination will see stale data. Fix: `WriteCtxU64(contextRecord, destinationOffset + sizeof(ulong), 0);`

**2. `RecordTextureAliasCopy` source layout hardcoded** (`VulkanVideoPresenter.cs`)
The source image barrier assumes `ShaderReadOnlyOptimal`, but the alias creation check allows aliasing when `source.InitialUploadPending` is true — in that case the source layout is `Undefined`, making the transition invalid per the Vulkan spec. Could trigger validation errors or undefined behavior.

**3. `IsAddressAdd` missing `VAddCoCiU32`** (`Gen5ShaderScalarEvaluator.cs`)
Flat-address base resolution checks `VAddI32`, `VAddcU32`, `VAddCoU32` but not `VAddCoCiU32` (opcode 0x128, which is decoded correctly in the translator). Shaders that chain 64-bit address arithmetic with carry-in would fail flat-address resolution. Fix: add `"VAddCoCiU32"` to the `IsAddressAdd` check.

**4. `KernelPreadUnderscore` missing `ObjectDisposedException`** (`KernelMemoryCompatExports.cs`)
The stream is retrieved from `_openFiles` under a lock, but the read happens outside the lock. If another thread closes the fd between lookup and read, `ObjectDisposedException` (inherits from `InvalidOperationException`, not `IOException`) is uncaught and would crash the emulator. Should add it to the catch filter.

**5. `BatchStart` unchecked `Contexts[contextId]` indexer** (`AjmExports.cs`)
The second access uses the raw indexer after a `ContainsKey` check in a separate lock acquisition. Currently safe since there is no destroy function, but use `TryGetValue` to be robust.

Lower-severity notes: INSERTQ `bitIndex+fieldLength > 64` returns false (guest crash instead of hardware-like wrapping), TLS `imageBase` fallback to 0 for low-loaded modules, `DumpGuestDisasmDiagnostics` called from VEH without try/catch, no AJM `InstanceDestroy` (unbounded growth across sessions), TLS fields overwritten on re-registration in `KernelModuleRegistry`.

No blocking issues — these can be addressed in follow-up. Thanks for the work on this.

par274:
need to revert back package lock files.

par274:
need rebase

Spooks4576:

❌ **FAILED — gameplay regressions detected. Do not merge.**

Tested `main` @ `b62f29eb` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ❌ failed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-195

PR #200, opened 2026-07-15
[Gpu] Backend-neutral shader compiler and guest-GPU renderer seam

### LINKED ISSUE #199: Proposal: pluggable guest-GPU backends (Vulkan / Metal / DX12) — shader-compiler and renderer seams
## Motivation

This is a proposal: I'd like us to move to per-platform GPU backends instead of one lowest-common-denominator path. The split I'm suggesting — the Vulkan backend stays with the contributors who maintain it today (Windows/Linux, and macOS via MoltenVK initially), I'm volunteering to own a native Metal backend for macOS, and a direct DX12 backend on Windows becomes possible later for whoever wants it. Today none of that is possible: Vulkan is hard-wired end-to-end — `VulkanVideoPresenter` (~7.5k lines) is windowing + swapchain + the entire guest renderer called statically from the export layers, and the AGC pipeline emits SPIR-V directly. This proposal describes the seams that would make backends pluggable; the same replace-the-LCD-with-our-own-contract move as the host platform abstraction (#181), applied to graphics.

## Proposed design: two seams

**1. Shader codegen splits at the IR handoff.** The pipeline is `decode → (Gen5ShaderState, Gen5ShaderEvaluation) → emitter → backend shader`. Everything left of the emitter is RDNA2/guest truth — instructions, operands, resource descriptors — and moves to a new backend-neutral `SharpEmu.ShaderCompiler` project. Each emitter is its own project consuming that contract: `SharpEmu.ShaderCompiler.Vulkan` (the existing SPIR-V translator, which notably needs no Vulkan bindings reference — emitters produce bytes, renderers own APIs), with `.Metal` (MSL) and `.Dx12` (DXIL) to follow. Shared semantics that backends must agree on (e.g. the inline-constant table) live in the core project so codegens cannot drift.

**2. The renderer hides behind `IGuestGpuBackend`.** The export layers (AGC/VideoOut/SystemService) talk to `GuestGpu.Current` — an interface expressed entirely in guest-domain terms. Two rules keep it honest:

- **No host-API value crosses the seam.** Guest addresses, guest format/number-type codes, raw CB_BLEND register bitfields, verbatim sampler descriptor dwords — each backend owns the guest→native translation. (An audit of the previous surface found this was *almost* already true despite the `Vulkan*` type names: the one genuine Vulkan value crossing, a `Silk.NET.Format` that callers never read, is now backend-internal.)
- **Submission is coarse-grained.** Draws, dispatches, blits, presents. Synchronization, render-pass structure, and descriptor/argument binding models are backend-internal — the places where Vulkan, Metal, and DX12 genuinely diverge never appear in the contract.

Shader compilation also lives behind the seam: backends own their codegen and return opaque `IGuestCompiledShader` handles that only the producing backend can submit. Nothing in the export layers knows which shader format exists.

## Evidence this isn't speculative

Before proposing this I built a working proof on the `shader-compiler-metal-spike` branch (galvesribeiro fork): an isolated Gen5→MSL compute codegen consuming the exact `(state, evaluation)` contract, with a three-tier test suite — golden MSL text, runtime compilation through Metal's OS compiler (fast-math off), and **execution on an Apple-silicon GPU asserting bit-exact results against reference math, including EXEC-mask semantics** (a masked `buffer_store` must not land; it doesn't). The pivot works on real hardware; the Metal emitter never needed a single Vulkan-shaped concept. The spike also validated the testing model each backend suite should follow: self-contained projects (no cross-test-project dependencies), synthetic hand-assembled fixtures only (game shader dumps are copyrighted and never land in the corpus), goldens as reviewable text diffs.

## Known follow-ups (deliberately not in the implementation PR)

- **Linked-program compilation.** The compile entry points still carry flat resource-slot bases (`globalBufferBase` etc., including a `+2` scalar-spill reservation) that encode the Vulkan emitter's packing policy. The right shape is a `TryCompileGraphicsProgram(vertexState/eval, pixelState/eval, outputs)` API where the backend owns slot assignment and the returned program carries its layout. That design needs the second backend's requirements (Metal argument tables) as input, so it's the first deliverable of the Metal backend work rather than a guess made with one consumer.
- **Presenter as an instance.** The Vulkan backend is currently a thin adapter over the existing presenter statics — correct for a mechanical extraction; folding the presenter into the backend type is cleanup that can happen anytime.
- **Control-flow reconstruction placement.** When the MSL emitter grows past straight-line code, structured-CF reconstruction should be lifted into the shared core rather than duplicated per emitter.

## Proposed roadmap, if this is accepted

1. Metal spike rebases onto the extracted projects (its temporary `InternalsVisibleTo` grants die).
2. macOS bring-up runs on MoltenVK (zero graphics changes needed — windowing is already portable) while the native Metal backend grows behind the seam: coverage first (control flow, loads, images, wave semantics behind a differential test matrix against the SPIR-V backend), presenter second.
3. DX12 backend follows the same recipe when someone wants it: implement `IGuestGpuBackend`, bring a DXIL emitter.

galvesribeiro:
Closes #199 — the proposal explaining the design, the neutrality rules, and the working Metal codegen proof it was validated against. This PR is the implementation, up for the same review as the proposal itself.

## What

Eight commits, each building and verified independently:

1. **`SharpEmu.ShaderCompiler`** — the backend-neutral half of shader compilation (Gen5 decoder, scalar evaluator, IR, metadata reader) extracted from `SharpEmu.Libs/Agc`. Three couplings surfaced and were fixed at the right depth: `GuestDrawKind` (a guest-domain, decoder-produced enum) was defined inside the Vulkan presenter and moves to the shared project; the evaluator's one HLE dependency (tracked-libc-heap read fallback) becomes an injectable hook installed by a Libs module initializer before any caller can reach it; the inline-constant table is promoted to a shared `Gen5InlineConstants` so codegens cannot drift.
2. **`SharpEmu.ShaderCompiler.Vulkan`** — the SPIR-V emitter moves whole, with no Vulkan bindings reference (emitters produce bytes; renderers own APIs). The ShaderDump tool drops all reflection for direct typed calls and no longer references `SharpEmu.Libs` at all — the conformance tool is emulator-independent now.
3. **`IGuestGpuBackend`** — the renderer seam. Export layers reach the renderer through `GuestGpu.Current` (mirroring `HostPlatform`); the Vulkan backend is a thin adapter over the existing presenter, so the extraction stays mechanical. The crossing types drop their `Vulkan` prefixes, which an audit showed were misnomers — every field is a neutral primitive or raw guest value. The one genuine Vulkan value in the old surface (a `Silk.NET.Format` callers never read) stops crossing: `TryGetRenderTargetOutputKind` surfaces only what callers consume.
4. **Shader compilation behind the seam** — backends own codegen and return opaque `IGuestCompiledShader` handles that only the producing backend can submit; the shader caches store handles; `AgcExports` no longer knows SPIR-V exists.
5–6. Review-finding fixes (rename collateral; diagnostics dumps labeled with the backend's payload extension instead of hardcoded `.spv`).
7. **Perf: the shader-cache hit path is now allocation-free and lock-free.** Every translated draw previously built its cache key with LINQ + `string.Join` + one interpolated string per render target — steady per-draw allocation even on cache hits — and looked it up under a lock shared with the tracing gate. The layout is now packed exactly (not hashed) into a `ulong`, the pixel-output array only materializes on cache misses, the caches are `ConcurrentDictionary`, and the seam-shaped render-target list is built once per translated draw instead of per submission. Same species of win as the reflection removal in the host abstraction PR.
8. LINQ removed from all code this branch introduced (explicit loops; adopting this as the going-forward rule for new code).

## Behavior

Everything except commit 7 is provably behavior-preserving — mechanical moves verified 1:1 against the originals by a multi-angle review (removed-behavior audit, cross-file caller trace, line-by-line): flag maps, cache keys, decode caching, and log strings unchanged; zero correctness findings. Commit 7 changes only allocation/locking on the hit path; cache keys remain exact (packed encoding, no hashing).

The one recorded design follow-up (see #199): the compile entry points still carry the Vulkan emitter's flat slot-base packing (`globalBufferBase` etc.). Deliberately deferred — the linked-program API that fixes it needs the Metal backend's requirements as design input, and folding a descriptor-layout rework into a mechanical lift would make this PR need deep game regression testing instead of being reviewable as a safe move.

## How verified

- Clean solution build, 0 warnings; locked-mode restore under the pinned SDK (lock diffs contain only project-reference additions, no package drift).
- Existing test suite passes; full ShaderDump conformance run passes (now driving the extracted projects typed).
- Adversarial multi-angle review with zero correctness findings; the two actionable cleanups it raised are fixed in commits 5–6.
- Needs a Windows game run for final confidence, as with any branch developed off-target: any known-rendering title exercising translated draws + compute, plus one run with `SHARPEMU_DUMP_SPIRV=1` to confirm shader dumps still appear (now honestly labeled). Draw-heavy titles should also show a lower allocation rate under `dotnet-counters` from commit 7.

### DIFF
--- a/.github/workflows/pr-build-links.yml
+++ b/.github/workflows/pr-build-links.yml
@@ -1,0 +1,105 @@
+# Copyright (C) 2026 SharpEmu Emulator Project
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Posts (and keeps updated) a single PR comment linking the per-platform build
+# artifacts once "Build and Release" finishes.
+#
+# This is a workflow_run workflow on purpose: PRs from forks run "Build and
+# Release" with a read-only GITHUB_TOKEN and cannot comment. workflow_run runs
+# afterwards in the base-repo context with a read-write token and does not check
+# out untrusted fork code, so it can comment safely. Because of that, GitHub only
+# triggers it from the copy on the default branch — it does nothing until merged
+# to main.
+name: PR Build Links
+
+on:
+  workflow_run:
+    workflows: ["Build and Release"]
+    types:
+      - completed
+
+permissions:
+  contents: read
+  actions: read
+  pull-requests: write
+
+jobs:
+  comment:
+    name: Post artifact links
+    runs-on: ubuntu-latest
+    # Only for successful PR builds — artifacts exist only when the build passed.
+    if: >-
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Post or update the artifact-links comment
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const run = context.payload.workflow_run;
+            const { owner, repo } = context.repo;
+
+            // Fork PRs leave workflow_run.pull_requests empty, so fall back to
+            // resolving the PR from the build's head commit.
+            let prNumber;
+            if (run.pull_requests && run.pull_requests.length > 0) {
+              prNumber = run.pull_requests[0].number;
+            } else {
+              const prs = await github.rest.repos.listPullRequestsAssociatedWithCommit({
+                owner, repo, commit_sha: run.head_sha,
+              });
+              const open = prs.data.find(pr => pr.state === 'open');
+              if (!open) {
+                core.info('No open PR for this build; nothing to comment.');
+                return;
+              }
+              prNumber = open.number;
+            }
+
+            // Collect the per-platform artifacts the build produced.
+            const artifacts = await github.paginate(
+              github.rest.actions.listWorkflowRunArtifacts,
+              { owner, repo, run_id: run.id, per_page: 100 },
+            );
+            const platforms = [
+              { key: 'win-x64', label: 'Windows (win-x64)' },
+              { key: 'linux-x64', label: 'Linux (linux-x64)' },
+              { key: 'osx-x64', label: 'macOS (osx-x64)' },
+            ];
+            const rows = [];
+            for (const platform of platforms) {
+              const artifact = artifacts.find(a => a.name.includes(platform.key));
+              if (!artifact) {
+                continue;
+              }
+              const url = `https://github.com/${owner}/${repo}/actions/runs/${run.id}/artifacts/${artifact.id}`;
+              rows.push(`| ${platform.label} | [\`${artifact.name}\`](${url}) |`);
+            }
+            if (rows.length === 0) {
+              core.info('No platform artifacts on this run; nothing to comment.');
+              return;
+            }
+
+            const marker = '<!-- pr-build-links -->';
+            const body = [
+              marker,
+              `### 📦 Build artifacts — \`${run.head_sha.substring(0, 7)}\``,
+              '',
+              '| Platform | Download |',
+              '| --- | --- |',
+              ...rows,
+              '',
+              `From [build run #${run.run_number}](${run.html_url}). ` +
+                'Downloads require a GitHub login and expire after 90 days.',
+            ].join('\n');
+
+            // Upsert one comment so repeated builds refresh it in place.
+            const comments = await github.paginate(github.rest.issues.listComments, {
+              owner, repo, issue_number: prNumber, per_page: 100,
+            });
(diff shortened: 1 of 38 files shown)


Spooks4576:
Looks good will merge now

PR #201, opened 2026-07-15
[Vulkan] Put a limit on the temporary buffer pool

Dafenxz0:
SharpEmu already recycles its temporary host-visible buffers once a guest submission is finished. The useful part was there, but the pool lived inside `VulkanVideoPresenter` and kept every allocation until shutdown. A game that rotates through enough buffer sizes and usages could therefore leave a lot of idle Vulkan memory behind.

This moves that bookkeeping into a small `VulkanHostBufferPool` and gives idle allocations a 128 MiB budget. Returned buffers are still reused by usage and power-of-two capacity. If keeping one would cross the budget, it is released instead.

The existing fence lifetime does not change: buffers only return to the pool after their submission completes. Upload behavior also does not change; the current CPU snapshot is copied on every use. This PR does not try to preserve guest buffer contents or infer CPU/GPU ownership.

In practice this should make long sessions and games with varied transient workloads less likely to accumulate host-visible memory, while taking one more resource-lifetime concern out of the presenter.

Validation:

- `dotnet test tests/SharpEmu.Libs.Tests/SharpEmu.Libs.Tests.csproj -c Release` (29 passed)
- `dotnet build SharpEmu.slnx -c Release --no-restore` (0 warnings, 0 errors)
- ShaderDump synthetic suite (all programs behaved as expected)

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VulkanHostBufferPool.cs
+++ b/src/SharpEmu.Libs/VideoOut/VulkanHostBufferPool.cs
@@ -1,0 +1,109 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+using Silk.NET.Vulkan;
+using VkBuffer = Silk.NET.Vulkan.Buffer;
+
+namespace SharpEmu.Libs.VideoOut;
+
+internal readonly record struct VulkanHostBufferPoolKey(
+    BufferUsageFlags Usage,
+    ulong Capacity);
+
+internal readonly record struct VulkanHostBufferAllocation(
+    VkBuffer Buffer,
+    DeviceMemory Memory,
+    VulkanHostBufferPoolKey Key,
+    nint Mapped);
+
+internal sealed class VulkanHostBufferPool : IDisposable
+{
+    private readonly Dictionary<VulkanHostBufferPoolKey, Stack<VulkanHostBufferAllocation>>
+        _available = [];
+    private readonly Dictionary<ulong, VulkanHostBufferAllocation> _allocations = [];
+    private readonly HashSet<ulong> _cachedHandles = [];
+    private readonly Action<VulkanHostBufferAllocation> _destroy;
+
+    public VulkanHostBufferPool(
+        ulong maximumCachedBytes,
+        Action<VulkanHostBufferAllocation> destroy)
+    {
+        MaximumCachedBytes = maximumCachedBytes;
+        _destroy = destroy;
+    }
+
+    public ulong MaximumCachedBytes { get; }
+
+    public ulong CachedBytes { get; private set; }
+
+    public bool TryRent(
+        VulkanHostBufferPoolKey key,
+        out VulkanHostBufferAllocation allocation)
+    {
+        if (!_available.TryGetValue(key, out var available) ||
+            !available.TryPop(out allocation))
+        {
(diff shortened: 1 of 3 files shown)


Dafenxz0:
Would anyone be able to test this branch on a few actual titles before I mark it ready for review?

This change only reuses and limits idle host-visible Vulkan allocations. It still uploads the current CPU data on every use and does not preserve guest buffer contents. Even so, I would like to verify it against `main` rather than relying only on CI and synthetic tests.

The most useful results would include the title and scene tested, GPU, FPS or frame-time comparison, memory usage, and any visual regressions, crashes, or hangs. Testing a title that regressed with #158 would be especially helpful.

kostyaff (approved):
## Review

Extracts the host buffer pool into a dedicated class `VulkanHostBufferPool` with a memory budget cap (128 MB max cached).

### `VulkanHostBufferPool` (new, 108 lines) — ✅

- `TryRent(key, out allocation)` — pops from per-key stack. Updates `_cachedHandles` and `CachedBytes`. ✅
- `Register(allocation)` — adds to `_allocations` by buffer handle. ✅
- `Return(buffer, memory)` — looks up by handle, verifies memory match, checks budget: if `allocation.Key.Capacity > MaximumCachedBytes - CachedBytes`, destroys instead of caching. ✅
- `Dispose()` — destroys all tracked allocations. ✅
- Double-return guard via `_cachedHandles` HashSet — prevents the same buffer from being cached twice. ✅

### Budget enforcement — ✅

`MaximumCachedHostBufferBytes = 128 MB`. When cache is full, returned buffers are destroyed instead of cached. This prevents unbounded memory growth from cached host buffers — a real problem in long-running emulation sessions.

### Integration — ✅

Old inline `Dictionary<HostBufferPoolKey, Stack<HostBufferAllocation>>` + `_hostBufferAllocations` replaced with `VulkanHostBufferPool`. `CreateHostBuffer` uses `TryRent`/`Register`, `RecycleHostBuffer` uses `Return`. `DisposeVulkan` calls `_hostBufferPool.Dispose()`. Clean migration.

### `DestroyHostBufferAllocation` callback — ✅

Passed as constructor delegate — the pool doesn't know about Vulkan devices, just calls the provided destroy function. Good separation.

### Tests (60 lines) — ✅

- `ReturnedAllocationCanBeRentedAgain` — rent → return → rent returns same allocation
- `ReturnDestroysAllocationThatWouldExceedBudget` — over-budget allocation is destroyed, not cached
- `UnknownAllocationIsNotClaimedByPool` — unregistered buffer is rejected

Good coverage of the three key paths.

## Verdict: ✅ Approve

Clean extraction with memory budget enforcement. Prevents unbounded GPU memory growth. Well-tested.

par274:
rebase and comment please.

Dafenxz0:
Rebased onto the current `main` and resolved the `VulkanVideoPresenter` conflict.

The resolution keeps the newer persistent host mappings and guest-buffer coherency/readback path from `main`; the bounded pool now owns those mapped allocations and unmaps them only when they are actually destroyed.

Validation:
- project builds successfully with .NET SDK 10.0.103
- `VulkanHostBufferPoolTests`: 3/3 passed
- full library suite: 100/101 passed locally; the remaining pre-existing `Sprintf_ReadsVariadicDoubleFromXmmRegister` failure is locale-specific (`0,5576` on es-ES vs `0.5576` expected)

PR #207, opened 2026-07-15
HLE/AGC: stopgap wake all graphics events on unmatched EVENT_WRITE

### LINKED ISSUE #173: AGC event queue never wakes waiters (EVENT_TYPE vs eventId mismatch)
### Emulator Version / Commit

d8397b0

### Bug Description

i've been trying to get poppy playtime chapter 1 [ppsa20591] running (past the point in #3, which just silently closed near `__cxa_guard_acquire`). on current main it boots a lot further now — full engine thread bring-up, rhithread and the agc threads all get scheduled — but then it just hangs. left it running for over 9 minutes and nothing changed. (tested on commit d8397b0, .net 10 release build, windows, native cpu engine.)

traced it to the agc event queue code. the game submits a real command buffer with shaders and `EVENT_WRITE` completion packets, and it registers a couple of event listeners first:

```
agc.driver_add_eq_event eq=0x2 id=0x20 udata=0x0
agc.driver_add_eq_event eq=0x2 id=0x00 udata=0x0
```

but when the command buffer parser fires those events, nothing ever matches:

```
agc.dcb.event type=0x07 queues=0
agc.dcb.event type=0x10 queues=0
agc.dcb.event type=0x2E queues=0
agc.dcb.event type=0x2C queues=0
```

every one comes back `queues=0`. `AgcInterruptThread` sits in `sceKernelWaitEqueue` the whole time — its import count doesn't move between the first stall log and one taken 9 minutes later.

looking at the code, the two sides just aren't using the same id space. `sceAgcDriverAddEqEvent` (`AgcExports.cs` ~line 1946) registers the listener using the `eventId` the game passed in directly (`0x20`/`0x00` here). the `IT_EVENT_WRITE` handler (`AgcExports.cs` ~line 2626) instead triggers using the raw pm4 `EVENT_TYPE` field off the packet (`0x07`/`0x10`/`0x2E`/`0x2C`) as the ident. those aren't the same numbering scheme, so they never match, and `TriggerRegisteredEvents` (`KernelEventQueueCompatExports.cs:465`) never has anything to wake up.

doesn't seem specific to this game — anything that waits on a gfx/compute completion event through `sceAgcDriverAddEqEvent` before continuing would hit the same hang.

i don't know the real ps4/ps5 mapping from `EVENT_TYPE` to that `eventId` though, so didn't want to just guess and hack something in. figured it's worth raising here first. couple options as i see it:

- someone knows the real mapping and we wire that up properly
- as a stopgap, treat any completion-style `EVENT_WRITE` as satisfying any registered graphics event on that queue — probably unblocks most games even if it's not 100% accurate

happy to test anything against this build/game if it helps.

LOG - [sharpemu-poppy-agc-trace.log](https://github.com/user-attachments/files/30023468/sharpemu-poppy-agc-trace.log)

### Log File

```shell

```

Rick21-bit:
Closes #173.

The issue shows that `sceAgcDriverAddEqEvent` registers listeners with an `eventId` (e.g. `0x20`, `0x00`) while PM4 `EVENT_WRITE` packets fire with `EVENT_TYPE` values (`0x07`, `0x10`, `0x2E`, `0x2C`). The two identifier spaces never match, so `sceKernelWaitEqueue` waits forever and games hang.

This change adds a fallback: when an exact-match trigger returns zero, we wake every `KernelEventFilterGraphics` registration using the raw `EVENT_TYPE` as event data. This is the stopgap option suggested in #173 and should unblock many games that wait on command-buffer completion events.

I cannot run Windows + Vulkan builds locally, so this needs testing against Poppy Playtime Chapter 1 or any title discussed in #173.

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -2723,6 +2723,18 @@ public static class AgcExports
                     eventType,
                     KernelEventQueueCompatExports.KernelEventFilterGraphics,
                     eventType);
+
+                // Stopgap: real PS5 command buffers use EVENT_TYPE values that don't
+                // line up with the eventId games register via sceAgcDriverAddEqEvent.
+                // If the exact match doesn't wake anything, fire all registered graphics
+                // events so completion waiters can proceed.
+                if (triggered == 0)
+                {
+                    triggered = KernelEventQueueCompatExports.TriggerAllRegisteredEvents(
+                        KernelEventQueueCompatExports.KernelEventFilterGraphics,
+                        eventType);
+                }
+
                 if (tracePackets)
                 {
                     TraceAgc($"agc.dcb.event type=0x{eventType:X2} queues={triggered}");
--- a/src/SharpEmu.Libs/Kernel/KernelEventQueueCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelEventQueueCompatExports.cs
@@ -589,6 +589,58 @@ public static class KernelEventQueueCompatExports
         return triggeredCount;
     }
 
+    /// <summary>
+    /// Wake every registration with the given filter, regardless of identifier.
+    /// This is a stopgap for command-buffer EVENT_WRITE packets where the PM4
+    /// EVENT_TYPE does not match the application-registered eventId.
+    /// </summary>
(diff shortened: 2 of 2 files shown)


par274:
I understand why this helps as a stopgap, but waking every registered graphics event is too broad to enable unconditionally.

PR #208, opened 2026-07-15
Void Terrarium (PPSA03061): fully playable — Gen5/RDNA2 shader coverage, ordered GPU queues, HLE and input work

xnetcat:
Based on current upstream `main` (includes the merged #47); the branch is a **small commit series on top of it** (port, APR streaming restore, gfx10 decode + validation-gate alignment, Windows TLS prefix). This brings `void tRrLM();++ //Void Terrarium++` [PPSA03061] from "boots to a frozen splash" to **fully playable** on the POSIX port: boot → title menu → New Game → intro → gameplay, verified end-to-end unattended (`SHARPEMU_AUTO_CROSS=repeat:4`) with clean rendering and no crash over multiple 15-minute soaks.

### What's in here

**GPU / AGC.** The Gen5 (RDNA2) shader front end learns the instructions this game's shaders actually use: `ds_write/ds_read_b96/b128` with graphics-stage LDS modeled as per-invocation private memory (the missing piece that silently dropped the scene geometry draws — the "black screen" was never a game stall), DPP, wave masks, gradient sampling, typed buffer resources, and stricter descriptor validation. Draws that still fail translation degrade (1×1 fallback texture, unresolved-scalar degradation) instead of being dropped. Texture format 50 (sampled A2R10G10B10) aliases the on-GPU render target. CPU-side detiling (GnmTiling) runs by default for the linearly-ordered base swizzle modes 1–12 with a precomputed per-block offset table; `_T`/`_X` variants stay opt-in behind `SHARPEMU_DETILE=1`.

**Command processor.** DCB submissions are queued per logical queue and pumped in order: an unsatisfied `WAIT_REG_MEM` suspends the queue and resumes when the label is written, instead of parsing packets out of order (this plus the `sceAgcDmaDataPatchSetSrcAddressOrOffsetOrImmediate` implementation fixes the one-frame-then-freeze render loop). PM4 completion and end-of-pipe writeback are modeled precisely enough for the game's fences.

**Presenter.** Offscreen draw batching into shared command-buffer submissions, texture/descriptor/pipeline caching with SIGSEGV-based guest-image write tracking for invalidation, flip pacing to the display cadence, MRT support for all color targets, an F1 perf overlay, and MoltenVK performance defaults (async queue submits, concurrent pipeline compilation, Metal argument buffers). Window close and Ctrl-C request cooperative guest shutdown and force a process exit if a wedged guest ignores it.

**Kernel / HLE.** APR file streaming (`sceKernelAprResolveFilepathsToIds`, `sceKernelAprGetFileStat` — without these, asset streaming returns NOT_FOUND and the game never leaves its splash), SCE fiber context switching (the game schedules on fibers), savedata transaction resources, pthread TLS destructor passes on thread exit, AudioOut with a CoreAudio backend and proper 5.1→stereo downmix, Ngs2, AvPlayer, message/error dialogs, and the trophy/CES/telemetry startup stubs — zero unresolved imports for this title.

**Scheduler.** A background Ready-thread dispatcher so a hot-spinning guest thread (this game busy-polls `sceAudioOutOutput`) cannot starve runnable threads; the old behavior only pumped the ready queue on blocking HLE calls and deadlocked the loader.

**Input.** Building on the host input seam from #192: `SHARPEMU_AUTO_CROSS=repeat[:seconds]` keeps pressing Cross on an interval (release guaranteed between presses) so a headless run drives itself through the menus into gameplay; a timed-offset list (`SHARPEMU_AUTO_CROSS="40,52,64"`) is also supported, with `SHARPEMU_AUTO_CROSS_HOLD_MS` and `SHARPEMU_LOG_AUTO_CROSS=1` to tune/observe it. `SHARPEMU_AUTO_WALK=<start-seconds>` rotates a held d-pad direction for unattended tutorial/dungeon traversal (start it past the title menu). Pad and AudioOut2 accept the user id the user service hands out — a mismatch there used to strand the game with an error code as its pad handle and no working input.

### Verified / performance

PPSA03061 v01.000.000 on macOS 27 (Apple M4, Rosetta 2, MoltenVK 1.4.1):

| state | before | this branch |
|---|---|---|
| boot | frozen splash, 0 draws | full boot → gameplay, unattended |
| title screen | — | **60.0 fps locked (16.7 ms)** |
| story / gameplay scenes | — | 13–20 fps |
| window close / Ctrl-C on a wedged guest | headless zombie process | forced exit after 2 s |

In-game frame times are bound by the guest's own x86-64 code under Rosetta 2: a `dotnet-trace` cpu-sampling capture at 13–18 fps gameplay attributes ~100 % of self-time to unmanaged guest execution (managed emulator code ≈ 0 %), and the intro cinematic renders the same ~110 draws/frame at 55–60 fps — heavier scenes simply run 3–4× more guest CPU per frame. Total process CPU in-game is ~1–1.6 cores. `dotnet test` passes (26/26). One Windows-facing change: the static-TLS prefix below the thread pointer grows from the historical 4 KB to the 64 KB POSIX hosts already map — titles whose PT_TLS exceeds a page (this one needs 0x1B00 bytes) previously failed to load on Windows.

### Known issues

- Occasional boot-time failure under Rosetta (unrecovered guest SIGSEGV during loading); a relaunch gets through. Pre-existing flakiness, not introduced here — tracked separately.
- `_T`/`_X` swizzle detile variants remain opt-in (block placement only approximated).

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.PosixSignals.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.PosixSignals.cs
@@ -310,6 +310,36 @@ public sealed unsafe partial class DirectExecutionBackend
 		}
 		if (disposition != -1 && !_posixSignalWarmup)
 		{
+			// Unrecovered: the process is about to die on the re-raised signal.
+			// Dump registers and the code window around RIP the way the Windows
+			// reporter does — the byte window is what identifies which emitted
+			// stub (and which instruction) an intermittent native fault is in.
+			try
+			{
+				var faultRip = ReadCtxU64(contextRecord, CTX_RIP);
+				Console.Error.WriteLine(
+					$"[LOADER][ERROR] posix-signal unrecovered: sig={signal} rip=0x{faultRip:X16} " +
+					$"fault=0x{record.ExceptionInformation[1]:X16} access={record.ExceptionInformation[0]} " +
+					$"rsp=0x{ReadCtxU64(contextRecord, CTX_RSP):X16} rax=0x{ReadCtxU64(contextRecord, CTX_RAX):X16} " +
+					$"rcx=0x{ReadCtxU64(contextRecord, CTX_RCX):X16} rdx=0x{ReadCtxU64(contextRecord, CTX_RDX):X16} " +
+					$"r11=0x{ReadCtxU64(contextRecord, CTX_R11):X16}");
+				var windowStart = faultRip >= 0x20 ? faultRip - 0x20 : 0;
+				var window = new byte[0x40];
+				if (TryReadHostBytes(windowStart, window))
+				{
+					Console.Error.WriteLine(
+						$"[LOADER][ERROR]   code [rip-0x20..]: {Convert.ToHexString(window)}");
+				}
+				else if (TryReadHostBytes(faultRip, window))
+				{
+					Console.Error.WriteLine(
+						$"[LOADER][ERROR]   code [rip..]: {Convert.ToHexString(window)}");
+				}
+				Console.Error.Flush();
+			}
+			catch
+			{
+			}
 			return false;
 		}
 
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -1759,6 +1759,17 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 		}
 	}
 
+	// ldmxcsr raises #GP for any set bit outside the low 16 (the observed
+	// intermittent "boot crash" was an emitted stub executing ldmxcsr with a
+	// polluted continuation value, e.g. 0x00102510). All 16 low bits are
+	// architecturally loadable on x86-64, so masking is always safe; an
+	// all-clear value falls back to the SDK default control state.
+	private static uint SanitizeGuestMxcsr(uint value)
+	{
+		value &= 0xFFFFu;
+		return value == 0 ? 0x1F80u : value;
+	}
+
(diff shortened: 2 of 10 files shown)


par274:
This is a call shader decoder part 2

xnetcat:
## Intermittent boot-crash root cause and fixes

A Windows crash report with code bytes finally identified the intermittent boot-time native crash (both platforms): `ldmxcsr` faulting inside an emitted stub on a polluted MXCSR immediate (`0x00102510` — reserved bit set ⇒ #GP).

The pollution source was a **mismatched trampoline/gateway pair** from the rebase: the import gateway reads guest RAX/R10/R11/MXCSR/x87-CW from the 0xB0-byte save area below the argument pack, but the branch carried the older trampoline that only reserves the 0x80-byte XMM area — so every non-leaf HLE dispatch loaded unsaved stack garbage into the guest context, and captured continuations (fiber switches, blocked-thread migration) baked it into stubs.

Landed:
- `fceebc3` restores the matched trampoline (also preserves the libSceFiber setjmp/longjmp register contract across HLE calls).
- MXCSR is sanitized (low 16 bits, default `0x1F80`) at every continuation load site as defense in depth.
- `2f5b9c7` guest-context transfers now probe the target RIP: a corrupt continuation logs and downgrades to a guest-thread exit instead of executing arbitrary host memory.
- `788d6f8` Windows: 64 KB static-TLS prefix (this title's PT_TLS needs 0x1B00 > the old 4 KB page).
- `51081a5` the POSIX signal bridge now dumps registers + the RIP code window on unrecovered faults, so any residual intermittent fault self-identifies.

Validated: full unattended boot→gameplay run on the fixed build (3488 flips, 17–18 fps in-game, zero faults). One boot in three cycles still hit a residual macOS-only fault at a stable stub offset — now self-diagnosing via the new dump; tracked in Known Issues.

Spooks4576:

Live status: https://pr.sharpemu.app/#pr-208

PR #211, opened 2026-07-15
Fix sceNetHtonl/Htons/Ntohl/Ntohs discarding their converted result

6can:
`sceNetHtonl`, `sceNetHtons`, `sceNetNtohl` and `sceNetNtohs` look right at a glance — each computes the byte swap into `Rax` — but they finish with `return ctx.SetReturn(0);`. `SetReturn` writes its argument into `Rax` and returns it, so that call overwrites the freshly computed value with `0`. The net effect is that all four conversions returned `0` to the guest regardless of input, which silently corrupts anything that byte-swaps ports/addresses (e.g. filling in a `sockaddr_in`).

The value-returning exports in this same file already show the intended shape: `sceNetPoolCreate` sets `Rax` and then `return (int)OrbisGen2Result.ORBIS_GEN2_OK;`. I changed the four helpers to do the same — leave the swapped value in `Rax` and return `ORBIS_GEN2_OK` as the dispatch status rather than routing through `SetReturn`.

Added a `NetExports` test suite covering the 32-bit and 16-bit swaps, that `htons`/`ntohs` only touch the low two bytes, an `htonl`→`ntohl` round-trip, and a regression guard that a non-zero input never converts to `0`.

Verified locally on .NET 10: all 70 tests in the project pass, and reverting just this fix fails 9 of the 10 new tests (the one that still passes swaps `0`, which is `0` either way).

### DIFF
--- a/src/SharpEmu.Libs/Network/NetExports.cs
+++ b/src/SharpEmu.Libs/Network/NetExports.cs
@@ -103,8 +103,10 @@ public static class NetExports
     public static int NetHtonl(CpuContext ctx)
     {
         var value = unchecked((uint)ctx[CpuRegister.Rdi]);
+        // The byte-swapped result is the return value and already lives in Rax; return OK as the
+        // dispatch status without going through SetReturn, which would overwrite Rax with 0.
         ctx[CpuRegister.Rax] = BinaryPrimitives.ReverseEndianness(value);
-        return ctx.SetReturn(0);
+        return (int)OrbisGen2Result.ORBIS_GEN2_OK;
     }
 
     [SysAbiExport(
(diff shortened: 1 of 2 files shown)


kostyaff (approved):
## Review

Fixes a real bug: `NetHtonl`, `NetHtons`, `NetNtohl`, `NetNtohs` computed the byte-swapped value into Rax, then called `ctx.SetReturn(0)` which overwrote Rax with 0. Every call returned 0 regardless of input.

### Fix — ✅

Before: `ctx.SetReturn(0)` overwrites Rax with 0, destroying the result.
After: `return (int)OrbisGen2Result.ORBIS_GEN2_OK` — returns OK status without touching Rax. Correct.

The comment explains the issue clearly: `SetReturn` writes 0 to Rax, which destroys the byte-swapped result.

### Tests — ✅

6 tests: all four functions, high-bit masking (16-bit functions ignore bits above 0xFFFF), 64-bit input masking for 32-bit functions, round-trip identity, and a regression guard that explicitly checks non-zero return for non-zero input.

The regression guard test is especially good — it would have caught the original bug.

## Verdict: ✅ Approve

Clean bugfix with good test coverage.

PR #215, opened 2026-07-15
[Vulkan] Reuse render passes for repeated MRT layouts

Dafenxz0:
The MRT path creates a framebuffer and a compatible `VkRenderPass` for every translated draw, then retires both after the submission fence. Framebuffers are tied to the current image views, but render passes can be reused when their attachment contract is identical.

This keeps one MRT render pass per compatible layout while continuing to create framebuffers per draw. Since the original version of this PR, `main` gained per-attachment load/clear state and optional depth attachments. The cache key now includes:

- ordered color formats and attachment count;
- the color Load/Clear mask;
- whether depth is present;
- whether depth uses Load or Clear.

That prevents a cached pass from changing attachment contents after the rebase. Cached render passes live until Vulkan shutdown and are destroyed after graphics pipelines. Single-target render passes remain owned by their guest images.

Nothing about guest memory, barriers, command order, framebuffer lifetime, or fences changes. The goal remains to remove repeated `vkCreateRenderPass`/`vkDestroyRenderPass` calls from G-buffer-style workloads without introducing another coherency model.

I deliberately did not cache framebuffers because safe invalidation depends on image-view lifetime.

Validation:

- Full Release build: 0 warnings, 0 errors
- `SharpEmu.Libs.Tests`: 235 passed
- `SharpEmu.SourceGenerators.Tests`: 33 passed
- Added coverage for format order/count, color Load/Clear state, depth presence, and depth Load/Clear state
- Previous gameplay run on Dreaming Sarah and Dead Cells passed; the updated commit needs the automated runner to repeat that check

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
+++ b/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
@@ -1706,6 +1706,43 @@ internal static unsafe class VulkanVideoPresenter
     private static bool IsKnownGuestTextureFormat(uint format) =>
         format is >= 1 and <= 19 or 34 or >= 169 and <= 182;
 
+    internal static string GetMrtRenderPassLayoutKey(
+        IReadOnlyList<Format> formats,
+        IReadOnlyList<bool> initialized,
+        bool hasDepth,
+        bool depthInitialized)
+    {
+        if (formats.Count is < 1 or > 8 || formats.Count != initialized.Count)
+        {
+            throw new ArgumentException(
+                "MRT formats and initialization states must have matching counts");
+        }
+
+        var initializedMask = 0u;
+        var key = new StringBuilder(formats.Count * 12 + 8);
+        for (var index = 0; index < formats.Count; index++)
+        {
+            if (index != 0)
+            {
+                key.Append(',');
+            }
+
+            key.Append((uint)formats[index]);
+            if (initialized[index])
+            {
+                initializedMask |= 1u << index;
+            }
+        }
+
+        key.Append('|').Append(initializedMask).Append('|');
+        if (hasDepth)
+        {
+            key.Append(depthInitialized ? "DL" : "DC");
+        }
+
+        return key.ToString();
+    }
+
     private static byte[] CreateBlackFrame(uint width, uint height)
     {
         if (width == 0 || height == 0 || width > 8192 || height > 8192)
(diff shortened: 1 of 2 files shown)


kostyaff (approved):
## Review

Caches MRT (multi-render-target) render passes by format layout key, instead of creating and destroying a new `RenderPass` per draw call.

### `GetRenderTargetLayoutKey` — ✅

`string.Join(,, formats.Select(format => (uint)format))` — deterministic key from format enum values. Same logic was already inline in `CreateTranslatedPipeline`, now extracted and reused. `ArgumentOutOfRangeException` for count < 1 or > 8 — correct guard.

### `GetOrCreateMrtRenderPass` — ✅

Dictionary lookup by layout key → cache hit returns existing `RenderPass`. Cache miss creates new, adds to `_mrtRenderPasses`, sets debug name. Correct.

### Refactor: `CreateRenderPassAndFramebuffer` split — ✅

Old method did both. Now split into:
- `CreateRenderPass(IReadOnlyList<Format>)` — creates RenderPass only
- `CreateFramebuffer(RenderPass, IReadOnlyList<ImageView>, width, height)` — creates Framebuffer only
- `GetOrCreateMrtRenderPass` — caches RenderPass, creates Framebuffer separately

This is the correct separation: RenderPass is cacheable (format-dependent only), Framebuffer is per-draw (view-dependent).

### `TransientRenderPass` removal — ✅

`RenderPass` is now cached, not transient. `TransientRenderPass` field removed from `TranslatedDrawResources`. Cleanup in `DestroyTranslatedDrawResources` and `ExecuteOffscreenDraw` no longer destroys the RenderPass. Correct — the RenderPass is owned by `_mrtRenderPasses` and destroyed in `DisposeVulkan`.

### Disposal — ✅

`_mrtRenderPasses` cleared in `DisposeVulkan` — all cached render passes destroyed. No leak.

### Error path — ✅

`GetOrCreateGuestImage` (single-target path): `CreateRenderPass` then `CreateFramebuffer`, with try/catch that destroys the RenderPass if `CreateFramebuffer` throws. Correct cleanup.

### Test — ✅

`LayoutKeyIncludesAttachmentOrderAndCount` — verifies key equality for same formats, inequality for different order, inequality for different count. Clean and sufficient.

## Verdict: ✅ Approve

Significant perf improvement — avoids vkCreateRenderPass/vkDestroyRenderPass per MRT draw. Clean separation of concerns. Good test.

Spooks4576:

Tested `agent/reuse-mrt-render-passes` @ `94bc94e3` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-215

PR #216, opened 2026-07-15
[AGC/Vulkan] Extend PS5 runtime and rendering compatibility

Acelogic:
## Why

This draft brings the generic PS5 compatibility work developed while booting ASTRO BOT onto current `main`. It fills gaps in native execution, HLE services, Gen5 shader translation, resource lifetime, depth/MRT handling, input, compute storage, and Vulkan presentation without relying on replacement title shaders.

This remains a draft: ASTRO now advances through its real scenes, but the final title/world-map composition is still incorrect.

## Upstream integration status

The branch is merged through upstream `ad5a7d3` (`[CI] fix asset`, #245), including the backend-neutral shader/GPU split from #200 and the later fixes through #232, #237, #238, #241, and #245. It is currently **0 commits behind** `main`.

- migrated the Gen5 and SPIR-V changes into `SharpEmu.ShaderCompiler` and `SharpEmu.ShaderCompiler.Vulkan`
- reconciled the renderer path with the backend-neutral guest-GPU seam
- retained upstream's corrected `V_READLANE_B32` / `V_WRITELANE_B32` decode while preserving the non-overlapping ASTRO opcode coverage
- retained POSIX condition-variable semantics from #223 while preserving the cooperative guest waiter/mutex-reacquisition path
- incorporated upstream's graphics-event filter routing while keeping ordered guest-GPU side effects
- restored runtime scalar-buffer descriptors on every translated graphics submission path
- restored eight generic AGC helper exports omitted during the #200 migration
- fixed promoted MRT targets whose primary Vulkan view existed before their mip-view array was populated
- preserved valid address-zero compute storage descriptors as scratch bindings instead of dropping the dispatch
- audited all compatibility exports: **875 attributes, 875 unique NIDs, 0 duplicate/shadowed NIDs**

## Commit structure

The PR is **14 commits ahead of `main`**: nine focused non-merge commits plus five explicit integration merges.

### Focused commits

1. `e58ff96` `[Core] Add POSIX native execution and PS5 SELF support`
2. `27484b9` `[HLE] Expand PS5 service and media compatibility`
3. `4b0dafc` `[AGC/Vulkan] Extend Gen5 shader and presentation support`
4. `8e4afa1` `[Core] Align static TLS reservation across hosts`
5. `6356056` `[Pad] Align primary user ID with UserService`
6. `aacc5dd` `[Gpu] Preserve runtime scalar buffers across renderer seam`
7. `e2ec3e9` `[AGC] Restore omitted command helper exports`
[...]

PR #222, opened 2026-07-15
[AGC] Implement RDNA2 buffer/image/DS 32-bit atomic instructions

missatjuhvdk1:
This picks up the "missing atomics" item from the ISA gaps list in #36.

The translator only handled `BUFFER_ATOMIC_ADD` before. This fills in the rest of the 32-bit atomics across all three memory paths:

- **MUBUF**: swap, cmpswap, sub, smin/umin, smax/umax, and, or, xor, inc, dec (add already existed)
- **MIMG**: swap, add, sub, smin/umin, smax/umax, and, or, xor, inc (cmpswap and dec were already in the decode table, now they translate too)
- **DS**: add, sub, inc, dec, signed/unsigned min and max, and, or, xor, cmpst, plus the RTN variants and wrxchg

On the SPIR-V side I generalized the existing `BufferAtomicAdd` path instead of duplicating it per opcode. Image atomics go through `OpImageTexelPointer` on the storage image variable (float formats are rejected, sint formats bitcast in and out). LDS atomics use Workgroup scope with workgroup memory semantics, buffer and image atomics use Device scope.

Two things worth a close look in review:

1. `DS_CMPST` has its operands the other way around from buffer/image cmpswap: DATA0 is the comparator and DATA1 is the new value. I checked this against gem5's GPU model and shadPS4 since the ISA doc is easy to misread here. There's a dedicated test that locks the order in so it can't silently regress.
2. AMD's `ATOMIC_INC`/`ATOMIC_DEC` wrap against a clamp value. I mapped them to `OpAtomicIIncrement`/`IDecrement`, which is exact for the usual `0xFFFFFFFF` clamp but not for arbitrary ones. That seemed like a better tradeoff than emitting a cmpxchg loop for a case games basically don't use. Happy to change it if you'd rather have exact semantics.

Out of scope: the 64-bit X2 variants, FLAT/GLOBAL atomics, and the DS float/RSUB/MSKOR/WRAP ops. Kept those out to keep the PR small, can follow up if there's interest.

**Testing**

9 new synthetic tests under `tests/SharpEmu.Libs.Tests/Agc`. Six cover decoding (control fields, operand wiring, dword counts, the cmpst operand order) and three run a small program end to end through decode, scalar evaluation and SPIR-V compilation, then check the expected `OpAtomic*` opcodes actually show up in the module. Full suite passes 35/35. Instruction encodings were taken from the public RDNA2 ISA reference, no proprietary material involved.

Developed with AI assistance per the CONTRIBUTING.md guidelines; I've gone through the implementation and can speak to the decisions in review.


kostyaff (approved):
Reviewed on head `799119d` against current `main`.

Clean PR. Fills the atomic opcode gap from the #36 ISA list across all three memory paths. I checked the MUBUF and DS opcode numbers against the AMD RDNA2 ISA XML — BUFFER_ATOMIC_SWAP=48 (0x30), DS_ADD_U32=0, DS_CMPST_RTN_B32=48 (0x30) all match.

What I verified:

**Decode tables**: MUBUF 0x30-0x3D, MIMG 0x0F-0x1C, DS 0x00-0x0B + 0x10 + 0x20-0x30 — all match the ISA. The `IsDataShareAtomic` classification and the `IsStorageImageOperation` extension are correct.

**DS_CMPST operand order**: DATA0=comparator, DATA1=new value — reversed from buffer/image cmpswap. The decode test `DsCmpstRtnB32_OrdersComparatorBeforeNewValue` locks this in. Good call checking against gem5 and shadPS4.

**SPIR-V emission**: `EmitAtomic` helper correctly handles three cases: Inc/Dec (no value operand), CompareExchange (value + comparator + unequal semantics with Release stripped), and the rest (single value). The `TryGetAtomicOp` suffix mapping is clean and avoids duplication.

**Memory semantics**: DS atomics use Workgroup scope (2) with workgroup memory semantics (0x108). Buffer atomics use Device scope (1) with 0x48. Image atomics use Device scope (1) with 0x808. These are reasonable — Acquire+Release for the common case, CompareExchange downgrades unequal to Acquire-only per SPIR-V spec.

**Image atomics**: `OpImageTexelPointer` + `OpAtomic*` on storage image. Float formats correctly rejected. Sint formats bitcast in/out. Coordinates clamped when bounds are known.

**Inc/Dec approximation**: `OpAtomicIIncrement`/`IDecrement` is exact for 0xFFFFFFFF clamp (the common case). The PR body acknowledges the gap for arbitrary clamp values — acceptable tradeoff.

**Tests**: 6 decode tests + 3 end-to-end SPIR-V translation tests. Good coverage of the tricky paths (cmpswap operand order, GLC return, RTN destination wiring, image texel pointer).

No findings. Looks good to merge.

par274:
Need rabase

missatjuhvdk1:
I rebased onto current main. The #200 seam refactor moved the translator files, so the changes now live in SharpEmu.ShaderCompiler and SharpEmu.ShaderCompiler.Vulkan. Two small adjustments came out of that: the test usings point at the new namespaces, and IsDataShareAtomic is public now because its caller sits on the other side of the new assembly boundary (same as IsStorageImageOperation). Full suite passing locally, 87/87.

missatjuhvdk1:
Rebased again on current main, this one was less mechanical because #216 landed in the meantime. What changed in the resolution:

- #216 added its own BufferAtomicAdd/UMax special case and a DsAddU32 case. Those are now folded into the general paths here, so this PR removes them. Behavior is preserved, and DsAddU32 picks up the exec guard it was missing.
- Buffer atomics now run inside the IsBufferWordInRange bounds check that main added, same as loads and stores.
- Image atomic coordinates use the runtime ImageQuerySize clamp like the other storage image paths, since TryGetImageBounds is gone.
- DS atomics take offset0 as a raw byte offset, following the pair/single offset split main made.
- The test harness sets up COMPUTE_PGM_RSRC2 and COMPUTE_USER_DATA registers now that TryCreateState validates the USER_SGPR count.

146/146 tests green locally.

PR #225, opened 2026-07-15
Renderer optimisations

anesr5:
- Removed the scalar register snapshots in
  `Gen5ShaderScalarEvaluator` it was never read anywhere, and it
  cloned 128 registers per instruction on every draw, on the guest thread.
- Descriptor pools are now recycled with `vkResetDescriptorPool` instead of
  being created/destroyed on every translated draw (same pattern as the
  existing `_hostBufferPool`).
- Texture staging buffers go through `CreateHostBuffer`/`RecycleHostBuffer`
  instead of a fresh buffer + allocation per upload.
- CPU pixel work now operates on 32/64-bit words instead of byte-by-byte, and `ScaleBgra` reuses
  its buffer across frames.
- Large transient guest reads use `GC.AllocateUninitializedArray`


par274:
The allocation and pooling changes look useful, but the new ComputeFingerprint implementation is not behavior-preserving. It replaces byte-wise FNV-1a with a different four-lane hash and therefore produces different fingerprints for the same input.

And need rebase.

PR #238, opened 2026-07-15
[HLE] Report offline NP reachability state

xdCloudy:
## What

Implement deterministic offline output for `sceNpGetNpReachabilityState`.

## Why

The export previously returned success without writing the requested reachability state, leaving the guest output value unchanged.

## Implementation

- Validate the output state pointer.
- Report the offline/unavailable NP reachability state (`0`).
- Return a guest memory fault if the output value cannot be written.

## Verification

- `dotnet restore SharpEmu.slnx --locked-mode`
- `dotnet test SharpEmu.slnx -c Release --no-restore` — 26/26 passed
- Synthetic shader validation completed successfully
- `git diff --check` passed


par274:
true emulation

PR #244, opened 2026-07-15
[Input] Map POSIX thumbstick endpoints exactly

999sian:
## What

Map Silk.NET’s Linux/macOS thumbstick range `[-1, 1]` exactly onto SharpEmu’s host input byte range `[0, 255]`:

- `-1` → `0`
- `0` → `128`
- `1` → `255`

The existing clamp remains in place for unexpected out-of-range backend values. A focused xUnit theory covers all three endpoints.

## Why

The previous expression used a scale of `127`, so full negative movement mapped to `1` instead of `0`. Windows XInput and keyboard input already use `0` for full negative movement. This change makes Linux/macOS controller input cover the same complete range without changing deadzones, axis direction, buttons, triggers, or Windows backends.

## Verification

- Before the mapping change, the endpoint theory failed for `-1`: expected `0`, actual `1`.
- After the change, all 3 endpoint cases pass.
- `dotnet test SharpEmu.slnx -c Release --no-build` — 114 passed, 0 failed
- `reuse lint` — REUSE 3.3 compliant, 0 missing licenses

## AI assistance

AI assistance was used for repository research, implementation, and verification. I reviewed the focused change, understand each line, and can explain, modify, debug, and maintain it.

### DIFF
--- a/src/SharpEmu.Libs/Pad/HostWindowInput.cs
+++ b/src/SharpEmu.Libs/Pad/HostWindowInput.cs
@@ -248,9 +248,9 @@ public static class HostWindowInput
         };
     }
 
-    private static byte ToStickByte(float value)
+    internal static byte ToStickByte(float value)
     {
-        return (byte)Math.Clamp((int)(128.0f + value * 127.0f), 0, 255);
+        return (byte)Math.Clamp((int)MathF.Round((value + 1.0f) * 127.5f), 0, 255);
     }
 
     private static HostGamepadButtons MapButton(ButtonName name) => name switch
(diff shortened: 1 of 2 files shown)


kostyaff (approved):
## Review

Fixes thumbstick mapping to reach full symmetric range: `-1.0 → 0`, `0.0 → 128`, `1.0 → 255`.

### Old vs new — ✅

Old: `(byte)Math.Clamp((int)(128.0f + value * 127.0f), 0, 255)`
- `1.0` → 255 ✅
- `-1.0` → 1 ❌ (should be 0)
- `0.0` → 128 ✅

New: `(byte)Math.Clamp((int)MathF.Round((value + 1.0f) * 127.5f), 0, 255)`
- `1.0` → 255 ✅
- `-1.0` → 0 ✅
- `0.0` → 128 ✅

`MathF.Round` (banker's rounding) — `127.5` rounds to `128` ✅.

### Test — ✅

Three test cases covering the three endpoints. Clean and sufficient.

### Visibility change

`ToStickByte` changed from `private` to `internal` — needed for the test. Acceptable.

## Verdict: ✅ Approve

Fixes a real input mapping issue. The old formula could not reach 0 at full deflection.

Issue #194, opened 2026-07-15
macOS (Apple Silicon, Rosetta 2) - Dreaming Sarah crashes ~5s in with rosetta error: ThreadContext::resume failed 37


Trojanhammer:
**Platform:** macOS, Apple Silicon (M4), running the osx-x64 build under Rosetta 2 (built from the linux-macos-support branch)

**Game:** Dreaming Sarah [PPSA02929]

**Command:**
```
./SharpEmu --cpu-engine=native --log-level=info /path/to/eboot.bin
```

**What happens:**
The game loads cleanly, sceAgc/VideoOut initialize, real textures render (title/splash), and background audio plays correctly (correct SFX keys firing: water_splash, water_swim_03, etc., with expected attenuation values). After roughly 5 seconds of active audio + video, the process crashes with:

```
rosetta error: ThreadContext::resume failed 37
[LOADER][INFO] VideoOut presentation window closed
zsh: terminated  ./SharpEmu ...
```

No SharpEmu-side exception or stack trace precedes it - it looks like Rosetta 2's translator failed to resume one of the guest threads (the run has ~20 concurrently scheduled guest threads: background loader workers, audio thread, trophy manager, save thread, etc.).

**Notes:**
- Since CpuExecutionEngine currently only exposes NativeOnly, there's no way to test whether an interpreter-based CPU path avoids this (no such path exists yet).
- I noticed issue #9 for this same game was only tested on Windows/native Vulkan, so this may be a macOS/Rosetta-specific issue nobody's hit yet.
- Happy to attach a full debug-level log if useful.

NickPom:
Reproduced this on a second machine, with some additional data that may narrow it down.

**Environment:** macOS 26.5.2, Apple M4 Pro (Mac16,8), running the `osx-x64` release build under Rosetta 2. Also rebuilt locally from `main` (.NET SDK 10.0.103) to test a fix.

**Games:** Dreaming Sarah and Arcade Game Zone same failure category on both.

### 1. A separate, 100%-reproducible blocker before this one

On this machine, both titles failed *before* a single guest instruction ran:

```
System.InvalidOperationException: Could not allocate main image at required base 0x0000000800000000 (size=0x...).
   at SharpEmu.Core.Loader.SelfLoader.LoadCore(...)
```

[...]

PR #242, opened 2026-07-15
[PlayGo] Fix authoritative chunk enumeration

peter-bonanni:
## Summary

- distinguish authoritative PlayGo chunk sets from sidecars whose chunk IDs cannot be interpreted
- keep metadata-free full installs authoritative with the existing synthetic chunk set `[0]`
- make `scePlayGoGetLocus` write `NotDownloaded` and return `ORBIS_PLAYGO_ERROR_BAD_CHUNK_ID` for the first authoritative unknown ID
- preserve permissive behavior for dat-only, malformed, and otherwise unrecognized metadata
- reset PlayGo static state and diagnostic counters between nonparallel export tests

This restores the terminating response expected by titles that enumerate chunk IDs sequentially, while keeping strict validation scoped to `GetLocus`. Other PlayGo APIs, `ValidateChunkIds`, the import-loop guard, and `playgo-chunk.dat` parsing are unchanged.

## Verification

Using the pinned .NET SDK 10.0.103:

```text
dotnet restore SharpEmu.slnx --locked-mode
dotnet build SharpEmu.slnx -c Release --no-restore
dotnet test SharpEmu.slnx -c Release --no-build
```

Result on rebased PR head `bbf899253ae5dc6ea0a023128d961ccea97d4bcf`: Release build succeeded with 0 warnings and 0 errors; 89/89 `SharpEmu.Libs.Tests` and 33/33 `SharpEmu.SourceGenerators.Tests` passed. Scoped `dotnet format --verify-no-changes` also passed for the two changed files.

The new tests cover the metadata-free `[0]` set, parsed chunk definitions, dat-only/malformed/unrecognized metadata, mixed known/unknown output, chunk/output memory faults, and the `uint.MaxValue` sentinel.

A stacked integration build of `origin/main@864cbb0`, allocator PR #246 (`1b029b5`), and this PlayGo patch (`89d3099`; identical PlayGo source and test content to rebased commit `bbf8992`) also built with 0 warnings and 0 errors and passed 90/90 library tests plus 33/33 source-generator tests.

### San Andreas / macOS x64 with allocator PR #246

- Title: Grand Theft Auto: San Andreas - The Definitive Edition, `PPSA03525`, version `01.006.000`
- Metadata: no PlayGo sidecars, selecting the authoritative synthetic chunk set `[0]`
- Stack: `origin/main@864cbb0`, allocator PR #246 at `1b029b5`, and this PlayGo patch at integration commit `89d3099`
[...]

PR #246, opened 2026-07-15
Fix macOS exact-address allocation without clobbering host memory

peter-bonanni:
## Summary

- replace Darwin's destructive `MAP_FIXED` retry with fixed `mach_vm_allocate` placement that omits `VM_FLAGS_OVERWRITE`
- track Mach-backed regions so release uses `mach_vm_deallocate`
- add a macOS regression test proving an exact guest allocation cannot replace an untracked host mapping

## Root cause

PR #214 made exact PS5 image placement possible on macOS by retrying with `MAP_FIXED` when an `mmap` hint was relocated. `MAP_FIXED` first removes anything already mapped in the requested range, including CLR, dyld, JIT, or Rosetta memory that is not represented in SharpEmu's region table. In the GTA V reproduction this allowed loading to continue temporarily, then the process segfaulted when the first adjacent module was mapped.

Darwin's Mach VM API provides the required no-replace behavior: fixed placement without `VM_FLAGS_OVERWRITE` succeeds only when the entire range is free. A collision now fails cleanly instead of corrupting the host process.

PR #216 reduces the common collision at the 32 GiB main-image base by using Workstation GC, but it does not make `MAP_FIXED` safe for other exact guest allocations. This PR keeps that configuration and focuses only on allocator safety.

## Validation

Using the pinned .NET SDK 10.0.103:

```text
dotnet restore SharpEmu.slnx --locked-mode
dotnet build SharpEmu.slnx -c Release --no-restore
```

- `SharpEmu.Libs.Tests`: 83/83 passed
- `SharpEmu.SourceGenerators.Tests`: 33/33 passed
- macOS collision regression: the allocator returned failure and preserved every byte of an independently mapped sentinel page

Runtime validation used GTA V, PPSA04264 v01.005.000, on the macOS x64/Rosetta path:

- baseline with #214 loaded the main image through `MAP_FIXED`, then terminated with SIGSEGV/exit 139 at the first module mapping
- the safe allocator plus Workstation GC mapped the main image at `0x0000000800000000`, mapped all three adjacent modules at their exact addresses, ran their initializers, and reached the later `BHouLQzh0X0` direct-memory-query loop without a host crash
- after rebasing onto current `main` at `864cbb0`, the main image still maps exactly without a host crash; execution now stops earlier at the newly introduced static-TLS capacity check (`0x13570` required, `0x10000` reserved), which is independent of this allocator change

No game files, extracted assets, or proprietary fixtures are included.

Follow-up to #214. Refs #194.

### DIFF
--- a/src/SharpEmu.HLE/Host/Posix/PosixHostMemory.cs
+++ b/src/SharpEmu.HLE/Host/Posix/PosixHostMemory.cs
@@ -163,12 +163,17 @@ internal sealed unsafe class PosixHostMemory : IHostMemory
         private const int PROT_EXEC = 0x4;
 
         private const int MAP_PRIVATE = 0x02;
-        private const int MAP_FIXED = 0x10;
         private static readonly int MAP_ANON = OperatingSystem.IsMacOS() ? 0x1000 : 0x20;
         private static readonly int MAP_NORESERVE = OperatingSystem.IsMacOS() ? 0 : 0x4000;
 
         // Linux-only: fail instead of clobbering an existing mapping.
         private const int MAP_FIXED_NOREPLACE = 0x100000;
+
+        private const int KERN_SUCCESS = 0;
+
+        // On Darwin fixed placement is represented by the absence of
+        // VM_FLAGS_ANYWHERE. Do not add VM_FLAGS_OVERWRITE: overlap must fail.
+        private const int VM_FLAGS_FIXED = 0;
 
         private static readonly nint MAP_FAILED = -1;
 
@@ -181,6 +186,7 @@ internal sealed unsafe class PosixHostMemory : IHostMemory
             public ulong Size;
             public uint DefaultProtect;
             public Dictionary<ulong, uint>? PageProtects;
+            public bool UsesMachAllocation;
 
             public ulong End => Base + Size;
 
@@ -251,6 +257,7 @@ internal sealed unsafe class PosixHostMemory : IHostMemory
                 }
 
                 nint result;
+                var usesMachAllocation = false;
                 if (address != null)
                 {
                     // Win32 maps at exactly the requested address or fails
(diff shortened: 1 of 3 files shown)


peter-bonanni:
Additional runtime validation from the PlayGo PR #242 stack:

- Stack: `origin/main@864cbb0` + this allocator at `1b029b5` + PlayGo integration commit `89d3099`.
- Pinned SDK 10.0.103 Release build passed with 0 warnings/errors; 90/90 library tests and 33/33 source-generator tests passed.
- San Andreas DE (`PPSA03525` v`01.006.000`) reserved its main executable exactly at `0x0000000800000000`, loaded all six adjacent modules (`loaded=6 failed=0`), and reached guest PlayGo execution under Rosetta rather than stopping at the former host-address blocker.
- The title then cleared its PlayGo enumeration loop and continued through import `#5453340` before a later `FMallocBinned3` unrecognized-free error and access violation at RIP `0x0000000801BF82D6`.

This provides a second title confirming that the allocator change removes the exact-address host blocker without the old host-mapping clobber. The later memory/runtime failure is retained as the next distinct blocker; full logs and all game data remain outside the repository.

kostyaff (approved):
## Review

Replaces `MAP_FIXED` (which silently clobbers existing mappings) with `mach_vm_allocate` on Darwin for exact-address allocation. Prevents the emulator from overwriting CLR/dyld/JIT/Rosetta memory that isn't tracked in the shadow table.

### `AllocateDarwinFixed` — ✅

- `mach_vm_allocate(mach_task_self(), ref address, size, VM_FLAGS_FIXED)` — correct. `VM_FLAGS_FIXED = 0` means "place at exactly this address, fail if overlap".
- Verifies result address matches requested — defensive, correct.
- `mprotect` after allocation to set correct protection — `mach_vm_allocate` defaults to `PROT_READ|PROT_WRITE`.
- On `mprotect` failure: deallocates and returns `MAP_FAILED` — correct cleanup.

### `Free` — ✅

`region.UsesMachAllocation ? mach_vm_deallocate : munmap` — correct dispatch. Removes from `Regions` only on success.

### P-Invoke declarations — ✅

`mach_task_self()`, `mach_vm_allocate()`, `mach_vm_deallocate()` from `libSystem.B.dylib`. Correct for macOS.

### Test — ✅

`ExactAllocationDoesNotReplaceUntrackedDarwinMapping` — creates external mapping, fills with sentinel 0xA5, attempts `Allocate` at same address, asserts failure and sentinel intact. Skips on non-macOS.

### `InternalsVisibleTo` — ✅

Added `SharpEmu.Libs.Tests` to `SharpEmu.HLE` IVT for test access.

## Verdict: ✅ Approve

Significant safety improvement for macOS users. The old `MAP_FIXED` path could silently corrupt host memory. `mach_vm_allocate` with `VM_FLAGS_FIXED` is the correct Darwin API.

No overlap with PR #248 — #248 is Windows (GC heap reservation), #246 is macOS (safe fixed placement).

PR #249, opened 2026-07-16
cpu: emulate BMI1/BMI2/ABM instructions in software when the host lac…

MikeyLITE69:
…ks them

## What

The native backend runs guest code directly on the host CPU. When the host doesn't implement a BMI1/BMI2/ABM instruction that the PS5's Zen 2 cores do, it raises #UD (STATUS_ILLEGAL_INSTRUCTION). Today the vectored handler just logs the faulting bytes and gives up, so the title dies.

This adds a software fallback. On an illegal-instruction fault we decode the opcode with Iced (the decoder already used elsewhere in the backend), evaluate it against the trapped register/memory state, write the result and flags back into the CONTEXT record, step RIP past the instruction, and resume.

Instructions covered (32- and 64-bit): ANDN, BLSI, BLSMSK, BLSR, BEXTR, BZHI, TZCNT, LZCNT, RORX, SARX, SHLX, SHRX, PDEP, PEXT.

## Why

Users on CPUs without these extensions currently can't get past code that uses them. This is a generic fix (no game-specific hacks) that improves compatibility on older hosts. MULX is intentionally left out for now — its dest_hi/dest_lo operand ordering is easy to get subtly wrong, so I'd rather add it separately with its own tests.

## How it's structured

- `BmiInstructionEmulator` holds the pure bit/flag semantics with no dependency on the unsafe CONTEXT plumbing, so it can be unit-tested directly.
- `DirectExecutionBackend.IllegalInstruction.cs` is the thin unsafe adapter (decode → read operands → emulate → write back → advance RIP). Anything it doesn't fully model returns false and falls through to the existing diagnostics unchanged, so it can never mis-handle an opcode it doesn't recognize.
- One hook in `DirectExecutionBackend.Exceptions.cs`, next to the other TryRecover* calls.
- Emits a single one-time "emulating in software" log line, not per-instruction spam.

## How I verified

- Added xUnit tests covering every instruction in both widths plus the CF/ZF/SF/OF edge cases (src == 0, shift-count masking, index beyond operand width, etc.).
- Cross-checked all the expected values against an independent reference implementation written from the Intel/AMD definitions; results match.
- `dotnet build` + `dotnet test` pass locally.

## Notes

New files follow .editorconfig (4-space, SPDX headers, REUSE-compliant).

### DIFF
--- a/src/SharpEmu.Core/Cpu/Emulation/BmiInstructionEmulator.cs
+++ b/src/SharpEmu.Core/Cpu/Emulation/BmiInstructionEmulator.cs
@@ -1,0 +1,305 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+using System.Numerics;
+
+namespace SharpEmu.Core.Cpu.Emulation;
+
+/// <summary>
+/// Pure software implementations of the BMI1, BMI2 and ABM general-purpose-register
+/// bit-manipulation instructions.
+///
+/// The direct-execution backend runs guest PS5 code natively on the host CPU. The PS5's
+/// Zen 2 cores implement BMI1/BMI2/ABM, but a host CPU that predates those extensions raises
+/// #UD (STATUS_ILLEGAL_INSTRUCTION) when it meets one of these opcodes. This class provides the
+/// register-only arithmetic so the exception handler can finish the instruction in software and
+/// resume, instead of aborting the title.
+///
+/// The methods deliberately operate on plain integers rather than on the OS CONTEXT record so the
+/// semantics can be unit-tested in isolation; the unsafe register/memory plumbing lives in the
+/// backend adapter. Each method returns the result already masked to the operand width and, where
+/// the instruction is defined to touch flags, updates <paramref name="eflags"/> in place. Flags
+/// documented as "undefined" by the vendor manuals are left untouched so behaviour stays
+/// deterministic across hosts.
+/// </summary>
+public static class BmiInstructionEmulator
+{
+    private const uint FlagCarry = 1u << 0;
+    private const uint FlagZero = 1u << 6;
+    private const uint FlagSign = 1u << 7;
+    private const uint FlagOverflow = 1u << 11;
+
+    private static ulong WidthMask(GprOperandSize size) =>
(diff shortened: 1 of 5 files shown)


kostyaff (approved):
## Review

Software fallback for BMI1/BMI2/ABM instructions when host CPU lacks those extensions. Guest code raises #UD, this PR decodes the faulting instruction, evaluates it, writes result back to CONTEXT, steps RIP, resumes.

### `BmiInstructionEmulator.cs` (pure math) — ✅

Excellent separation: pure methods on plain integers, no CONTEXT dependency. Correct flag handling (CF/ZF/SF/OF) per Intel/AMD manuals. Width masking correct for 32 and 64-bit. SARX correctly handles arithmetic shift (sign extension) for both widths. PDEP/PEXT use bit-by-bit loop — correct, fine for #UD fallback path.

Covers: ANDN, BLSI, BLSMSK, BLSR, BEXTR, BZHI, RORX, SARX, SHLX, SHRX, PDEP, PEXT, LZCNT, TZCNT, POPCNT.

### `DirectExecutionBackend.IllegalInstruction.cs` — ✅

- Uses Iced for decoding (already a dependency)
- Decode window shrinks 15→11→8→4→2 bytes for page-boundary faults — robust
- `TryReadOperand` handles both register and memory operands
- `TryComputeMemoryAddress`: base+index*scale+displacement and IP-relative addressing
- FS/GS segment prefixes correctly rejected (guest segment base not modelled)
- "Return false for anything not fully modelled" design ensures safety — unknown opcodes fall through to existing diagnostics

### Tests — ✅ good coverage

ANDN (bit clearing, zero/sign flags, direction flag preservation), BLSI/BLSMSK/BLSR (carry semantics), BEXTR (edge cases), BZHI, RORX/SARX/SHLX/SHRX (masking), PDEP/PEXT, LZCNT/TZCNT/POPCNT. Both 32 and 64-bit widths tested.

## Verdict: ✅ Approve

Well-structured, well-tested, correct approach.

**Non-blocking notes:**
- CI has not run yet (`mergeStateStatus: BLOCKED`) — verify CI passes before merge
- This is Windows x64 only (POSIX would need a SIGILL handler) — worth a brief comment
- No integration test for the full #UD → recover → resume path, but acceptable for an optional fallback

par274:
its good.

PR #251, opened 2026-07-16
Boot and render compatibility fixes

StealUrKill:
## Summary

Compatibility fixes that get **DS** to render its intro
sequence and improve boot behavior for unpackaged UE.
Three self-contained commits. A boot-compatibility batch, a shader decoder
fix, and Vulkan device-loss resilience.

### `[Boot]` Compatibility for UE titles + HLE exports
- **PlayGo:** `scePlayGoGetLocus` returns `BAD_CHUNK_ID` for ids outside the
  known set, so titles that enumerate chunk ids until that error terminate
  instead of scanning the full `ushort` range forever. Missing/unreadable
  metadata falls back to a single installed chunk 0 (never an empty set, which
  would make every id "valid").
- **Kernel:** `SHARPEMU_WRITABLE_APP0` opt-in (default off). Unpackaged UE dumps
  write their `Saved/` tree under `/app0` during PS5-component init and treat
  the denial as a fatal boot error.
- **Pad:** accept handle 0 as the primary pad; some titles read pad state with
  handle 0 and reject a never-valid state otherwise.
- **VideoOut:** transition offscreen guest images through their whole mip chain
  at creation, so full-chain sampled binds never read an `Undefined` layout.
- **HLE exports:** `sceSystemServiceGetNoticeScreenSkipFlag`,
  `sceSystemServiceGetMainAppTitleId`, `sceNpWebApi2CreateUserContext`.
- **GUI:** `SHARPEMU_LOG_IO` and `SHARPEMU_WRITABLE_APP0` toggles in the
  Environment tab.

### `[ShaderCompiler]` Fix `VReadlaneB32` scalar destination field
The scalar destination lives in the low `vdst` byte (bits 0–7); it was read
from bits 8–14 (the VOP3B carry-out field, which this opcode lacks), sending
**every** `v_readlane` result to `s0`. Verified against raw gfx10 encodings and
LLVM's assembler tests.

### `[VideoOut]` Survive device loss and flip-order asserts without dying
- Handle `VK_ERROR_DEVICE_LOST` without unwinding the window thread: a typed
  `VulkanDeviceLostException`, an early drain of queued guest work, and
  retirement of fences that can never signal after loss. The window keeps
  pumping so the game continues instead of the process terminating.
- Downgrade the flip-order `Debug.Assert` (fires ~100×/run on Demon's Souls and
  fail-fasts debug builds with nothing logged) to a once-only warning.

## Testing
- Solution build (Windows) + `SharpEmu.Libs.Tests` (83/83 passing).
- Manual boot runs of Demon's Souls (renders the intro logo) and Monster Truck
  Championship.

## Notes for reviewers
- The `[Boot]` commit also includes **env-gated diagnostic/experiment code**
  (Bluetooth-HID callback hooks, an RTC busy-wait caller probe) that is off by
  default. Happy to drop these into a separate change or remove them if you'd
  prefer the PR carry only the verified fixes.
- The three commits are independent; I can split this into separate PRs if that
  reviews more easily.


src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
```diff
@@ -4806,8 +4806,19 @@ private static void EnsureOpenParentDirectoryExists(string guestPath, string hos
...
+    // /app0, which is read-only on retail hardware. Opt in via SHARPEMU_WRITABLE_APP0=1
+    // to allow those writes so such dumps can boot; defaults off to keep retail semantics.
+    private static readonly bool _writableApp0 =
+        string.Equals(Environment.GetEnvironmentVariable("SHARPEMU_WRITABLE_APP0"), "1", StringComparison.Ordinal);
+
     public static bool IsReadOnlyGuestMutationPath(string guestPath)
     {
+        if (_writableApp0)
+        {
+            return false;
```
peter-bonanni, line 4819: **[P1] Contain writable `/app0` paths before bypassing the guard.** With `SHARPEMU_WRITABLE_APP0=1`, this disables the mutation check while `ResolveGuestPath` merely combines an untrusted suffix. A path such as `/app0/../../victim` therefore escapes `SHARPEMU_APP0_DIR` and can be created, overwritten, or deleted. Canonicalize with `Path.GetFullPath` and enforce root containment before allowing mutation.

src/SharpEmu.Libs/SystemService/SystemServiceExports.cs
```diff
@@ -15,6 +15,62 @@ public static class SystemServiceExports
...
+        var flagAddress = ctx[CpuRegister.Rdi];
+        if (flagAddress == 0)
+        {
+            return ctx.SetReturn(OrbisSystemServiceErrorParameter);
+        }
+
+        // No system notice screen to skip in the emulator; report "do not skip".
+        Span<byte> flagBytes = stackalloc byte[sizeof(int)];
+        BinaryPrimitives.WriteInt32LittleEndian(flagBytes, 0);
+        return ctx.Memory.TryWrite(flagAddress, flagBytes)
```
peter-bonanni, line 43: **[P2, non-blocking] Write one byte for the notice flag.** A public DOLL capture reproduced this mismatch: the caller passed a one-byte field immediately below its stack canary, the four-byte write smashed the canary and reached `__stack_chk_fail`, and a one-byte write was boot-verified afterward. Use a single-byte write and add a boundary-pointer regression test. [Evidence](https://github.com/mattias800/ps5ys/pull/313).

src/SharpEmu.Libs/PlayGo/PlayGoExports.cs
```diff
@@ -682,12 +690,23 @@ private static PlayGoMetadata LoadPlayGoMetadata()
...
             return new PlayGoMetadata(true, [(ushort)0]);
         }
 
         var chunkIds = LoadChunkIds(chunkDefsXml);
+        if (chunkIds.Length == 0)
+        {
+            // Unreadable/empty sidecar: fall back to chunk 0, not an empty set
+            // (which IsKnownChunkId would treat as "every id valid").
+            TracePlayGo("metadata_unreadable; fully-installed single chunk");
+            return new PlayGoMetadata(true, [(ushort)0]);
```
peter-bonanni, line 707: **[P1] Preserve unknown PlayGo metadata.** The code recognizes `playgo-chunk.dat` and scenario JSON as metadata, but only parses XML. Dat-only, malformed, or unrecognized packages therefore become authoritative chunk 0, causing every valid nonzero chunk to return `BAD_CHUNK_ID`. Preserve an `Unknown` state until the metadata can actually be parsed.

peter-bonanni (changes requested):
Requesting changes for two blockers: writable `/app0` must remain contained within its host root, and unparsed PlayGo metadata must not become authoritative chunk 0. The notice-flag width is retained as a non-blocking ABI correction. Reviewed at exact head `e08ffedd4113e181a879a6c936637f830ccd500d`.

kostyaff (approved):
## Review

3 well-structured, self-contained commits. Reviewed commit by commit.

### Commit 1: Boot compatibility fixes — ✅

- **PlayGo BAD_CHUNK_ID**: `scePlayGoGetLocus` now rejects chunk IDs outside the package set. Fixes Monster Truck Championship scanning 0,1,2,... forever. `IsKnownChunkId` no longer treats empty chunk set as "everything valid" — good fix. Fallback to chunk 0 instead of empty set when sidecar unreadable is correct.
- **Pad handle 0**: `IsPrimaryPadHandle(handle) => handle is 0 or PrimaryPadHandle` — correct. Real firmware hands out small non-negative handles, 0 is valid.
- **Writable app0**: Opt-in via `SHARPEMU_WRITABLE_APP0=1`. Defaults off → retail semantics preserved. Clean.
- **NpWebApi2 createUserContext**: Returns InvalidArgument — makes title online layer back off. Good.
- **SystemService**: Two new exports (`getMainAppTitleId`, `getNoticeScreenSkipFlag`). Both check null pointer, bounded writes. `ConfigureApplicationInfo` wired into runtime init correctly.
- **Rtc probe**: Opt-in diagnostic. Zero cost when unset, fires once. Fine.
- **BluetoothHid experiment**: Opt-in via env vars. `AllocHGlobal` leak is acceptable — `_fired` guard ensures one allocation per process.
- **GUI toggles**: Clean wiring in AXAML + code-behind + en.json.

### Commit 2: VReadlaneB32 scalar destination field — ✅ (fixes bug from #237)

This commit fixes a bug introduced by merged PR #237. Our #237 decoded VReadlaneB32 destination as `Scalar((word >> 8) & 0x7F)` (SDST field, bits 8-14). This PR correctly changes it to `Scalar(word & 0xFF)` (VDST field, bits 0-7).

V_READLANE_B32 is VOP3A, not VOP3B. In VOP3 encoding:
- VDST (bits 0-7) is the destination — for readlane, the SGPR index
- SDST (bits 8-14) is only used by VOP3B instructions (carry-out). Readlane is NOT VOP3B
- For non-VOP3B, bits 8-14 are control/modifiers, not a scalar destination

Our comment "Bits 0-7 are unused for this opcode" was incorrect — bits 0-7 ARE the destination. This PR is correct. The emission part of #237 (GroupNonUniformBroadcast) remains valid and is untouched here.

### Commit 3: Vulkan device-loss resilience — ✅

- Typed `VulkanDeviceLostException` instead of message-string matching — good improvement
- `Render()` wraps `RenderCore()` in try/catch, keeps window pumping on device loss
- `CollectCompletedGuestSubmissions` handles device loss: retires submissions without fence wait
- Flip-order assert → warn-once log (Demon's Souls triggers it) — reasonable

## Verdict: ✅ Approve

Well-commented, well-structured. The VReadlaneB32 fix corrects our #237 decode bug. Device-loss resilience is defensive and correct.

StealUrKill:
Working on these plus a few other changes. Will fix some other issues found as well. Don't fold this in yet.

StealUrKill:
This was merged with another. No big deal. Closing out and uploading a few race conditions for these.

PR #248, opened 2026-07-16
[CLI] Reserve guest address space before GC heap initialization (#235)

jackby03:
The .NET GC may reserve up to 256 GB of virtual address space on systems with limited RAM (e.g. 16 GB), which can overlap the fixed PS5 image base at 0x800000000. This causes SelfLoader to fail with 'Could not allocate main image at required base'.

Added a ModuleInitializer that reserves the critical guest address region (0x800000000 - 0x900000000, 4 GB) as early as possible, before the GC heap is established. The reservation is released just before the loader needs it.

Changes:
- Added GuestAddressSpaceReservation.cs with ModuleInitializer
- Release reservation before SharpEmuRuntime.CreateDefault()

Fixes #235

PR #253, opened 2026-07-16
[HLE/Kernel/Pad/Ngs2] Fix semaphore race condition, add main-thread diagnostics, implement missing pad/ngs2 imports

### LINKED ISSUE #254: GC Deadlock: IL2CPP threads blocked on SuspendSemaphore/work semaphores (Arcade Game Zone)
## Symptoms
- Game: Arcade Game Zone (PPSA19015) - Unity/IL2CPP
- 13 AssetGarbageCollectorHelper threads blocked on sceKernelWaitSema (work semaphores)
- 1 coordinator thread blocked on sceKernelWaitSema (SuspendSemaphore)
- Main thread blocked on handshake semaphore
- Stall detected after 20s, exit code 4

## Log evidence

```
[LOADER][ERROR] Stall import-stub: rip=0x00006FFFFD000EE0 nid=Zxa0VhQVTsk -> libKernel:sceKernelWaitSema
[LOADER][ERROR] Stall guest-thread: handle=0x00000207AC514E50 name='AssetGarbageCollectorHelper' state=Blocked imports=24 nid=Zxa0VhQVTsk block=sceKernelWaitSema
[LOADER][ERROR] Stall guest-thread: handle=0x00000207AB5CD9A0 name='Thread-207AB5CD9A0' state=Blocked imports=38 nid=Zxa0VhQVTsk rdi=0x0000000000000083 block=sceKernelWaitSema
[LOADER][ERROR] Stall main-thread: managed_id=2 state=Blocked block=sceKernelWaitSema:0x00000081:Baselib_SystemSemaphore
```

## Analysis
Circular deadlock in IL2CPP GC suspension: coordinator waits for workers to signal SuspendSemaphore, but workers wait for coordinator to signal work semaphores. No progress possible.

## Possible causes
- sceKernelSuspendScheduler not implemented
- Race condition in semaphore implementation (fixed in #253 but didn't resolve this)
- Workers should be woken via a mechanism not yet supported

jackby03:
## Summary

Fixes the semaphore race condition between `RequestCurrentThreadBlock` and `WakeBlockedThreads`, adds main-thread stall diagnostics, implements missing HLE imports, fixes the Unity IL2CPP GC deadlock, and resolves several correctness issues flagged in review.

---

### Semaphore race fix
Between `RequestCurrentThreadBlock` (import handler) and `TryYieldGuestThreadToHostStub`, a signal can arrive. Current code unconditionally yields, blocking the thread even though the wake condition is already satisfied. Fix: re-evaluate the wake predicate and skip yield if satisfied.
- `DispatchImport`: re-check `blockWaiter.TryWake()`
- `TryDispatchLeafImport`: re-check `blockWakeHandler()`
- `KernelSemaphoreCompatExports`: re-check `Count >= needCount` under the gate

### GC Deadlock fix (Unity IL2CPP)
Arcade Game Zone deadlocks because the GC coordinator raises `sceKernelRaiseException(type=0x1E)` on the main thread while it is blocked in a host semaphore wait. The exception is queued but never delivered since the thread never reaches an HLE boundary.
- Track pending exception state + blocked semaphore gate in `HostMainThread`
- When an exception targets the main thread, set flag and pulse gate so the wait returns early
- `DeliverPendingGuestExceptionAtSafePoint` also clears the flag to prevent spurious wakeups
- Semaphore count is never fabricated or consumed -- wait returns OK without touching `Count`

### Pending exception flag race fix
When the exception was delivered via the normal safe-point path (instead of the host-semaphore early-return), the `PendingException` flag was left set, causing the next wait to spuriously return OK. This caused the game loop to run unthrottled and exhaust memory.
- `DeliverPendingGuestExceptionAtSafePoint` now clears the flag when delivering for the main thread

### POSIX ABI wrappers (open/fstat)
`open()` and `fstat()` now follow the POSIX ABI: return -1 and set errno on failure instead of passing Orbis error codes directly, which callers checking `fd == -1` treat as success.

### Ngs2 rack handles
Rack handles are now per-system sequential IDs starting at 0, matching the convention other SCE handle pools use. Dead Cells hardcodes rack handle 0 for its first rack.

### Missing HLE imports
| Export | NID | Library |
|--------|-----|---------|
| `scePadGetHandle` | `u1GRHp+oWoY` | libScePad |
| `scePadDeviceClassGetExtendedInformation` | `AcslpN1jHR8` | libScePad |
| `sceNgs2SystemCreate` | `koBbCMvOKWw` | libSceNgs2 |
| `sceNgs2RackCreate` | `cLV4aiT9JpA` | libSceNgs2 |
| `sceImeKeyboardGetInfo` | `VkqLPArfFdc` | libSceIme |
| `sceSysmoduleGetModuleInfoForUnwind` | `4fU5yvOkVG4` | libSceSysmodule |
| `unlink` | `VAzswvTOCzI` | libKernel |

### Diagnostics
- `HostMainThread` tracks main thread blocking state for stall snapshots
- Main-thread import tracing with call graphs during stall detection
- Logging gated behind `SHARPEMU_LOG_GUEST_THREADS=1`

### Other fixes
- `GuestAddressSpaceReservation`: reserve guest address space before GC heap (with TOCTOU race documented)
- `PhysicalVirtualMemory`: POSIX over-allocation scoped to macOS + env override (`SHARPEMU_POSIX_FIXED_ALLOC=1`)
- SaveData/SaveDataDialog exports for Dreaming Sarah
- Test updates for POSIX over-allocation guard

---

## Results

| Title | Status |
|-------|--------|
| Arcade Game Zone (PPSA19015) | Boots, renders 1920x1080, 4+ GC cycles |
| Dreaming Sarah (PPSA02929) | Boots, loads assets, plays audio |
| Dead Cells (PPSA15552) | Past audio init with sequential rack handles |

## Refs
- Closes #248 (reconciled overlap)
- Fixes #254 (GC deadlock)
- Related to #256, #259


kostyaff (approved):
## Review

9 files, +406/−22. Semaphore race-condition fix + missing HLE imports + main-thread diagnostics + guest address space reservation.

## ⚠️ Overlap with PR #248 — needs reconciliation

PR #248 (same author) already adds `GuestAddressSpaceReservation.cs` with identical `[ModuleInitializer]` logic. This PR creates the same file — they will conflict.

PR #248 also includes a `PhysicalVirtualMemory.cs` behavioral change (`OperatingSystem.IsMacOS()` → `!OperatingSystem.IsWindows()`) that is NOT in this PR.

**Recommendation:** Close #248, rebase its `PhysicalVirtualMemory.cs` change onto this PR. Or drop `GuestAddressSpaceReservation.cs` from #253 and rebase the rest onto #248. Either way, one must be reconciled.

## Semaphore race-condition fix — ✅

Between `RequestCurrentThreadBlock` (import handler) and `TryYieldGuestThreadToHostStub`, a signal can arrive. Current code unconditionally yields, blocking the thread even though the wake condition is already satisfied.

Fix: re-evaluate the wake predicate. If satisfied, set RAX from `blockWaiter.Resume()` / `blockResumeHandler()` and skip yield.

- `DispatchImport`: `if (blockWaiter is not null && blockWaiter.TryWake())` — correct
- `TryDispatchLeafImport`: `if (blockWakeHandler is not null && blockWakeHandler())` — correct
- `KernelSemaphoreCompatExports`: re-checks `semaphore.Count >= needCount` under the gate — correct, consumes tokens if available

`TryConsumeCurrentThreadBlock` after late-arrival — correct, cancels pending cooperative block.

## Missing HLE imports — ✅

- `sceNgs2SystemCreate` (Nid `koBbCMvOKWw`) — delegates to allocator variant logic. ✅
- `sceNgs2RackCreate` (Nid `cLV4aiT9JpA`) — validates system handle, creates rack. ✅
- `scePadGetHandle` (Nid `u1GRHp+oWoY`) — returns `PrimaryPadHandle` for port 0 / standard type / primary user. ✅
- `scePadDeviceClassGetExtendedInformation` (Nid `AcslpN1jHR8`) — writes 20 bytes (standard DualSense, no extended class data). ✅

## Main-thread diagnostics — ✅ with notes

`HostMainThread` gets tracking fields: `_managedThreadId`, `_lastRip`, `_lastImportNid`, `_isBlocked`, `_blockReason`.

**Non-blocking:**
- Fields are not `volatile` or `Interlocked`-protected. Acceptable for diagnostics-only, but worth a brief comment noting best-effort.
- `Console.Error.WriteLine` in `SetBlocked` always logs (not gated by env var). Only on state transitions, so acceptable.

## Verdict: ✅ Approve (with reconciliation request)

The race-condition fix is correct and addresses a real concurrency bug. Missing imports are well-implemented. **Author must reconcile the `GuestAddressSpaceReservation.cs` overlap with PR #248** before merge.

jackby03:
> ## Review
> 9 files, +406/−22. Semaphore race-condition fix + missing HLE imports + main-thread diagnostics + guest address space reservation.
> 
> ## ⚠️ Overlap with PR #248 — needs reconciliation
> PR #248 (same author) already adds `GuestAddressSpaceReservation.cs` with identical `[ModuleInitializer]` logic. This PR creates the same file — they will conflict.
> 
> PR #248 also includes a `PhysicalVirtualMemory.cs` behavioral change (`OperatingSystem.IsMacOS()` → `!OperatingSystem.IsWindows()`) that is NOT in this PR.
> 
> **Recommendation:** Close #248, rebase its `PhysicalVirtualMemory.cs` change onto this PR. Or drop `GuestAddressSpaceReservation.cs` from #253 and rebase the rest onto #248. Either way, one must be reconciled.
> 
> ## Semaphore race-condition fix — ✅
> Between `RequestCurrentThreadBlock` (import handler) and `TryYieldGuestThreadToHostStub`, a signal can arrive. Current code unconditionally yields, blocking the thread even though the wake condition is already satisfied.
> 
> Fix: re-evaluate the wake predicate. If satisfied, set RAX from `blockWaiter.Resume()` / `blockResumeHandler()` and skip yield.
> 
> * `DispatchImport`: `if (blockWaiter is not null && blockWaiter.TryWake())` — correct
> * `TryDispatchLeafImport`: `if (blockWakeHandler is not null && blockWakeHandler())` — correct
> * `KernelSemaphoreCompatExports`: re-checks `semaphore.Count >= needCount` under the gate — correct, consumes tokens if available
> 
> `TryConsumeCurrentThreadBlock` after late-arrival — correct, cancels pending cooperative block.
> 
> ## Missing HLE imports — ✅
> * `sceNgs2SystemCreate` (Nid `koBbCMvOKWw`) — delegates to allocator variant logic. ✅
> * `sceNgs2RackCreate` (Nid `cLV4aiT9JpA`) — validates system handle, creates rack. ✅
> * `scePadGetHandle` (Nid `u1GRHp+oWoY`) — returns `PrimaryPadHandle` for port 0 / standard type / primary user. ✅
> * `scePadDeviceClassGetExtendedInformation` (Nid `AcslpN1jHR8`) — writes 20 bytes (standard DualSense, no extended class data). ✅
> 
> ## Main-thread diagnostics — ✅ with notes
> `HostMainThread` gets tracking fields: `_managedThreadId`, `_lastRip`, `_lastImportNid`, `_isBlocked`, `_blockReason`.
> 
> **Non-blocking:**
> 
> * Fields are not `volatile` or `Interlocked`-protected. Acceptable for diagnostics-only, but worth a brief comment noting best-effort.
> * `Console.Error.WriteLine` in `SetBlocked` always logs (not gated by env var). Only on state transitions, so acceptable.
> 
> ## Verdict: ✅ Approve (with reconciliation request)
> The race-condition fix is correct and addresses a real concurrency bug. Missing imports are well-implemented. **Author must reconcile the `GuestAddressSpaceReservation.cs` overlap with PR #248** before merge.

Thanks for the review. Here's what I changed:

- Closed #248 and brought its !OperatingSystem.IsWindows() change over to PhysicalVirtualMemory.cs along with a comment explaining the POSIX/Rosetta case.
- Added a note that the HostMainThread fields are best-effort (not volatile), and gated the blocking logs behind SHARPEMU_LOG_GUEST_THREADS=1.
- Also slipped in the missing SaveData exports (sceSaveDataMount/2, sceSaveDataInitialize/2, dialog param init) since I was already touching imports.

Should be good now, let me know if there's anything else.

xnetcat (changes requested):
Strong PR overall — the lost-wakeup race is real and the guard design (re-check under the gate after the block request, cancel the pending cooperative block on late arrival) is the right shape, mirrored correctly at both dispatch paths. The main-thread stall diagnostics and the NID additions (proper out-param structs, sensible error codes) are also welcome. Requesting changes on a few correctness points before merge:

**1. Semaphore state corruption in the pending-exception paths (`WaitSemaphoreOnHostThread`)**
- The early-return path does `semaphore.Count -= needCount` without checking `Count >= needCount` — `Count` can go negative.
- The in-loop path does `semaphore.Count = Math.Max(semaphore.Count, needCount)` — this fabricates tokens that were never signaled, then consumes them and returns OK. The caller now believes it acquired the semaphore; the producer/consumer balance is permanently off by `needCount`. Under repeated GC cycles this drifts further each time.
- Suggestion: deliver the exception via an interrupt-style return (re-enter the wait after the safe-point delivers, or return a distinct status the import layer understands) rather than pretending the acquire succeeded. The semaphore's counter should only ever change by real signal/wait pairs.
- Related: `_pendingException` is only cleared on this path. If the exception gets delivered at a normal safe-point first (thread was briefly unblocked), the stale flag makes the *next* wait spuriously "succeed" through the same fabrication path.

**2. Speculative wake checks — confirm predicate purity**
The non-leaf guard calls `blockWaiter.TryWake()` and the leaf guard calls `blockWakeHandler()`. If either has consume semantics (e.g. decrements a token that the scheduler's wake path would also consume when the continuation fires), calling it speculatively here double-consumes. If both are pure predicates, a comment saying so would prevent future regressions; if not, this needs a consume-once guarantee shared with `RegisterBlockedGuestThreadContinuation`.

**3. Discarded result of `TryConsumeCurrentThreadBlock` in the late-arrival branch**
`_ = GuestThreadExecution.TryConsumeCurrentThreadBlock(out _, ...)` ignores whether the block was actually consumed. If the scheduler raced and already consumed/parked it, silently proceeding can leave a registered continuation that fires later against a thread that already moved on. At minimum check the bool and trace/assert the unexpected case.

**4. Ungated hot-path logging**
The `[MAIN_THREAD] Import#...` writes in `DispatchImport` (sema NIDs) are unconditional `Console.Error` in the hottest path in the emulator, while `HostMainThread.SetBlocked` correctly gates behind `SHARPEMU_LOG_GUEST_THREADS`. Please gate these the same way.

**5. `GuestAddressSpaceReservation` timing caveat**
`[ModuleInitializer]` runs well after CLR startup — the GC's *initial* heap reservation has already happened by then, so this guards against lazy heap growth rather than initial placement (worth correcting the comment). Also the `ReleaseReservation()` → loader-allocation window is itself a race: the GC can grab the range between the free and the loader's reserve. On Win10+ consider placeholder handoff (`VirtualAlloc2` + `MapViewOfFile3`/placeholder replacement) for an atomic transfer, or at least document the residual race. `GCRetainVM=false` + a documented `GCHeapHardLimit` would be a more deterministic belt-and-braces.

**6. `PhysicalVirtualMemory`: macOS → !Windows**
This silently changes Linux from exact-address search to over-allocation (extra `alignment` bytes reserved per allocation). If that's intentional for Rosetta-for-Linux/Docker scenarios, say so in the comment; otherwise scope it to macOS + an env override.

**7. `HostMainThread` shared state**
`BlockedSemaphoreGate` / `ExternalGuestThreadHandle` are plain mutable statics read/written across threads (exception-raiser vs main thread), and the single-slot gate can't represent nested waits. Fine for the diagnostic fields (as commented), but the gate/pending-exception pair is correctness-bearing — please make them `volatile`/`Interlocked` and document the single-outstanding-wait assumption.

The Ngs2/Pad/SaveDataDialog additions look correct as stubs (out-params written, plausible error codes). With items 1–3 addressed I'd be happy to approve.

NickPom:
Tested this branch against Arcade Game Zone on macOS (Apple Silicon, Rosetta 2). Good news: with two additional small fixes on top, the game now fully boots, initializes the entire Unity job system (Job.worker 0-12, Background Job.worker 0-15, GfxFlipThread, UnityGfxDeviceWorker, FMOD audio), and renders real frames (splash + first gameplay frame at 1920x1080). Writing up what I found in case it saves time.

**Fix 1: WakeBlockedThreads never checks BlockWakeHandler**

`WakeBlockedThreads` (DirectExecutionBackend.cs) only checks `thread.BlockWaiter?.TryWake()` before marking a thread Ready. Threads parked via the other `RegisterBlockedGuestThreadContinuation` overload (BlockResumeHandler/BlockWakeHandler, used by sceKernelWaitEventFlag) have `BlockWaiter == null`, so that check is silently skipped and the thread gets marked Ready without ever confirming its actual wait condition. It resumes still unsatisfied.

Fix, mirrors what PR #164 already does for this same line (that PR predates the BlockWaiter mechanism from #67 though, so it does not know about it and would need reconciling, not a straight cherry-pick):

```csharp
if (thread.BlockWaiter is not null && !thread.BlockWaiter.TryWake())
{
    continue;
}
if (thread.BlockWakeHandler is not null && !thread.BlockWakeHandler())
{
    continue;
}
```

This alone got Arcade Game Zone from stalling at import 81408 (first GC cycle, as this PR already handles) to import 158720+ and multiple completed GC cycles.

**Fix 2: open()/fstat() don't follow the POSIX ABI on failure**

Same class of bug as the existing `stat`/`PosixStat` wrapper already documents in KernelMemoryCompatExports.cs ("Returning the raw Orbis kernel code here makes callers treat a missing file as a non-negative success value"). `open` (wuCroIGjt2g) and `fstat` (mqQMh1zPPT8) call straight into the Orbis-convention KernelOpenUnderscore/KernelFstat without the same conversion `stat` gets. A failed open returns ORBIS_GEN2_ERROR_NOT_FOUND (0x80020002) directly in RAX instead of -1+errno, so a caller checking `fd == -1` doesn't see the failure and goes on to call fstat/lseek/close/read with 0x80020002 as the fd, eventually dereferencing something that was never populated. Wrapped both the same way PosixStat already does (return -1, set errno via KernelRuntimeCompatExports.TrySetErrno, leave the Orbis-convention sceKernelOpen/sceKernelFstat untouched since those callers expect the raw code).

This is what was crashing the emulator further in (unrecovered SIGSEGV, fault address 0x0) once fix 1 got past the first deadlock.

**Where it's stuck now, and this is the same thing you already flagged in review**

With both fixes in, the game gets past two full GC stop-the-world cycles and starts rendering, then the main thread deadlocks again on the same SetEventFlag -> block-on-ResumeSemaphore pattern, just on a later cycle. Traced this to review point #1 on `WaitSemaphoreOnHostThread`: the in-loop pending-exception path does `semaphore.Count = Math.Max(semaphore.Count, needCount)` then `semaphore.Count -= needCount` right after, which discards whatever real count was already accumulated instead of preserving it. First cycle happens to net out fine since Count is usually 0 going in; once real signal/wait pairs are interleaved with this fabrication across multiple cycles the balance drifts and a later cycle's wait never resolves through a real signal. Matches your note "under repeated GC cycles this drifts further each time" exactly, just confirmed concretely against a real title.

Didn't attempt that fix since it needs the interrupt-and-retry redesign you already outlined (re-enter the wait after the safe-point delivers), and that touches the exception-context/CPU-redirection machinery I don't have full context on. Also added a small, purely diagnostic change to VulkanVideoPresenter's queue-visibility fence wait (bounded retry loop with a log line instead of vkWaitForFences(..., ulong.MaxValue) so a genuinely stuck GPU fence would show up instead of hanging silently) while ruling that out as a cause here, might be useful to keep regardless.

Happy to send patches for fix 1 and 2 if useful, or just leave this here for whoever picks up the semaphore-count fix.

Spooks4576:

Tested `fix/semaphore-race-and-missing-imports` @ `926998da` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-253

ftre120:
Tested this against a legally-owned dump of Dead Cells (PPSA15552) to check whether it resolves #259, and found one missing piece: `sceNgs2RackCreate`/`sceNgs2SystemCreate` here still generate the rack handle via `TryCreateHandle`, which hands out large pointer-shaped values. Dead Cells' audio init doesn't reuse the value `sceNgs2RackCreate` returns — it hardcodes literal `rdi=0` for all 16 `sceNgs2RackGetVoiceHandle` calls right after creating its first (and only) rack:

```
Import#37587 result: -2142633375 (MwmHz8pAdAo) rdi=0x0000000000000000 rsi=0x0000000000000000 ...
Import#37589 result: -2142633375 (MwmHz8pAdAo) rdi=0x0000000000000000 rsi=0x0000000000000001 ...
... (all 16 voice indices, rdi always 0)
```

Since our rack handle is never `0` (it's a heap-ish address like `0x1C1630000`), every one of those 16 lookups fails with `InvalidRackHandle`, the game's per-object voice-handle cache field stays `0`, and a per-frame audio tick (`sceNgs2VoiceControl`/`sceNgs2VoiceGetState`) then spins on handle `0` forever matching exactly what's described in #259, and still reproducible with this PR's implementation applied.

Disassembling the call site (`sceNgs2SystemCreate`'s continuation, walking back to the containing function via the compiler's `int3` padding to get an unambiguous instruction boundary) confirms the game never re-reads a variable after `sceNgs2RackCreate` for these 16 calls — `0` is passed as a literal immediate. That only makes sense if rack handles are meant to be small sequential IDs (0 for the first rack a system creates), the same idiom several other SCE handle pools use, not addresses.

Fix: give rack handles their own per-system sequential counter starting at 0, instead of routing them through `TryCreateHandle`. Only rack handles — System/Voice handles are untouched.

```csharp
private static readonly Dictionary<ulong, uint> NextRackHandleBySystem = new();

// Rack handles are small sequential integers starting at 0 per system, not
// the pointer-shaped handles TryCreateHandle hands out elsewhere: Dead
// Cells hardcodes rack handle 0 for its first (and only) rack rather than
// reusing the value sceNgs2RackCreate returned, so the first rack a system
// creates must actually be handle 0 for that lookup to succeed.
private static ulong AllocateRackHandle(ulong systemHandle)
{
    lock (StateGate)
    {
        var next = NextRackHandleBySystem.TryGetValue(systemHandle, out var value) ? value : 0;
        NextRackHandleBySystem[systemHandle] = next + 1;
        return next;
    }
}
```

Then in both `sceNgs2RackCreateWithAllocator` and `sceNgs2RackCreate`, replace:

```csharp
if (!TryCreateHandle(ctx, type: 2, systemHandle, out var handle) ||
    !ctx.TryWriteUInt64(outHandleAddress, handle))
{
    return SetReturn(ctx, (int)OrbisGen2Result.ORBIS_GEN2_ERROR_MEMORY_FAULT);
}
```

with:

```csharp
var handle = AllocateRackHandle(systemHandle);
if (!ctx.TryWriteUInt64(outHandleAddress, handle))
{
    return SetReturn(ctx, (int)OrbisGen2Result.ORBIS_GEN2_ERROR_MEMORY_FAULT);
}
```

Verified on the real Dead Cells binary: with this change on top of this PR's `scePadGetHandle`/`sceNgs2SystemCreate`/`sceNgs2RackCreate`, the `InvalidRackHandle`/`InvalidVoiceHandle` spam disappears entirely and the game proceeds well past this point into real AGC/shader/Vulkan draw submission (2M+ imports of forward progress vs. previously spinning forever around the same ~2.9M import count). It then hits a separate, unrelated GPU offscreen-queue backpressure condition, but that's clearly a different subsystem from this PR's scope.

kostyaff:
Follow-up review after the latest commits (rebase + review feedback address).

## Partially addressed ✅

**Item 1 (semaphore count corruption):** The `Count -= needCount` is now guarded by `if (semaphore.Count >= needCount)` in the late-arrival path, and the `Math.Max(Count, needCount)` fabrication appears gone. The pending-exception path now decrements `WaitingThreads` and returns OK without mutating `Count` — better than before, though it still returns `ORBIS_GEN2_ERROR_OK` to the caller as if the acquire succeeded (the exception is delivered separately). Worth confirming this is the intended semantics vs. returning a distinct status, but the token-balance drift is fixed.

**Item 7 (`_pendingException`):** Now `volatile bool` — good. `BlockedSemaphoreGate` / `ExternalGuestThreadHandle` are still plain mutable statics, but the single-outstanding-wait assumption is documented by the structure. Acceptable.

## Still open ⚠️

**Item 4 (ungated hot-path logging) — not fixed.** Two `Console.Error.WriteLine` calls in the semaphore wait path are unconditional:
- Line ~823: `[MAIN_THREAD] Blocking on semaphore handle=...` (before `WaitSemaphoreOnHostThread`)
- Line ~915: `[MAIN_THREAD] Unblocked from semaphore handle=...` (after wake)

`HostMainThread.SetBlocked` correctly gates its own logging behind `_logBlocking`, but these two direct `Console.Error.WriteLine` calls in `KernelSemaphoreCompatExports` are not gated. Since the wait path is one of the hottest in the emulator, please gate these behind `HostMainThread._logBlocking` (or a similar env flag) the same way.

**Item 5 (`GuestAddressSpaceReservation` race):** The `ReleaseReservation() → loader-allocation` window is still a TOCTOU race — the GC can grab the range between the free and the loader's reserve. The comment now acknowledges timing but not the race itself. Not a blocker if the reservation is best-effort, but worth a one-line comment noting the residual race and that `VirtualAlloc2` placeholder replacement would make it atomic on Win10+.

**Item 6 (macOS → !Windows):** Still present — Linux silently switches from exact-address search to over-allocation. If this is intentional for Rosetta-on-Linux/Docker, a comment saying so would help; otherwise scope to macOS + env override.

## Gameplay regression
Spooks' bot reports Dreaming Sarah ❌ failed on this branch. The failure reason isn't linked here — worth checking whether it's related to the ungated logging (unlikely) or the semaphore exception-path semantics (more likely). Dead Cells ✅ passed.

Items 4 and 5 are quick fixes; item 6 is a one-line scope change. With those addressed I'd be comfortable approving.

j92580498-max:
https://docs.amd.com/v/u/en-US/rdna2-shader-instruction-set-architecture

j92580498-max:
https://docs.discoverer.bg/zen2.html

jackby03:
Closing in favor of #? (new PR: \ix/gc-deadlock-and-missing-imports-v2\). Rebased onto latest main, consolidated to 1 commit, and dropped changes already upstream (Ngs2 rack handles, SSE4a recovery, space-in-path fix). Same fixes: GC deadlock, semaphore race, POSIX ABI wrappers, HLE imports.

PR #264, opened 2026-07-16
[Perf] Behavior-preserving hot-path allocation/LINQ wins

galvesribeiro:
Follow-up to the performance sweep. Each commit is one item, behavior-preserving, verified with a full Release build + the xunit suites (83 + 33 green) after each. These are the findings that are (a) still present on current `main` and (b) safe to land without game testing.

## Commits

1. **Gate event-flag tracing** — `TraceEventFlag` built its interpolated string (and, on the wait path, `FormatFrameChain` + a `new StringBuilder(256)` in `FormatGuestWaitObject` + ~12 guest reads) on every `Set`/`Clear`/`Poll`/`Wait`, then checked the env var *inside* the method. Now gated behind a cached `_traceEventFlag`.
2. **Hoist `IsNoBlockLeaf`** — the leaf-dispatch path called `IsNoBlockLeafImport(nid)` (a ~30-literal string match) per import; joins the already-precomputed `IsLeaf`/`NidHash` on `ImportStubEntry`.
3. **Cache trace env-var flags** — `SHARPEMU_LOG_FIBER` (per fiber transfer) and `SHARPEMU_LOG_DIRECT_MEMORY` (~8 sites) were read via `GetEnvironmentVariable` per op; cached once.
4. **Idle-pump snapshot** — `PumpUntilGuestThreadsIdle` allocated a `GuestThreadState[]` (LINQ `Values.ToArray()`) every spin to tally three booleans; now tallied under the lock allocation-free, snapshot only for the gated diagnostic. `SnapshotGuestThreads` uses `CopyTo`.
5. **De-LINQ `GetPixelColorExportMask`** — a `Select/OfType/Where/Aggregate` scan over all instructions, called per render target (twice per draw); manual scan, identical result.
6. **De-LINQ per-draw render-target selection** (AgcExports) — `Where/OrderBy/ToArray` + fallback `Where/ToArray` → manual filter + `List.Sort` (slots distinct ⇒ matches the stable `OrderBy`).
7. **De-LINQ per-draw render-target validation** (Vulkan presenter) — `Any` ×2 + `Select(...).Distinct().Count()` (per-draw HashSet) + `Enumerable.Repeat(...).ToArray()` → manual scans + `Array.Fill`.

## Already implemented on `main` (verified — nothing to do)

The headline dispatch item is already done: the hot path reads the precomputed `ImportStubEntry.IsLeaf`/`entry.NidHash`; the ~92-literal `IsLeafImport` cascade runs only at stub setup (cold). `CpuContext.TryReadNullTerminatedUtf8` is already chunked/pooled. Semaphore/event-queue/pthread trace paths are already cached+gated. The AGC↔presenter transfer already uses a bounded `GuestDataPool`.

## Left for a follow-up, game-tested PR (not pushed blind)

These change runtime behavior in ways only running real games validates (this project's verification method), so they are intentionally not in this PR:
- Pool guest-thread host threads instead of `new Thread` per wake
- Lock-free guest-memory region reads (replace the per-read `ReaderWriterLockSlim`)
- Index blocked threads by wake key + a deadline heap (replace the O(n) wake scan)
- Persistent/reset descriptor pool instead of create+destroy per draw
- Earlier shader cache that short-circuits translation/evaluation per draw
- Decoded-texture/index-buffer caching to skip per-draw guest-memory copies
- Per-event-queue locks instead of the single global gate
- Pipeline/descriptor cache key → value struct, dropping `Submit*` defensive `.ToArray()` copies (ownership + key-equality subtleties), and the `sceKernelVirtualQuery` sorted-region rework — behavior-preserving in principle but subtle enough to want game validation

## Verification
Release build clean (zero new warnings); `SharpEmu.Libs.Tests` + `SharpEmu.SourceGenerators.Tests` green after each commit.


kostyaff (approved):
## Review

6 files, +192/−78. Behavior-preserving perf optimisations: eliminates LINQ iterator/closure allocations on hot paths, caches env-var reads, replaces O(n) memory region scan with binary search.

### Changes reviewed

1. **`_logFiber` cached field** — `GetEnvironmentVariable` hoisted from per-call to init time. ✅
2. **`IsNoBlockLeaf` on `ImportStubEntry`** — pre-computed at stub setup. ✅
3. **`GetGuestThreadActivity` allocation-free tally** — idle spin loop no longer allocates `GuestThreadState[]` per iteration. `SnapshotGuestThreads()` only for gated diagnostic dump. ✅
4. **`SnapshotGuestThreads` uses `CopyTo` instead of `ToArray()`** ✅
5. **AGC render targets: manual `foreach` + `Sort` instead of `Where().OrderBy().ToArray()`** — at most 8 targets. ✅
6. **`GetPixelColorExportMask`: manual `foreach` instead of `Select().OfType().Where().Aggregate()`** — same result. ✅
7. **EventFlag trace: `_traceEventFlag` cached once** — P/Invoke eliminated from every call site. ✅
8. **`_mappedRegions`: `Dictionary` → `SortedList`** — enables binary search. Values enumerate in ascending address order. ✅
9. **Binary search in `TryFindVirtualQueryRegionLocked`** — O(n) → O(log n). Overflow-safe midpoint: `(int)(((uint)lo + (uint)hi) >> 1)`. Floor index + findNext logic correct. ✅
10. **`_traceDirectMemory` cached once** ✅
11. **Vulkan `AnyRenderTargetInvalid` / `RenderTargetsMismatchedOrAliased`** — manual loops instead of `Any()` / `Select().Distinct().Count()`. ✅
12. **`Enumerable.Repeat().ToArray()` → `Array.Fill`** ✅

## Concerns

`SortedList<ulong, MappedRegion>.Add()` is O(n) for insertion. If there are many mapping operations at runtime, this could be slower than `Dictionary`. However, `VirtualQuery` (binary search) is called far more frequently than `mmap` (load-time). The tradeoff is correct for an emulator.

## Verdict: ✅ Approve

All changes are behavior-preserving with clear comments explaining each equivalence. The `SortedList` + binary search is the most impactful — O(n) → O(log n) on every `VirtualQuery` call.

PR #265, opened 2026-07-16
[Vulkan] Validate and isolate persistent pipeline caches

Dafenxz0:
## What changed

This makes the persistent Vulkan pipeline cache device-aware instead of treating one blob as valid for every Vulkan implementation:

- default cache files are keyed by vendor ID, device ID, and `pipelineCacheUUID`
- the 32-byte `VkPipelineCacheHeaderVersionOne` header is checked before any blob reaches `vkCreatePipelineCache`
- files smaller than the header or larger than the existing 256 MiB export limit are rejected before allocation
- an explicitly configured `SHARPEMU_VK_PIPELINE_CACHE_PATH` still works, but its contents now receive the same compatibility check

## Why

#216 made pipeline-cache persistence useful, especially for expensive MoltenVK shader compilation. The default path is currently shared by all GPUs and drivers, though. On a hybrid system, switching between the integrated and discrete GPU can therefore turn the cache into a ping-pong: each implementation rejects and later overwrites the other implementation's blob.

Vulkan already exposes the exact identity required for this through `pipelineCacheUUID`, and places that UUID together with the vendor and device IDs in the cache header. Using both sides of that contract keeps warm caches independent and avoids passing truncated or unrelated files into Vulkan.

This does not change pipeline keys, shader compilation, rendering state, synchronization, or when snapshots are saved.

## User impact

- cached pipeline work survives switching between Vulkan devices or driver implementations
- corrupt/truncated cache files fall back to a clean cache without relying on driver behavior
- an unexpectedly large cache file cannot cause an equally large managed allocation during startup

The previous unqualified `vulkan-pipeline-cache.bin` is intentionally not migrated because it cannot be attributed to a device without validating it first. The first launch after this change may rebuild that one old cache; subsequent launches use the device-specific file.

## Validation

- Release solution build: 0 warnings, 0 errors
- `SharpEmu.Libs.Tests`: 86/86 passed
- `SharpEmu.SourceGenerators.Tests`: 33/33 passed
- focused pipeline-cache tests: 3/3 passed
- `git diff --check`

The focused tests cover a compatible cache, a cache from a different driver UUID, and a sparse file above the size limit.

Reference: [VkPipelineCacheHeaderVersionOne](https://registry.khronos.org/vulkan/specs/latest/man/html/VkPipelineCacheHeaderVersionOne.html)

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VulkanPipelineCacheStorage.cs
+++ b/src/SharpEmu.Libs/VideoOut/VulkanPipelineCacheStorage.cs
@@ -1,0 +1,64 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+using System.Buffers.Binary;
+
+namespace SharpEmu.Libs.VideoOut;
+
+internal static class VulkanPipelineCacheStorage
+{
+    internal const int UuidSize = 16;
+    internal const long MaximumCacheSize = 256L * 1024L * 1024L;
+
+    private const int HeaderSize = 32;
+    private const uint HeaderVersionOne = 1;
+
+    internal static string GetFileName(
+        uint vendorId,
+        uint deviceId,
+        ReadOnlySpan<byte> pipelineCacheUuid)
+    {
+        ArgumentOutOfRangeException.ThrowIfNotEqual(pipelineCacheUuid.Length, UuidSize);
+        return $"vulkan-{vendorId:X8}-{deviceId:X8}-{Convert.ToHexString(pipelineCacheUuid)}.bin";
+    }
+
+    internal static bool TryReadCompatible(
+        string path,
+        uint vendorId,
+        uint deviceId,
+        ReadOnlySpan<byte> pipelineCacheUuid,
+        out byte[] data)
+    {
+        data = [];
+        using var stream = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read);
+        if (stream.Length is < HeaderSize or > MaximumCacheSize)
+        {
+            return false;
+        }
+
+        data = new byte[checked((int)stream.Length)];
+        stream.ReadExactly(data);
+        if (HasCompatibleHeader(data, vendorId, deviceId, pipelineCacheUuid))
+        {
+            return true;
+        }
+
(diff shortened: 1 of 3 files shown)


kostyaff (approved):
## Review

3 files (1 new), +202/−5. Validates Vulkan pipeline cache header before loading. Device-keyed file naming prevents cross-device cache contamination.

### `VulkanPipelineCacheStorage.cs` (new, 64 lines) — ✅

Clean separation. Header validation: size (32-byte min, 256MB max), version (must be 1), vendor ID, device ID, pipeline cache UUID. `BinaryPrimitives.ReadUInt32LittleEndian` for all fields. `ArgumentOutOfRangeException.ThrowIfNotEqual` for UUID length. `checked()` cast for stream length. `FileShare.Read` on FileStream — correct.

### `VulkanVideoPresenter.cs` integration — ✅

`CreatePipelineCache` calls `GetPhysicalDeviceProperties` once, passes vendor/device/UUID to `GetPipelineCachePath`. `TryReadCompatible` replaces raw `File.ReadAllBytes` — incompatible cache logged and rebuilt. `SavePipelineCache` uses shared `MaximumCacheSize` constant.

### `GetPipelineCachePath` — ✅

File name now includes vendor ID, device ID, UUID: `vulkan-{vendor:X8}-{device:X8}-{uuid}.bin`. Prevents loading cache from different GPU/driver.

### Tests (110 lines) — ✅

Three tests: compatible cache loaded, mismatched UUID rejected, oversized cache rejected before reading.

## Non-blocking notes

- Old `vulkan-pipeline-cache.bin` orphaned after update. Not a problem — new cache builds fresh.
- `ReadExactly` reads entire file into memory (up to 256MB max). Fine for pipeline cache.

## Verdict: ✅ Approve

Well-structured, well-tested, correct validation logic.

Spooks4576:

Tested `agent/harden-vulkan-pipeline-cache` @ `70a1654a` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-265

PR #268, opened 2026-07-16
[HLE/Net] Fix sceNetHtonl/Htons/Ntohl/Ntohs returning 0

abdulazizFarhan:
## What This PR Does

Fixes a silent correctness bug in four PS5 network exports where the byte-swapped result was being overwritten with `0` before being returned to the guest.

## Why

`sceNetHtonl`, `sceNetHtons`, `sceNetNtohl`, and `sceNetNtohs` correctly compute the byte swap into `Rax` but finish with `return ctx.SetReturn(0);`. `SetReturn` writes its argument into `Rax`, overwriting the swapped value with `0`. Every one of the four exports returned `0` to the guest regardless of input.

This silently corrupts anything that byte-swaps ports or addresses — for example filling in a `sockaddr_in` structure — even though the calling code looks correct.

## Change

`src/SharpEmu.Libs/Network/NetExports.cs` — for each of the four exports, replace the last line `return ctx.SetReturn(0);` with `return (int)OrbisGen2Result.ORBIS_GEN2_OK;`. This is the same pattern already used 8 times in the same file (e.g. `sceNetPoolCreate` at lines 318-319, `sceNetSocket` at lines 108-109, `sceNetAccept` at lines 264-265).

Diff: 4 lines changed across one file.

## Tests

New `tests/SharpEmu.Libs.Tests/Network/NetExportsTests.cs` (REUSE SPDX header included) with 10 xUnit `[Fact]`s modeled on the existing `KernelMemoryCompatExportsTests` pattern (FakeCpuMemory + CpuContext):

1. `Htonl_SwapsBytesAndPreservesRax` — `0x12345678 -> 0x78563412`
2. `Htons_SwapsBytesAndPreservesRax` — `0x1234 -> 0x3412`
3. `Ntohl_RoundTripsHtonl` — `Htonl(Ntohl(x)) == x`
4. `Ntohs_RoundTripsHtons` — same for 16-bit
5. `Htonl_ZeroInputReturnsZero` — guards the dispatch status on the zero path
6. `Htonl_NonZeroInputNeverReturnsZero` — fuzz 1000 inputs, regression guard
7. `Htons_NonZeroInputNeverReturnsZero` — fuzz 1000 inputs, regression guard
8. `Htons_OnlyTouchesLowTwoBytes` — verifies the ushort cast
9. `Ntohl_NonZeroInputNeverReturnsZero` — fuzz 1000 inputs, regression guard
10. `Ntohs_NonZeroInputNeverReturnsZero` — fuzz 1000 inputs, regression guard

The four `NonZeroInputNeverReturnsZero` tests are the regression guards that would have failed before the fix.

## Verification

```
$ dotnet build SharpEmu.slnx -c Release --no-restore
Build succeeded. 0 Warning(s). 0 Error(s).

$ dotnet test tests/SharpEmu.Libs.Tests/SharpEmu.Libs.Tests.csproj -c Release --no-build --filter "FullyQualifiedName~NetExports"
Passed!  - Failed:     0, Passed:    10, Skipped:     0, Total:    10

$ dotnet test SharpEmu.slnx -c Release --no-build
Passed: 95 / Failed: 0 (full test suite, no regressions)
```

## Risk

Minimal. Behavior change is local to four leaf-level exports that previously returned a constant `0`. Callers that depended on the (broken) `0` return value would have been silently broken already. The new behavior matches the documented semantics of `htonl`/`htons`/`ntohl`/`ntohs` on every Unix system.

## Reference

Closes #211.

### DIFF
--- a/src/SharpEmu.Libs/Network/NetExports.cs
+++ b/src/SharpEmu.Libs/Network/NetExports.cs
@@ -345,7 +345,7 @@ public static class NetExports
     {
         var value = unchecked((uint)ctx[CpuRegister.Rdi]);
         ctx[CpuRegister.Rax] = BinaryPrimitives.ReverseEndianness(value);
-        return ctx.SetReturn(0);
+        return (int)OrbisGen2Result.ORBIS_GEN2_OK;
     }
 
     [SysAbiExport(
@@ -357,7 +357,7 @@ public static class NetExports
     {
         var value = unchecked((ushort)ctx[CpuRegister.Rdi]);
         ctx[CpuRegister.Rax] = BinaryPrimitives.ReverseEndianness(value);
-        return ctx.SetReturn(0);
+        return (int)OrbisGen2Result.ORBIS_GEN2_OK;
     }
 
     [SysAbiExport(
@@ -369,7 +369,7 @@ public static class NetExports
     {
         var value = unchecked((uint)ctx[CpuRegister.Rdi]);
         ctx[CpuRegister.Rax] = BinaryPrimitives.ReverseEndianness(value);
-        return ctx.SetReturn(0);
+        return (int)OrbisGen2Result.ORBIS_GEN2_OK;
     }
 
     [SysAbiExport(
@@ -381,7 +381,7 @@ public static class NetExports
     {
         var value = unchecked((ushort)ctx[CpuRegister.Rdi]);
         ctx[CpuRegister.Rax] = BinaryPrimitives.ReverseEndianness(value);
-        return ctx.SetReturn(0);
+        return (int)OrbisGen2Result.ORBIS_GEN2_OK;
     }
 
     [SysAbiExport(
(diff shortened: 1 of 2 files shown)


6can:
this is the same four-line fix as #211, which i opened yesterday and kostyaff
approved this morning. the fuzz tests here are a nice addition, happy to pull
those into #211 if they're wanted.

could you drop the "Closes #211" though? merging this would auto-close a PR
that's already approved and came first.

par274:
Closed because #211 duplicate.

abdulazizFarhan:
Hi canbedir — thanks for the heads-up. I missed your PR #211 when I scanned the issues; my duplicate check only looked at /issues?state=open and did not catch the PR conflict. Closing #268 in favor of yours is the right call.

The fuzz tests in #268 are yours to use however you want — REUSE/SPDX compliant, no new dependencies, pull from a [Fact] pattern that should slot in next to whatever test setup you already have for #211. If you would rather I open a separate PR that just adds the fuzz tests against the post-fix state of #211, say the word and I will do that — but only if it is net-positive for you; otherwise feel free to cherry-pick.

I have added a mandatory duplicate-check step to my dev-loop workflow so this does not happen again (gh pr list plus gh search prs, not just gh issue list). Lesson learned.

Issue #257, opened 2026-07-16
Unimplemented: sceVideoOutIsOutputSupported (Quake/Kex Engine crash)


jackby03:
## Symptoms
- Quake PPSA01880 (Kex Engine) crashes at startup with Access Violation

## Log evidence

```
[LOADER][WARN] Import#115078 unresolved: nid=Nv8c-Kb+DUM ret=0x000000080032B3A1 rdi=0x0000000000000001 rsi=0x000000000000000F rdx=0x0000000000000000 rcx=0x0000000000000000 r8=0x0000000000000000 r9=0x00000000000001FC
```

NID identified in aerolib.bin: Nv8c-Kb+DUM -> sceVideoOutIsOutputSupported

## Context
Quake also has AVPlayer errors (cannot open intro .mp4 videos). This function is likely needed by 3D engines that check video output capabilities before initializing rendering.

PR #269, opened 2026-07-16
VideoOut: implement Gen5 output support query

chengboonrong:
## What

- replace the generic dual-generation stub from #270 with the Gen5 `sceVideoOutIsOutputSupported` ABI
- validate its handle, options, and reserved arguments
- report default mode as supported and 119.88 Hz according to the represented port refresh rate

## Why

#270 resolves `Nv8c-Kb+DUM`, but public Gen5 callers use a five-argument handle, mode, options, and reserved-data ABI. The merged stub interprets the handle as a bus type and can report unsupported configurations as supported. Quake queries mode `15`, while SharpEmu currently represents a 60 Hz output.

## Verification

- focused `VideoOutOutputSupportTests` dispatch test
- 179 SharpEmu.Libs tests passed
- 33 SharpEmu.SourceGenerators tests passed
- `git diff --check`

## Limitations

No Quake PPSA01880 dump was available locally, so title-level startup validation remains for the issue reporter. This does not address the separate AVPlayer failures mentioned in #257.

Related to #257 and follows #270.

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VideoOutExports.cs
+++ b/src/SharpEmu.Libs/VideoOut/VideoOutExports.cs
@@ -23,6 +23,7 @@ public static class VideoOutExports
     private const int OrbisVideoOutErrorInvalidHandle = unchecked((int)0x8029000B);
     private const int OrbisVideoOutErrorInvalidEventQueue = unchecked((int)0x8029000C);
     private const int OrbisVideoOutErrorInvalidEvent = unchecked((int)0x8029000D);
+    private const int OrbisVideoOutErrorUnsupportedOutputMode = unchecked((int)0x80290016);
     private const int OrbisVideoOutErrorInvalidOption = unchecked((int)0x8029001A);
     private const int SceVideoOutBusTypeMain = 0;
     private const int SceVideoOutBufferAttributeOptionNone = 0;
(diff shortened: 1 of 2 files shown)


par274:
The Gen5-specific approach in this PR looks more correct than the currently merged stub, especially the handle, mode, options, and reserved-argument validation.

Since #270  is already merged, please rebase onto the latest main and replace the existing implementation rather than adding a duplicate export for the same NID. And remove the .md files.

chengboonrong:
@par274 Updated PR #269 based on your feedback. I rebased onto the latest main, replaced the implementation from #270 instead of adding a duplicate export, and removed the Markdown files.  
The focused test and full test suites pass, and the Release build completes without warnings or errors. I’m excited to start contributing to SharpEmu—please let me know if there’s anything else you’d like adjusted!

anes66 (approved):
Verified the five-argument Gen5 ABI and validation behavior against the corresponding Kyty implementation: handle validation, the 0x40-byte zeroed options structure, reserved arguments, supported modes, and the 119.88 Hz capability result all match. The focused test also dispatches through the generated registry and closes the opened port in `finally`, so it does not leak shared VideoOut state. All repository checks pass on this head.

Spooks4576:

Tested `codex/issue-257-output-support` @ `23d5ff84` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-269

PR #284, opened 2026-07-16
[Logging] Collapse repeated log lines centrally over Console

shadowbeat070:
### What this changes

Adds a central log-collapsing layer so repeated console output folds to a single line plus a summary, instead of each spammy call site needing its own throttling. Three things:

- New src/SharpEmu.Logging/CollapsingTextWriter.cs — a TextWriter installed over Console.Out/Console.Error. It assembles complete lines and compares them by a normalised key (hex addresses 0x… and #counters masked, via source-generated regex), so structurally identical spam that differs only by a pointer or an incrementing id still folds. The first occurrence of a run is printed verbatim; repeats are counted and a (previous message repeated N more times) summary is emitted when a different line arrives, on an idle timer, or at shutdown via Drain(). Flush() deliberately stays silent so Console.Error's auto-flush doesn't print a summary after every repeat.
- src/SharpEmu.CLI/Program.cs — InstallConsoleCollapsing() wraps the console after the file-mirror setup (outermost, so the log file is collapsed too); DrainConsoleCollapsing() runs in Main's finally so a trailing run isn't lost.
- tests/SharpEmu.Libs.Tests/Logging/CollapsingTextWriterTests.cs — 8 tests.

Purely additive: 3 files, +486 lines, no existing behaviour paths modified.

### Why it's needed / the problem it solves

Two real sources flood the log. A chatty guest can printf the same line every frame, and the HLE import tracer emits one [LOADER][WARN] per call — e.g. on shutdown sceAudioOutOutput returns INVALID_ARGUMENT once per audio-thread iteration after the port is closed, producing hundreds of near-identical lines. Those import lines are never byte-identical (each carries an incrementing Import#N and an rsi=0x… address that alternates between two buffers), so a plain "fold identical consecutive lines" approach can't catch them.

About 90% of the emulator's output goes straight to Console rather than through the logger, so a collapser at the log-sink level would miss nearly everything. The one place that can fold all of it is the Console stream itself — hence a TextWriter decorator, with token normalisation so the varying counter/address lines collapse too.

### How I verified it

- New unit tests: 8/8 pass, including the real import-line format collapsing to one line + summary, genuinely-different NIDs staying apart, the Flush()-emits-nothing guarantee, the idle-timer summary (driven deterministically via a manual TimeProvider), and drain-at-shutdown.
- Build is clean with dotnet build SharpEmu.slnx -c Release (0 warnings, 0 errors) using the SDK pinned in global.json (10.0.103).
- Full suite with dotnet test SharpEmu.slnx -c Release: SharpEmu.SourceGenerators.Tests 33/33 pass; SharpEmu.Libs.Tests 92/93 pass.
  - The single failure — KernelMemoryCompatExportsTests.Sprintf_ReadsVariadicDoubleFromXmmRegister (Expected "0.5576", Actual "0,5576") — is a pre-existing locale bug on main, not introduced by this PR. It reproduces on a host whose OS locale uses a comma decimal separator (I'm on es-ES Windows) and lives in KernelMemoryCompatExports, a file this PR does not touch. It is unrelated to this change.

### Scope / follow-up

I kept this PR to the central collapser only. One deliberate trade-off: because comparison is on the normalised form, two genuinely different lines that differ only by a masked token (an address or counter) will collapse together — the first is always shown verbatim, so one real sample is preserved, but per-address detail within a repeated run is lost. If a subsystem ever needs every address kept, that call site can bypass the writer.

Separately, while verifying I confirmed the pre-existing es-ES Sprintf locale failure noted above. That's a distinct conformance fix (CultureInfo.InvariantCulture in the printf numeric paths) and I'd rather address it in its own PR than mix it in here.

### Note on AI assistance

This change was developed with AI assistance (per the project's AI-assisted contributions policy). I understand the change fully and can explain, modify, and maintain it. Happy to answer any review questions.

### DIFF
--- a/src/SharpEmu.CLI/Program.cs
+++ b/src/SharpEmu.CLI/Program.cs
@@ -56,6 +56,7 @@ internal static partial class Program
         }
         finally
         {
+            DrainConsoleCollapsing();
             DropConsoleFileMirror();
             SharpEmuLog.Shutdown();
         }
@@ -84,6 +85,12 @@ internal static partial class Program
         {
             TryEnableConsoleFileMirror(earlyLogFilePath);
         }
+
+        // Funnel all console output through one collapser so repeated log lines
+        // (guest printf floods, per-call import-result warnings) fold to a single
+        // line plus a summary instead of drowning the log. Installed last, over
+        // any file mirror, so the log file is collapsed too.
+        InstallConsoleCollapsing();
 
         if (!CheckHostArchitecture())
         {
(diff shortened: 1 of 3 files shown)


Spooks4576:

Tested `feat/central-log-collapser` @ `0fac6d4c` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-284

PR #286, opened 2026-07-16
fix: Unity IL2CPP GC deadlock + ClearEventFlag bitmask correction

jackby03:
## Summary

Fixes the Unity IL2CPP garbage collection deadlock (Arcade Game Zone PPSA19015) and a bug in \sceKernelClearEventFlag\.

### GC Deadlock Fix (cherry-pick from #253)

The GC coordinator raises \sceKernelRaiseException(type=0x1E)\ on the main thread while it's blocked in a host semaphore wait. The exception is queued but never delivered since the thread never reaches an HLE boundary.

- Track pending exception state and the semaphore gate the main thread is blocked on in \HostMainThread\
- When an exception targets the main thread, pulse the gate and set a flag so the semaphore wait returns early, allowing the import dispatch's safe-point check to deliver the exception handler
- Track the main thread's managed thread ID via \SetManagedThreadId\

### ClearEventFlag Fix

KernelClearEventFlag was using bits &= pattern instead of bits &= ~pattern. On real PS5, sceKernelClearEventFlag(handle, 0) is a no-op — it clears no bits. With the bug, pattern=0 cleared all bits, incorrectly resetting the resumeEvent flag used by IL2CPP GC coordination.

### Results

Arcade Game Zone (PPSA19015): **81k → 161k imports**, 4 GC stop-the-world cycles complete. Previously deadlocked on the first cycle.

### Notes

- Cherry-picked GC deadlock commit will be redundant once #253 merges
- Related to #254

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
@@ -588,6 +588,25 @@ public sealed partial class DirectExecutionBackend
 			cpuContext[CpuRegister.R15] = value8;
 			cpuContext[CpuRegister.Rdi] = value;
 			cpuContext[CpuRegister.Rsi] = value2;
+			
+			// Track main thread activity for diagnostics
+			if (Environment.CurrentManagedThreadId == HostMainThread.GetManagedThreadId() && !GuestThreadExecution.IsGuestThread)
+			{
+				HostMainThread.SetLastRip(num7); // return address
+				HostMainThread.SetLastImportNid(importStubEntry.Nid);
+				
+				// Log synchronization operations from main thread
+				if (importStubEntry.Nid == "Zxa0VhQVTsk" || // sceKernelWaitSema
+				    importStubEntry.Nid == "4czppHBiriw" || // sceKernelSignalSema
+				    importStubEntry.Nid == "188x57JYp0g")   // sceKernelCreateSema
+				{
+					Console.Error.WriteLine(
+						$"[MAIN_THREAD] Import#{num} {importStubEntry.Nid} " +
+						$"ret=0x{num7:X16} rdi=0x{value:X16} rsi=0x{value2:X16} rdx=0x{num3:X16} " +
+						$"result={orbisGen2Result}");
+				}
+			}
+			
 			if (GuestThreadExecution.TryConsumeCurrentContextTransfer(out var transferTarget))
 			{
 				if (!TryPrepareGuestContextTransfer(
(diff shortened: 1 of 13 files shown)


jackby03:
Added one more fix: PFS file descriptors (fd >= 0x80000000) were failing in fstat and close because they're not tracked in _openFiles. Now fstat returns a default stat and close succeeds for unknown fds instead of returning NOT_FOUND.

Result with Arcade Game Zone (PPSA19015): The game now boots, renders frames at 1920x1080 via Vulkan, and runs continuously with working GC stop-the-world cycles. Previously deadlocked at 81k imports — now runs for minutes without crashing.

anes66:
Compatibility-tested head `eb59faf` with **Bud Spencer & Terence Hill - Slaps And Beans 2 (`PPSA15929`)** on Windows 11 (Ryzen 7 9800X3D, RTX 3060, 32 GB RAM).

- PR test suite: 186 passed (153 library + 33 source-generator)
- upstream `2db1fae`: reaches roughly import 8.3k, then reproduces the Unity worker / `pthread_cond_wait` stall path
- this PR: two bounded runs reached imports 81,454 and 81,452 respectively
- `AcslpN1jHR8` no longer reaches the unresolved-import path
- no native access violation and no stall watchdog snapshot during the 10s/15s windows
- no VideoOut milestone yet; both runs were stopped by the external safety timeout with no remaining process

This independently confirms a substantial Unity GC/semaphore progression improvement on a second title. The last observed state was the newly scheduled guest worker waiting on `SuspendSemaphore`, so this is not a full boot confirmation.

xnetcat (changes requested):
Thanks for the investigation — the GC-deadlock delivery mechanism is attacking a real problem (pending exceptions never delivered to a thread blocked in a host wait). Requesting changes on the event-flag semantics and a few silent-success stubs before this merges:

**1. `KernelClearEventFlag`: `bits &= ~pattern` contradicts both reference emulators — needs hardware evidence**
Both mature references implement the bitPattern argument as the *keep mask*, exactly like the code being changed:
- shadPS4 `src/core/libraries/kernel/threads/event_flag.cpp` — `Clear(u64 bits) { ... m_bits &= bits; }`, called directly from `sceKernelClearEventFlag(ef, bitPattern)`.
- Kyty `src/kernel/eventflag.cpp:355` — `KernelClearEventFlag(ef, bit_pattern)` → `ef->Clear(bit_pattern)` with the same AND-with-pattern semantics.

Those two implementations have been validated against very large game libraries, so flipping to `&= ~pattern` needs stronger evidence than one title improving — especially since this PR *also* changes GC exception delivery, which touches the same IL2CPP suspend/resume flags; the AGZ improvement may come from that half. Note `sceKernelClearEventFlag(ef, 0)` being a no-op is exactly what you'd expect a game to rely on **either way** (with keep-mask semantics games pass `~bitsToClear`, and IL2CPP passing literal `0` would mean "clear everything" intentionally). Suggest: split this into its own PR with either (a) a homebrew test on real hardware, or (b) a trace showing the exact call sequence where keep-mask semantics break the resumeEvent, plus a unit test pinning whichever semantics wins.

**2. `sceSysmoduleGetModuleInfoForUnwind` stub returns success without writing the out-struct**
The (VAddr, flags, info*) ABI expects a ~0x130-byte info struct written on success; this stub returns 0/OK and leaves the caller's stack buffer uninitialized. libunwind then consumes garbage module bounds during C++ exception handling — strictly worse than an honest error. Either delegate to the real `sceKernelGetModuleInfoForUnwind` (same rdi/rsi/rdx layout) or return `SCE_KERNEL_ERROR_ESRCH`-style failure.

**3. Unknown-fd fallbacks convert errors into silent corruption**
- `read(_)` on an unknown fd now returns 0/OK — i.e. silent EOF. Any future fd-mapping bug becomes "file mysteriously empty" instead of a visible error.
- `fstat` fabricates a size-0 regular file with current timestamps.
- `getdirentries` returns 0 entries/OK.
If the goal is PFS handles, track PFS fds explicitly (a set populated at PFS open time) and scope the fallbacks to those, keeping honest errors for genuinely unknown fds — otherwise these will mask real bugs and produce very confusing downstream failures.

**4. `sceAppContentDownloadDataGetAvailableSpaceKb` returns OK without writing the out param**
The caller reads an uninitialized sizeKb. Write a deterministic value (e.g. a large fixed free-space figure) before returning OK.

**5. GC delivery mechanism**
Same feedback as on #253 (the canonical copy): the pending-exception early-out must not fabricate semaphore tokens or decrement `Count` unchecked, and the pending flag needs clearing on *any* delivery path, not only the semaphore one. Since this PR's copy is a cherry-pick that becomes redundant when #253 lands, consider rebasing this PR to contain only the event-flag change + the independent stubs so the two don't drift.

Minor: `KernelExports.cs` loses its trailing newline.

`sceAgcDriverSetTFRing`/`SetHsOffchipParam` OK-stubs and the Ime/AudioOut additions (which do write their out-structs) look fine.

Spooks4576:

Tested `fix/unity-scheduler-suspension` @ `641a6248` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-286

PR #288, opened 2026-07-16
[Perf] Cache pixel shader export masks

999sian:
## What

Cache the eight four-bit pixel color export masks on each immutable decoded `Gen5ShaderProgram`.

`AgcExports` now reads that packed `uint` once per draw and uses constant-time nibble lookups for:

- render-target filtering;
- shader trace output;
- blend write masks.

Repeated exports to one target still combine with bitwise OR, non-color targets are ignored, and the existing slot-0 fallback is unchanged. A focused xUnit test covers all eight slots, repeated exports, and an out-of-range target.

## Why

The previous helper scanned the pixel shader’s complete decoded instruction list for each render target while filtering and then scanned it again while creating blend state. With eight exported targets, that is roughly 16 scans of the same immutable program per draw. Decoded programs are already cached and reused, so the export masks can be computed once when the program is constructed.

## Measurement

Uncommitted .NET microbenchmark:

- synthetic 256-instruction shader;
- 8 color targets;
- 25,000 draw iterations;
- 5 samples, median reported;
- identical checksum (`1,800,000`) for both algorithms.

| Path | Median |
| --- | ---: |
| Existing repeated scans, before the change | 304.974 ms |
| Existing repeated scans, committed comparison run | 310.944 ms |
| Cached packed-mask lookups | 0.246 ms |

The committed comparison run reports a **1,264.51× isolated algorithm speedup**. This is a microbenchmark of the removed CPU work, not a claim of a 1,264× frame-rate improvement.

## Verification

- Red test: `CS1061` before `PixelColorExportMasks` existed
- Focused packed-mask test: 1 passed, 0 failed
- `dotnet test SharpEmu.slnx -c Release --no-build` — 193 passed, 0 failed
- `reuse lint` — REUSE 3.3 compliant, 0 missing licenses

## AI assistance

AI assistance was used for repository research, benchmarking, implementation, and verification. I reviewed the focused change, understand each line, and can explain, modify, debug, and maintain it.

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -5936,13 +5936,14 @@ public static partial class AgcExports
         // Every bound color target the shader exports to. Deferred renderers
         // draw a multi-render-target G-buffer (up to eight slots) in one pass.
         // Fall back to slot 0 if we cannot match any export to a bound target.
+        var pixelColorExportMasks = pixelState.Program.PixelColorExportMasks;
         var allBoundTargets = GetRenderTargets(state.CxRegisters);
         // At most 8 slots; a manual filter avoids the per-draw LINQ iterator/
         // closure allocations. Slots are distinct, so sorting by slot is stable.
         var selectedTargets = new List<RenderTargetDescriptor>(allBoundTargets.Count);
         foreach (var target in allBoundTargets)
         {
-            if (HasPixelColorExport(pixelState, target.Slot))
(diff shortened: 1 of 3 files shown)


par274:
Regression on many games

999sian:
Could you share one or two affected titles and the observable regression (for example black screen, missing targets, or crash), ideally from the f1704dc artifact? The packed-mask result is equivalent to the previous per-target scan for slots 0–7, so I need a failing shader/draw path to isolate the behavioral difference rather than guess at the packing.

PR #293, opened 2026-07-16
Fix GC signal delivery and SSE4a compatibility

flying-dinonugget:
## What changed

- force-wake parked scheduler guest threads for cross-thread signal delivery without losing their blocked continuations
- populate the Orbis/FreeBSD `ucontext_t` passed to GC signal handlers from the interrupted guest register state
- implement `sceKernelIsStack` and the POSIX `pthread_cond_destroy` export
- emulate AMD SSE4a `INSERTQ` and `EXTRQ` on Windows hosts that lack SSE4a
- improve illegal-instruction diagnostics and add focused regression coverage

## Why

Unity's Boehm GC deadlocked when signal 30 targeted a scheduler-owned thread parked in `pthread_cond_wait`. After delivery was enabled, the zeroed signal context caused a post-resume abort. Once GC progressed, direct execution encountered a valid PS4 Jaguar SSE4a `INSERTQ` instruction that raised `#UD` on an Intel host.

These changes preserve nested scheduler waits during signal delivery, provide the register context Boehm requires for conservative scanning, and bridge the PS4/host CPU feature mismatch.

## Impact

The real-game repro advanced from a GC deadlock at roughly 298k imports to booting briefly and reaching roughly 1.29M imports with repeated successful two-thread GC suspend/resume cycles. The deterministic SSE4a trap at `0x8014B9B70` is now emulated.

## Validation

- `dotnet test tests/Hyper5.Libs.Tests/Hyper5.Libs.Tests.csproj --no-restore` — 84 passed
- `dotnet build Hyper5.slnx --no-restore` — succeeded with 0 warnings and 0 errors
- `python scripts/check_sysabi_aerolib.py` — 0 failures


flying-dinonugget:
Closing because this branch was intended to be published only to the fork, not submitted upstream yet.

PR #294, opened 2026-07-16
[VideoOut] Fall back to 8-bit RGBA for unknown pixel formats

### LINKED ISSUE #255: vk.flip_capture_failed: display buffer not registered for Vulkan flip (Dead Cells, Arkanoid)
## Symptoms
- Dead Cells (PPSA15552): vk.flip_capture_failed version=1 queue=dcb.graphics addr=0x102440000 found=False
- Arkanoid Eternal Battle (PPSA06328): Debug.Assert in ExecuteOrderedGuestFlipWait
- Both crash with Access Violation (0xC0000005)

## Log evidence

```
[LOADER][WARN] vk.flip_capture_failed version=1 queue=dcb.graphics addr=0x0000000102440000 found=False initialized=False
```

Arkanoid stack trace:
```
at System.Diagnostics.Debug.Fail(...)
at VulkanVideoPresenter+Presenter.ExecuteOrderedGuestFlipWait(...)
at VulkanVideoPresenter+Presenter.Render(Double)
```
Assertion: An ordered wait-safe marker must execute after its flip capture.

## Analysis
1. Game calls sceVideoOutRegisterBuffers with a pixel format
2. MapPixelFormatToGuestTextureFormat only supports RGB8 (fmt 56) and RGB10 (fmt 9)
3. If the game uses a different format, RegisterKnownDisplayBuffer skips registration
4. When SubmitFlip arrives, the image address isn't in _availableGuestImages or _guestImages
5. ExecuteOrderedGuestFlip fails silently (vk.flip_capture_failed)
6. ExecuteOrderedGuestFlipWait triggers Debug.Assert because capture was expected

mandoof1:
Fixes the display buffer registration for Dead Cells, Arkanoid, and any other
game passing a pixel format not yet in the known list (8-bit sRGB variants → 56,
10-bit packed variants → 9).

**Before:** `MapPixelFormatToGuestTextureFormat` returned 0 for unknown formats.
`RegisterKnownDisplayBuffer` skipped registration. Flips produced `vk.flip_capture_failed`
errors, and `ExecuteOrderedGuestFlipWait` hit a `Debug.Assert`.

**After:** Unknown formats fall back to format 56 (8-bit RGBA) with a logged warning:
```
[LOADER][WARN] vk: unknown pixel format 0x... — falling back to format 56 (8-bit RGBA)
```

This gets games rendering and displaying output (colors will be wrong until
the exact format is identified and mapped). The warning makes it easy for
developers to identify the format and add proper support.

## Related issue
Closes #255

## Validation
- Build: 0 warnings, 0 errors
- New tests: 11/11 passed
- Full suite: 203/203 passed

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VideoOutExports.cs
+++ b/src/SharpEmu.Libs/VideoOut/VideoOutExports.cs
@@ -1646,8 +1646,12 @@ public static class VideoOutExports
     // Maps the PS5 VideoOut pixel format space to the AGC "guest texture format" tags
     // the backend keys its guest-image registry on (see VulkanVideoPresenter.
     // GetGuestTextureFormat: format=10 => 56 for 8-bit RGBA variants, format=9 => 9 for 10-bit).
-    private static uint MapPixelFormatToGuestTextureFormat(ulong pixelFormat) =>
-        NormalizePixelFormat(pixelFormat) switch
+    // Unknown formats default to 56 (8-bit RGBA) with a logged warning so games
+    // display something rather than silently failing the flip pipeline.
+    private static uint MapPixelFormatToGuestTextureFormat(ulong pixelFormat)
+    {
+        var normalized = NormalizePixelFormat(pixelFormat);
+        var result = normalized switch
         {
             SceVideoOutPixelFormatA8R8G8B8Srgb or
             SceVideoOutPixelFormatA8B8G8R8Srgb or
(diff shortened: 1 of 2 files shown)


anes66 (changes requested):
I ran the focused test class on .NET 10.0.103 with build parallelism limited to one worker: all 11 tests pass.

The current change does not yet demonstrate the reported failure is fixed. `vk.flip_capture_failed version=1` can only be emitted after `TrySubmitOrderedGuestImageFlip` accepted the address: that method checks `_availableGuestImages.ContainsKey(address)` before it assigns a non-zero version and enqueues the flip. The failure then occurs in `ExecuteOrderedGuestFlip`, where the corresponding `_guestImages` entry is missing or not initialized. Changing the VideoOut mapping only adds an `_availableGuestImages` entry, so it does not create or initialize the source that the failing capture needs.

There is also no captured numeric format from Dead Cells or Arkanoid in #255, while the new fallback deliberately treats even `0` and arbitrary values such as `0xDEADBEEF` as RGBA8. Without knowing layout/bytes-per-pixel, that may permit an invalid flip rather than safely reject it.

Could you add a regression test at the registration/ordered-flip boundary that reproduces the missing or uninitialized source, and record the actual affected pixel-format value? At minimum, format zero should remain invalid, and #255 should not be closed until the reported capture path is shown to advance.

Issue #116, opened 2026-07-13
Bud Spencer & Terence Hill - Slaps And Beans 2 (Crash error unknown)


SixSeven-off:
### Game Name

Bud Spencer & Terence Hill - Slaps And Beans 2

### Title ID

PPSA15929

### Game Version

01.000.000

### Current Status

Crashes on boot

### Description

This is the error: Process exited with code -1073741819 (unknown).

### Log File

[SharpEmuLog.txt](https://github.com/user-attachments/files/29983845/SharpEmuLog.txt)

### Last Milestone / First Error

\

### Operating System

Windows 11

### CPU

Intel Core i9-13900K

### GPU

NVIDIA GeForce RTX 4080

### RAM

48 GB

### Emulator Version / Commit

50f78f3

### Renderer

Vulkan

anes66:
Retested `PPSA15929` on Windows 11 with upstream commit `2db1fae` (Ryzen 7 9800X3D, RTX 3060, 32 GB RAM).

The title loads all 13 adjacent modules successfully (`failed=0`), rebinds imported data with `unresolved=0`, dispatches the main entry point, and starts the Unity asset-GC worker threads. The first missing HLE call on upstream was:

```
AcslpN1jHR8 = scePadDeviceClassGetExtendedInformation
```

The standalone implementation in #295 was closed to avoid overlapping with #286, whose current head includes the same export together with the Unity GC/semaphore fixes required by this title. Two bounded runs on #286 head `eb59faf` reached roughly 81.4k imports without the earlier native access violation or stall-watchdog snapshot.

There is still no VideoOut milestone, and both runs were stopped by the external safety timeout. This issue should remain open: the active PR represents substantial progression, not a complete boot fix.

PR #295, opened 2026-07-16
[Pad] Add device class extended information export

anes66:
## Summary
- implement `scePadDeviceClassGetExtendedInformation` for Gen4 and Gen5
- return pad-specific argument and handle errors while preserving guest memory faults
- initialize the complete 20-byte standard-controller result
- add focused coverage for valid, compatibility, and error paths

## Root cause
The NID `AcslpN1jHR8` was not registered, so titles querying device-class information reached the unresolved-import path after opening the primary controller.

## Compatibility validation
`PPSA15929` on upstream commit `2db1fae` reproduced the missing import with handle `1`. With this change, that import resolves and the title continues through the same controller initialization path. A later native access violation still remains and is outside the scope of this change.

Related to #116.

## Validation
- focused pad export tests: 6 passed
- full test suite: 198 passed
- REUSE 3.3 lint
- `git diff --check`
- bounded `PPSA15929` compatibility run

### DIFF
--- a/src/SharpEmu.Libs/Pad/PadExports.cs
+++ b/src/SharpEmu.Libs/Pad/PadExports.cs
@@ -10,6 +10,7 @@ namespace SharpEmu.Libs.Pad;
 
 public static class PadExports
 {
+    private const int OrbisPadErrorInvalidArgument = unchecked((int)0x80920001);
     private const int OrbisPadErrorInvalidHandle = unchecked((int)0x80920003);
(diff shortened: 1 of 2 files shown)


anes66:
Closing as superseded by #286, which already carries the same `scePadDeviceClassGetExtendedInformation` export alongside the Unity GC/semaphore fixes needed by the compatibility case that exposed it.

The focused ABI/error-path tests from this branch remain useful reference material, but keeping a second implementation PR open would create unnecessary overlap in `PadExports.cs`.

PR #296, opened 2026-07-16
[Core/GUI] Add managed ZArchive game loading

Aspenini:
Adds read-only support for loading standard `.zar` archives directly without extraction.
Introduces a shared game filesystem for extracted directories and archives, with a pure managed C# ZArchive reader, zstd decompression, random-access streams, block caching, `/app0` integration, and CLI/GUI/library support.


anes66:
I tested the new source normalization against a real extracted-game layout containing both a retail root `eboot.bin` and a runnable `decrypted/eboot.bin`:

```text
<game>/eboot.bin            magic 5414F5EE
<game>/decrypted/eboot.bin  magic 7F454C46
<game>/Media/
<game>/sce_sys/param.json
```

`GameSource.Open` correctly normalizes the decrypted source to the parent content root, so both candidates resolve the same title, title ID, version, and `PhysicalRootPath`. However, `ScanFolders` still enumerates and deduplicates by the original full executable path. The GUI therefore creates two cards for the same title: one launches the encrypted root image and fails, while the other launches the decrypted ELF.

It may be worth grouping scan candidates by normalized content identity (for example `PhysicalRootPath`, with care for archive sources) and preferring `decrypted/eboot.bin` when both candidates belong to the same extracted title. A regression test with the layout above would cover this case without executing guest code.

The full GameContent test suite passes here (172 tests), as do the 33 generator tests.

Aspenini:
Thanks for testing this. I’ve updated library scanning to group candidates by normalized content identity and prefer decrypted/eboot.bin when both executables belong to the same physical game. Archive sources remain distinct by archive path.
I also added regression coverage for the layout you described and rebased onto the latest main.

anes66 (changes requested):
The new normalized-identity selection fixes the duplicate retail/decrypted entries, and all 15 focused GameContent tests pass locally on .NET 10.0.103.

There is one scanner regression to address before approval. The old `ScanFolders` path explicitly caught `InvalidDataException` while opening each candidate, so one malformed `.zarchive` was skipped. `GameSourceDiscovery.OpenPreferredSources` now catches `IOException`, `UnauthorizedAccessException`, and `NotSupportedException`, but not `InvalidDataException`. Since `OpenPreferredSources` is called outside the per-source try/catch in `ScanFolders`, a malformed archive can now abort the entire library scan and prevent valid games from appearing.

Please include `InvalidDataException` in the candidate-open filter and add a regression case containing one malformed archive plus one valid physical source, asserting that the valid source is still returned.

Spooks4576:

Tested `zarchive-loading` @ `425383ce` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-296

par274:

Issue #11, opened 2026-07-01
Astro Bot [PPSA21564]


par274:
### Game Name

Astro Bot

### Title ID

PPSA21564

### Game Version

1.07

### Current Status

Crashes on boot

### Description

* `sceAgcInit` returns INVALID_ARGUMENT: the game requests register-defaults version 13; `IsSupportedRegisterDefaultsVersion` only accepts 7/8/10. 
* `sceAgcCreateShader "unexpected-registers type=4 lo=0x8A hi=0x8B"`: It doesn't recognize switch shader type 4 (GS phase; PGM_LO/HI offsets 0x8A/0x8B) in AgcExports.cs current map PS=0x8, ES=0xC8, LS=0x148, CS=0x20C. The game dereferences the null shader without checking for the error and crashes at address 0x8.

### Log File

[PPSA21564.txt](https://github.com/user-attachments/files/29888749/PPSA21564.txt)

### Operating System

Windows 11

### CPU

5800X3D

### GPU

4070S

### RAM

64 GB

### Emulator Version / Commit

c3019b7

### Renderer

Vulkan

12nikitosupporter-13:
I'm very interested in whether this game will be compatible, I really hope so. 
:3

Pgeniebox:
[LOADER][TRACE] VEH_AV first-chance at 0x000001F581D903C6 type=1 target=0x0000000300000020


[LOADER][DEBUG] TryResolveDirectImportTarget: 9BcDykPmo1I (libKernel:__error) -> HLE (kernel library)
[LOADER][DEBUG] SetupImportStubs: Trampoline for 9BcDykPmo1I -> 0x000001F5877F0000

par274:
#34 

Yes, this wasn't related to AVX. I found and confirmed this through in-depth research.

The problem is actually `map its heap to a fixed address` using the `MAP_FIXED` flag. Other games worked because they fulfilled this condition later.

After this stage, the game progresses to the AGC commands.

[PPSA21564.txt](https://github.com/user-attachments/files/29888749/PPSA21564.txt)

ynckz:
so it should work now? i'm too dumb to understand

par274:
> so it should work now? i'm too dumb to understand

The right question is how it works. Yes, there are no obstacles right now, and yes, there are obstacles to the image right now. We are working on it. However, we're not working on a single game specifically, but on compatibility in general. If you understand how the real hardware works, these games will already work.

[...]

PR #297, opened 2026-07-16
[SaveData] Implement save data memory2 exports

samto6:
Part of #11 (Astro Bot).

## What

Implements the Gen5 save data memory surface: `sceSaveDataSetupSaveDataMemory2`,
`sceSaveDataGetSaveDataMemory2`, `sceSaveDataSetSaveDataMemory2`, and
`sceSaveDataSyncSaveDataMemory`.

## Why

Astro Bot calls `sceSaveDataSetupSaveDataMemory2` during boot and asserts and
null-writes when the call fails, so the missing import surfaces as a named
crash. Setup alone would only relocate the crash to the first autosave, so the
companion get, set, and sync operations are included.

## How

- The store is one zero-filled file per user and title at
  `sce_sdmemory/memory.dat` under the save root (`SHARPEMU_SAVEDATA_DIR`).
- Readiness is the backing file's existence. Get, set, and sync return
  `SCE_SAVE_DATA_ERROR_MEMORY_NOT_READY` before setup.
- Setup grows the file when the requested size is larger and reports the
  previously existing size through the result struct, matching the retail
  "existedMemorySize" semantics.
- Transfers are range-checked against the file length and go through a single
  rented buffer. On set, guest memory is read in full before the file is
  touched, so a guest memory fault never leaves a partial write behind.
- Setup sizes are bounded by a 64 MiB emulator guard. It protects
  `SetLength` from corrupt or misread guest structs and is not a platform
  limit; the code comments it as such.
- Struct offsets (setup param: option 0x00, userId 0x04, memorySize 0x08;
  get/set param: userId 0x00, data pointer 0x08; data struct: buffer, size,
  offset) follow the publicly documented homebrew savedata headers.

## Verification

- 14 new tests cover setup and regrow semantics, small and 192 KiB round
  trips through the backing file, out-of-range rejection on both get and set,
  and not-ready behavior before setup and after the backing file is removed.
- Full `SharpEmu.Libs.Tests` suite passes (173 tests).
- Astro Bot: setup no longer fails during boot. Together with the companion
  stub PR (#298) the title reaches its splash image.

### DIFF
--- a/src/SharpEmu.Libs/SaveData/SaveDataExports.cs
+++ b/src/SharpEmu.Libs/SaveData/SaveDataExports.cs
@@ -719,24 +719,28 @@ public static class SaveDataExports
         try
         {
             var path = ResolveSaveDataMemoryPath(userId);
-            ulong existedSize;
             lock (_memoryGate)
             {
-                Directory.CreateDirectory(Path.GetDirectoryName(path)!);
-                using var stream = new FileStream(path, FileMode.OpenOrCreate, FileAccess.ReadWrite);
-                existedSize = (ulong)stream.Length;
+                var backing = new FileInfo(path);
+                var existedSize = backing.Exists ? (ulong)backing.Length : 0;
+
+                // The result write comes first so a faulted result pointer
+                // cannot leave created or grown setup state behind.
+                if (resultAddress != 0 && !ctx.TryWriteUInt64(resultAddress, existedSize))
+                {
+                    return ctx.SetReturn((int)OrbisGen2Result.ORBIS_GEN2_ERROR_MEMORY_FAULT);
+                }
+
                 if (existedSize < memorySize)
                 {
+                    Directory.CreateDirectory(Path.GetDirectoryName(path)!);
+                    using var stream = new FileStream(path, FileMode.OpenOrCreate, FileAccess.ReadWrite);
                     stream.SetLength((long)memorySize);
                 }
+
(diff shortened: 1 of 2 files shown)


anes66:
There is one failure-atomicity edge case in `sceSaveDataSetupSaveDataMemory2`: the backing file is created/grown before `resultAddress` is written. If the non-null result pointer is invalid, the export returns `ORBIS_GEN2_ERROR_MEMORY_FAULT`, but the failed call has already persisted the requested setup state.

I reproduced this with a focused test using a valid setup struct and an out-of-range result pointer. The call returned an error as expected, while `sce_sdmemory/memory.dat` still existed at the requested length afterward.

Could the implementation either validate/commit the result output before mutating the backing file, or roll the file back when the result write fails? A regression test asserting that an invalid result pointer does not create or grow the file would cover both the new-file and existing-file cases.

The 14 tests added by this PR pass unchanged on .NET 10.0.3.

samto6:
Yes; it now reads the existing size with `FileInfo` and writes the result output before touching the file, so a faulted result pointer returns before anything is created or grown.

Added both regression tests (new file and existing file).

anes66 (approved):
Verified commit `adcf4a7` locally on .NET 10.0.103.

The setup result is now written before any backing-file mutation, and both requested regression cases are covered: an invalid result pointer neither creates a new file nor grows an existing one. All 16 focused `SaveDataMemoryExportsTests` pass.

The failure-atomicity concern is resolved.

PR #298, opened 2026-07-16
[HLE] Stub ContentExport, Font, and Pad calls Astro Bot needs to boot

samto6:
Part of #11 (Astro Bot).

## What

- `sceContentExportInit2` (new libSceContentExport surface): reports success.
  No host media library exists to export captures into, so there is nothing
  more for init to do.
- Eight libSceFont calls: `sceFontBindRenderer`, the scale and effect setters,
  and their setup-render variants report success.
  `sceFontGetHorizontalLayout` writes baseline, line advance, and decoration
  extent using the same invented geometry as the existing
  `sceFontGetRenderCharGlyphMetrics`.
- `scePadSetTiltCorrectionState`: validates the handle like the neighboring
  pad setters and reports success.

## Why

Astro Bot asserts and null-writes when a subsystem init call fails, so each
missing import surfaces as a named crash. These are the blockers observed
during bring-up after save data memory setup (companion PR #297).

## Verification

- New tests: `sceFontGetHorizontalLayout` writes exactly three floats with a
  sentinel guarding the byte after the struct, and
  `scePadSetTiltCorrectionState` accepts primary handles and rejects others
  with `SCE_PAD_ERROR_INVALID_HANDLE`.
- Full `SharpEmu.Libs.Tests` suite passes (164 tests).
- Astro Bot: together with the companion save data memory2 PR (#297) the title
  boots to its splash image and runs its font glyph rendering path.

### DIFF
--- a/src/SharpEmu.Libs/ContentExport/ContentExportExports.cs
+++ b/src/SharpEmu.Libs/ContentExport/ContentExportExports.cs
@@ -1,0 +1,18 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+using SharpEmu.HLE;
+
+namespace SharpEmu.Libs.ContentExport;
+
+// No host media library exists to export captures into, so initialization
+// reports success.
+public static class ContentExportExports
+{
+    [SysAbiExport(
+        Nid = "0GnN4QCgIfs",
+        ExportName = "sceContentExportInit2",
+        Target = Generation.Gen5,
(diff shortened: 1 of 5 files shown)


anes66 (approved):
Reviewed the added ContentExport, Font, and Pad surfaces. The horizontal-layout ABI size and field order match the three-float layout, the pad export preserves the neighboring handle-validation behavior, and the new tests cover the output boundary and accepted handles.

Verified the full `SharpEmu.Libs.Tests` suite locally on .NET 10.0.3: 164/164 passed.

PR #300, opened 2026-07-16
Fix Linux POSIX signal register mapping

j92580498-max:
## Root cause

On Linux, `ucontext_t.uc_mcontext.gregs` stores the x86-64 general registers in the order defined by `REG_*`, but the POSIX signal bridge treated that array as if it were the contiguous Win64 `CONTEXT` register block. It therefore reconstructed the wrong guest RIP/RSP/GPR values before invoking the common recovery and diagnostic path. The supplied Record of Lodoss War run consequently reported a bogus unrecovered access violation, then corrupted the signal-frame stack and terminated with `*** stack smashing detected ***`.

## Fix

Use the Linux `gregs` offsets for each register and keep the Win64-shaped temporary exception record sized by the shared context-offset definition. Darwin's existing mapping is unchanged.

## Validation

- `dotnet build SharpEmu.slnx --configuration Release --no-restore`
- `dotnet test tests/SharpEmu.Libs.Tests/SharpEmu.Libs.Tests.csproj --configuration Release --no-build --no-restore`
- 159 tests passed
- The supplied crash log was traced to the Linux signal-register ABI mismatch; the exact game run is not executed in this container because the test environment does not provide the repository's GPU/display test harness.


j92580498-max:
Closing this draft while the POSIX crash diagnosis is revalidated; the submitted register mapping change is not yet confirmed.

PR #304, opened 2026-07-16
[Kernel] Implement sceKernelDirectMemoryQuery flags (find-next block)

lordnyx0:
The flags parameter (RSI) passed to sceKernelDirectMemoryQuery was previously ignored. When bit 0 of flags is set, the syscall is documented to return the first direct-memory allocation whose start address is strictly greater than the given offset, rather than the allocation that contains the offset.

This 'find-next' mode is used by titles that enumerate all direct memory pools by walking from offset 0 upwards until NOT_FOUND is returned. Without this fix, repeated calls with flags=1 all return the same first block, causing an infinite loop or a stale-memory read that can crash the engine.

Changes:
- KernelMemoryCompatExports.cs: read and apply the flags argument; when flags & 0x1, scan _directAllocations for the block with the smallest Start > offset instead of the containing block.
- KernelMemoryCompatExportsTests.cs: add three new Fact tests: DirectMemoryQuery_WithFlags0_ReturnsContainingBlock DirectMemoryQuery_WithFlags1_ReturnsNextBlock DirectMemoryQuery_WithFlags1_PastLastBlock_ReturnsNotFound

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
@@ -3282,7 +3282,7 @@ public static partial class KernelMemoryCompatExports
     public static int KernelDirectMemoryQuery(CpuContext ctx)
     {
         var offset = ctx[CpuRegister.Rdi];
-        _ = ctx[CpuRegister.Rsi]; // flags
+        var flags = ctx[CpuRegister.Rsi];
         var infoAddress = ctx[CpuRegister.Rdx];
         var infoSize = ctx[CpuRegister.Rcx];
         if (infoAddress == 0 || infoSize < 24)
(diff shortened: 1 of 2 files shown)


anes66 (changes requested):
There are two correctness issues to address before this is safe to merge:

1. `flags & 1` should not skip an allocation that already contains `offset`. Both the current shadPS4 direct-memory query path and the PS5-oriented `ps5ys` implementation first return the containing direct-memory area; the find-next behavior is only used when the address falls in a gap. The analogous `sceKernelVirtualQuery` implementation in this repository follows the same pattern. With the current change, querying the start or middle of block A with `flags=1` incorrectly returns block B (or `NOT_FOUND` when A is the last block).

   Please preserve the containing-block lookup first, then—only when no allocation contains the offset and bit 0 is set—select the allocation with the smallest start greater than the offset. Tests should cover: inside/exact start returns the current block, an address in a gap returns the next block with bit 0 set, the same gap returns `NOT_FOUND` without it, and an address after the final block returns `NOT_FOUND`.

2. The new tests leak allocations into the static `_directAllocations` table. This makes them order-dependent: for example, `DirectMemoryQuery_WithFlags1_PastLastBlock_ReturnsNotFound` can observe a block left by another test and stop being a true “past last” case. Please release every allocated range in `finally`, or add a test-only reset plus a non-parallel collection so each test starts and ends with deterministic state.

The three focused tests pass on this head, but they currently encode the skip-current interpretation and do not isolate the shared allocator state.

Spooks4576:

Tested `fix/direct-memory-query-flags` @ `d8ca782e` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-304

PR #305, opened 2026-07-16
[Perf] Reduce runtime and Vulkan draw overhead

xnetcat:
## Summary

- replace the per-continuation capturing emitter helpers with a stack-only writer, removing a compiler-generated display-class allocation from every blocked guest-thread resume
- skip AGC wait-registry snapshots and detailed producer strings when AGC tracing is disabled, and remove several per-draw LINQ intermediates
- keep Vulkan debug-utils names and command labels off during normal gameplay; validation still enables them automatically and `SHARPEMU_VK_DEBUG_LABELS=1` enables them explicitly
- cache startup-only Vulkan diagnostic settings and avoid inactive trace-image/hash-set work and per-draw validation collections

## Relationship to existing work

This branch is based on current `main`, which includes #301 and #290. It does not duplicate the open shader, pipeline, render-pass, guest-image, scheduler, or synchronization PRs.

## Measurement

Terrarium was profiled in the same rendered gameplay scene on an Apple M4 through the macOS x64/Rosetta build:

- managed allocation rate: **70.6 -> 48.8 MB/s** (about **31% lower**)
- gen0 collections: **7.64 -> 5.63/s** (about **26% lower**)
- FPS remained scene/noise sensitive in the 10-16 range, so this PR does not claim a reproducible frame-rate uplift from one machine; it removes measured runtime overhead and GC pressure that affects frame consistency and other games as well

A more aggressive guest-runner scheduling experiment was discarded after it failed runtime validation.

## Validation

- Release solution build: zero warnings/errors
- tests: 199 passed
- publishes: macOS x64, Linux x64, Windows x64
- Terrarium: reaches and renders gameplay on Apple M4 with the merged #301/#290 path

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -5127,64 +5127,51 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 			BindTlsBase(context);
 			byte* ptr2 = (byte*)ptr;
 			ulong hostRspSlot = (ulong)hostRspStorage;
-			int offset = 0;
+			var emitter = new NativeCodeEmitter(ptr2);
 
-			void Emit(byte value) => ptr2[offset++] = value;
-			void EmitU64(ulong value)
-			{
-				*(ulong*)(ptr2 + offset) = value;
-				offset += sizeof(ulong);
-			}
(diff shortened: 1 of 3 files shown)


xnetcat:
Superseded by the independently reviewable performance PRs #306, #308, and #307. The changes and validation are unchanged; this avoids coupling the CPU, AGC, and Vulkan optimizations in one broad draft.

PR #311, opened 2026-07-17
[Agc] Report base console from the Trinity mode query

samto6:
## What this adds

`sceAgcGetIsTrinityMode` (`BfBDZGbti7A`, real catalog name): returns 0.

## Why

The emulator models the base console; Trinity is the PS5 Pro hardware mode,
so the accurate answer is a constant. Titles probe this to select quality
modes and Pro-only register paths, and with the export missing every such
probe surfaced as an unresolved import. Astro Bot's boot-time query (#11)
surfaced it first.

## Verification

- The query resolves and callers take the base-console path (verified with
  Astro Bot's boot). No memory is written, so there is nothing beyond the
  constant to test; `SharpEmu.Libs.Tests` passes unchanged (210 tests).

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -632,6 +632,18 @@ public static partial class AgcExports
     public static int GetRegisterDefaults2Internal(CpuContext ctx) =>
         ReturnRegisterDefaults(ctx, internalDefaults: true);
 
+    // The emulator models the base console; Trinity is the PS5 Pro hardware mode.
+    [SysAbiExport(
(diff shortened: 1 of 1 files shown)


anes66 (approved):
The base-console return value and Gen5-only registration are consistent with the emulator's modeled hardware. The generated export registry also covers duplicate NID/name validation, and all repository checks pass on this head. A focused lookup/invocation test would be useful future coverage, but I do not consider it a merge blocker for this constant query.

par274:
This isn't how the real hardware behaves; it only just a stub. If we make it return success, the game may detect the system as a Pro console, which is not the behavior we want.

PR #312, opened 2026-07-17
ci: rebuild the website when a release is published

xnetcat:
Publishing a release now also refreshes https://sharpemu.app/downloads, which reads the release list at build time.

On `release: published` (and edited/deleted), this workflow sends a `repository_dispatch` to `sharpemu/sharpemu-site`, whose deploy workflow rebuilds and ships to Cloudflare Pages — the new build shows up within about a minute instead of waiting for the next site commit.

**One-time setup:** add a repository secret `SITE_DISPATCH_TOKEN` — a fine-grained PAT scoped to `sharpemu/sharpemu-site` with **Contents: read and write**. Without it, the job logs a notice and exits 0, so releases never fail because of the website.

No emulator code, build, or release artifacts are touched. The site repo also rebuilds daily as a safety net.

### DIFF
--- a/.github/workflows/notify-site.yml
+++ b/.github/workflows/notify-site.yml
@@ -1,0 +1,28 @@
+# Tells the website repo to rebuild when a release is published, so
+# sharpemu.app/downloads lists the new build within a minute.
+name: Notify website
+
+on:
+  release:
+    types: [published, released, edited, deleted]
+  workflow_dispatch:
+
+jobs:
+  dispatch:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


anes66 (commented):
The failing REUSE checks are both caused by the new workflow lacking SPDX metadata. Please add the same header used by the other workflow files at the top of `.github/workflows/notify-site.yml`:

```yaml
# Copyright (C) 2026 SharpEmu Emulator Project
# SPDX-License-Identifier: GPL-2.0-or-later
```

The Actions log reports this file as the only one missing copyright and licensing information (367/368 files compliant).

PR #321, opened 2026-07-17
[CI] Build release workflow for Windows + Linux

Sh-TB:
## Summary

Adds GitHub Actions workflow that builds SharpEmu for both platforms on every tag push.

## What it does

On `git tag v*` + `git push`:
1. Builds **Windows x64** (self-contained) → `SharpEmu-win-x64.zip`
2. Builds **Linux x64** (self-contained) → `SharpEmu-linux-x64.tar.gz`
3. Uploads both as release assets

## Workflow file

`.github/workflows/build-release.yml`

- Uses `actions/checkout@v4`
- Uses `actions/setup-dotnet@v4` with .NET 10
- Linux build includes `mesa-vulkan-drivers` for headless Vulkan support
- Release upload uses `softprops/action-gh-release@v2`
- Proper `permissions: contents: write` for release upload

## No Breaking Changes

Only adds a new workflow file. Does not modify any existing code.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -1,0 +1,64 @@
+name: Build Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  build-windows:
+    runs-on: windows-latest
+    steps:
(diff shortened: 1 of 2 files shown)


par274:
We have already for release CI. And v4 is deprecated.

PR #322, opened 2026-07-17
Fix event flag wake check and POSIX open/fstat ABI

NickPom:
## Before submitting

➡️ [**CONTRIBUTING.md**](CONTRIBUTING.md)

## Checklist
By submitting this PR, you confirm that:

- [x] I tested my changes.
- [ ] I wrote this description myself and did not copy AI-generated explanations.
- [x] I listed the game(s) I tested, if applicable.

> This PR description was drafted with AI assistance (research, tracing, and writeup). I reviewed it, understand every line of the change, and can explain/maintain it. Leaving that box unchecked rather than claiming otherwise.

## Summary

- `WakeBlockedThreads` only checked `thread.BlockWaiter` before marking a parked thread Ready. Threads blocked via the other registration overload (`BlockResumeHandler`/`BlockWakeHandler`, used by `sceKernelWaitEventFlag`) have `BlockWaiter == null`, so that check was silently skipped and the thread got marked Ready without ever confirming its wait condition (event-flag pattern) was actually satisfied. It resumed still unsatisfied.
- `open()` (`wuCroIGjt2g`) and `fstat()` (`mqQMh1zPPT8`) called straight into the Orbis-convention `KernelOpenUnderscore`/`KernelFstat` without converting the result to the POSIX ABI on failure, same issue already documented and handled for `stat`/`PosixStat` a few lines above. A failed `open()` returned the raw Orbis error code (`ORBIS_GEN2_ERROR_NOT_FOUND` = `0x80020002`) directly in RAX instead of -1 with errno set, so a caller checking `fd == -1` missed the failure and went on to call `fstat`/`lseek`/`close`/`read` with that value as the file descriptor.
- Bonus: made the Vulkan queue-visibility fence wait in `VulkanVideoPresenter` bounded (chunked retries with a log line) instead of an unconditional `vkWaitForFences(..., ulong.MaxValue)`, so a submission whose fence never actually signals shows up as a clear repeating log line instead of a silent, permanent hang. Purely diagnostic, the wait still continues past a logged timeout exactly like before.

## Test plan

Tested on macOS (Apple Silicon, Rosetta 2) against **Arcade Game Zone [PPSA19015]**, a Unity/IL2CPP title, in combination with the exception-delivery fix from #253:

- Before the event-flag fix: hard stall at import 81408, every time, with a thread parked on `sceKernelWaitEventFlag` never getting woken despite its condition being met.
- With it: the game gets past multiple GC stop-the-world cycles, initializes the full Unity job system (`Job.worker 0-12`, `Background Job.worker 0-15`, FMOD audio, `GfxFlipThread`, `UnityGfxDeviceWorker`), and renders real frames (splash + first gameplay frame at 1920x1080).
- Before the open/fstat fix: once past the deadlock, the process hit an unrecovered SIGSEGV (fault address 0x0) a bit further in, traced to a failed `open()` on a savedata path whose bogus "successful" return value (the raw Orbis error code) got passed straight into `fstat` as a file descriptor.
- With it: that crash is gone; the failure is now reported through errno like real POSIX `open`/`fstat` and the caller's own error handling takes over correctly.

Left a detailed writeup of a further, still-open issue (a semaphore-count corruption on repeated GC cycles, matching a point already raised in review on #253) as a comment there rather than attempting a blind fix, since properly fixing it needs the interrupt-and-retry redesign already outlined by the reviewer and touches CPU-context redirection machinery I don't have full context on.

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -3141,6 +3141,19 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 					continue;
 				}
 
+				// BlockWaiter and BlockWakeHandler are mutually exclusive (set by
+				// the two RegisterBlockedGuestThreadContinuation overloads), but a
+				// thread parked via the resume/wake-handler overload (e.g.
+				// sceKernelWaitEventFlag) has BlockWaiter == null, so the check
+				// above never runs for it. Without this, WakeBlockedThreads marks
+				// it Ready unconditionally on any matching wake key, without ever
+				// confirming its actual wait condition (event-flag pattern, etc.)
+				// is satisfied -- the thread resumes later still unsatisfied.
+				if (thread.BlockWakeHandler is not null && !thread.BlockWakeHandler())
+				{
+					continue;
+				}
+
 				thread.State = GuestThreadRunState.Ready;
 				thread.BlockReason = null;
 				thread.BlockDeadlineTimestamp = 0;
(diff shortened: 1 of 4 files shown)


par274:

NickPom:
Sorry about that, missed the template on the first submission. Fixed now, checklist filled in above.

PR #326, opened 2026-07-17
[Agc] Implement fused shader half exports

samto6:
## What this adds

Two libSceAgc NIDs captured from shipped titles, following the existing captured-NID convention (`23LRUSvYu1M`, `HV4j+E0MBHE`, `V++UgBtQhn0`):

- `dolOmWH+huQ` as `sceAgcGetFusedShaderSize`: reports the byte size of the back half's SH register image (count x 8, align 4).
- `fd5Bp5tGTgo` as `sceAgcFuseShaderHalves`: joins a GS or HS front/back shader half pair into one shader. The fused header is the back half retyped to kGs/kHs with cleared user data; the back half's SH registers are copied into caller scratch when provided (patched in place otherwise); the front half donates its program address (SPI_SHADER_PGM_LO/HI_ES for GS, LO/HI_LS for HS, hi keeps its upper 24 bits) and, for GS pairs, both SPI_SHADER_PGM_CHKSUM_GS occurrences.

## Why

libSceAgc ships geometry and tessellation shaders as front/back half pairs that the runtime joins at load time; the emulator implements shader creation but had no join, so both NIDs were unresolved. A title that fuses shaders then works with a garbage fused struct: the size query returns nothing, the caller passes a null scratch pointer, and every pipeline built from the pair inherits the garbage. Any Gen5 title running fused geometry pipelines goes through these two calls; Astro Bot (#11) is the first observed consumer, with 33 GS pairs fused during boot.

## How

- Semantics follow KytyPS5's reverse-engineered `agc.cpp` (`GraphicsUnknownGetFusedShaderSize` / `GraphicsUnknownFuseShaderHalves`), cross-checked against this title: shader struct offsets match the ones `sceAgcCreateShader` already uses, and a traced run shows the game allocating scratch at exactly the size the query reports (10 registers -> 0x50-byte strides) and passing type 4/6 pairs.
- The true symbol names are unknown: both flat `sceAgc*` names and the catalog's mangled `sce::Agc::{getFusedShaderSize,fuseShaderHalves}` signatures hash to different NIDs (checked against scripts/ps5_names.txt and generated mangling variants), so the exports carry the same captured-NID comment and SHEM004 suppression as the existing trio. The catalog lists both flat names, and the NID derivation assigns them `nApJjpKNBl4` and `nQT5kYLv0cg`, so the real catalog demonstrably uses these names for other symbols.
- Every fuse-specific behavior is taken from the KytyPS5 routines: the VGT_SHADER_STAGES_EN wave-size gate (GS_W32_EN bit 22 and HS_W32_EN bit 21 per the public GFX10 register headers, so the gate is both halves agreeing on wave32 mode; skipped when either half lacks specials), the program-address patch preserving the hi register's upper 24 bits (the create-time patch owns the whole dword), a missing or unpaired lo/hi register being left untouched instead of failing like the create-time patch (traced as `pgm_absent`/`pgm_unpaired` so a skipped patch leaves a breadcrumb), and the two-occurrence CHKSUM_GS graft (images carry the register twice with distinct values, so the copy is positional rather than copy-all-matches; the register image is searched by offset rather than assuming array position).
- The create-time patch's shader-type switch and IsEsGeometryShaderType now share the named type constants this change introduces instead of repeating the raw values.
- Not addressed here: the GS-phase program pointer (0x88/0x89) of the back half is never rebased to an absolute address (create only patches the first register pair, 0xC8/0xC9). The draw path currently consumes only the ES pointer, which the fuse sets to the front half's code, so this only matters once the GPU learns to run the GS phase itself.

## Verification

- 10 new tests cover the size query, pair-type rejection, the fused header fields (retype, cleared user data, relocated register image), scratch and in-place register flows, hi-register bit preservation, checksum grafting on GS pairs and its absence on HS pairs, wave-size mismatch rejection and its skip when a half lacks specials, and the leave-untouched patch paths (program register absent, unpaired, or at the image end).
- Full `SharpEmu.Libs.Tests` suite passes (220 tests).
- Astro Bot: all 33 boot-time fuses succeed with well-formed inputs (`types=4/6 registers=10`), and a traced run confirms the game consumes the size query: it allocates scratch blocks at exactly the reported size (0x50-byte strides) and passes them to the fuse call, where the unresolved stub previously left it passing null. The title still parks at its splash afterwards for unrelated reasons (it never issues its first draw), tracked in #11.

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -78,6 +78,7 @@ public static partial class AgcExports
     private const uint SpiShaderPgmHiLs = 0x149;
     private const uint SpiShaderPgmLoGs = 0x8A;
     private const uint SpiShaderPgmHiGs = 0x8B;
+    private const uint SpiShaderPgmChksumGs = 0x80;
     private const uint SpiPsInputEna = 0x1B3;
     private const uint SpiPsInputAddr = 0x1B4;
     private const uint ComputePgmLo = 0x20C;
@@ -173,6 +174,16 @@ public static partial class AgcExports
     private const ulong ShaderNumOutputSemanticsOffset = 0x56;
     private const ulong ShaderTypeOffset = 0x5A;
     private const ulong ShaderNumShRegistersOffset = 0x5C;
+    private const int ShaderStructBytes = 0x60;
+    private const ulong FusedShaderImageAlignment = 4;
+    private const byte ComputeShaderType = 0;
+    private const byte PsShaderType = 1;
+    private const byte GsShaderType = 2;
+    private const byte HsShaderType = 3;
+    private const byte GsFrontShaderType = 4;
+    private const byte HsFrontShaderType = 5;
+    private const byte GsBackShaderType = 6;
+    private const byte HsBackShaderType = 7;
     private const ulong CommandBufferCursorUpOffset = 0x10;
     private const ulong CommandBufferCursorDownOffset = 0x18;
     private const ulong CommandBufferCallbackOffset = 0x20;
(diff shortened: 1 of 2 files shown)


par274:

PR #329, opened 2026-07-17
[VideoOut] Rank integrated GPUs above software rasterizers

### LINKED ISSUE #325: [VideoOut] Software CPU rasterizer outranks integrated GPU in ScorePhysicalDevice
### Emulator Version / Commit

0.0.2-beta.2 / 24b82a7

### Bug Description

`ScorePhysicalDevice` in `src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs` ranks a software CPU rasterizer **above** any integrated GPU:

```csharp
var score = properties.DeviceType switch
{
    PhysicalDeviceType.DiscreteGpu => 300,
    PhysicalDeviceType.VirtualGpu => 100,
    PhysicalDeviceType.Cpu => 50,
    PhysicalDeviceType.IntegratedGpu => -100,
    _ => 10,
};
```

#97 introduced the -100 deliberately, to stop hybrid-graphics laptops from picking an integrated AMD GPU whose driver access-violates inside `vkCreateGraphicsPipelines`. The stated rule was "an integrated GPU is only chosen when nothing else can present." The problem is that `PhysicalDeviceType.Cpu` counts as "something else": a software rasterizer scores 50 and wins against a real integrated GPU at -100.

Two consequences:

1. **Linux, integrated-only hosts.** Mesa ships lavapipe, which enumerates as `VK_PHYSICAL_DEVICE_TYPE_CPU`. On a machine whose only real GPU is Intel/AMD integrated, SharpEmu would select lavapipe (score 50) over the iGPU (score -100) and silently software-render. The user gets no warning — just an unplayable frame rate that looks like an emulator performance problem rather than a device-selection one.

2. **Apple Silicon.** MoltenVK reports every Apple Silicon GPU as `IntegratedGpu`, so an M-series GPU always scores -100. It is selected today only because it is the sole candidate (`bestScore` starts at `int.MinValue`, so -100 wins uncontested). Any future `Cpu`-type ICD alongside MoltenVK would displace it.

The AMD-integrated crash that motivated #97 is a driver/vendor problem, but the fix is expressed as a device-*type* penalty, which catches every integrated GPU on every platform. A vendor/device-specific block, or simply ranking `IntegratedGpu` above `Cpu`, would preserve #97's intent without making software rendering a preferred outcome.

**Expected:** a real GPU is preferred over a software rasterizer.
**Actual:** a `Cpu`-type device outranks any `IntegratedGpu`.

**Reproduction note:** case 2 is observed (log below). Case 1 is derived from reading the scoring code — I do not have an integrated-only Linux host with lavapipe installed to confirm it directly, so that half should be verified before fixing.

### Log File

```shell
Observed on an Apple M1 Max (macOS 26.5.2, osx-x64 build under Rosetta 2), Dead Cells PPSA15552.
MoltenVK is the only ICD present, so the GPU wins despite the negative score:

[mvk-info] Created VkInstance for Vulkan version 1.2.323 ...
[LOADER][INFO] Vulkan candidate: Apple M1 Max (IntegratedGpu) score=-100
[mvk-info] Created VkDevice to run on GPU Apple M1 Max ...
[LOADER][INFO] Vulkan VideoOut ready: 2560x1440, format=B8G8R8A8Unorm
```

chengboonrong:
Fixes #325

## Before submitting

➡️ [**CONTRIBUTING.md**](CONTRIBUTING.md)

## Checklist
By submitting this PR, you confirm that:

- [ ] I tested my changes.
- [x] I wrote this description myself and did not copy AI-generated explanations.
- [x] I listed the game(s) I tested, if applicable.

---

## Notes (raw material — delete before merge)

**The bug**
- `ScorePhysicalDevice` scored `Cpu` = 50, `IntegratedGpu` = -100 → a software rasterizer outranked every integrated GPU.
- #97 set the -100 deliberately (integrated AMD driver access-violates in `vkCreateGraphicsPipelines` compiling translated guest shaders). Its rule: integrated is chosen only when nothing else can present. A `Cpu`-type device counts as "something else" — that's the hole.

**Impact**
- Linux, integrated-only host: Mesa's lavapipe enumerates as `Cpu` → wins → guest software-renders, no warning. Reads as an emulator perf problem, not a device-selection one.
- Apple Silicon: MoltenVK reports Apple GPUs as `IntegratedGpu`. Wins today only because it's the sole candidate (`bestScore` starts at `int.MinValue`). One `Cpu`-type ICD away from the same fate.

**The change**
- `IntegratedGpu` 50, `Cpu` -100. Everything else unchanged.
- Discrete still beats integrated by 250 (750 for NVIDIA) → #97's hybrid-laptop behaviour is not altered. Only the software-vs-integrated comparison flips.
- `ScorePhysicalDevice` moved from the private nested `Presenter` to the outer class as `internal`, so it's reachable from `SharpEmu.Libs.Tests` (which already has `InternalsVisibleTo`). Mirrors the existing `ShouldAttachGuestDepth` precedent. No behaviour change from the move — the call site is unchanged.

**Verification**
- New `VulkanPhysicalDeviceScoreTests` (8 cases). Confirmed they *fail* on the old ordering — reverted the two values and got exactly 3 failures (`IntegratedGpu`, `Other`, and the Apple Silicon case) while the #97 discrete-beats-integrated, NVIDIA, and `SHARPEMU_VK_DEVICE` override tests still passed. Restored → all 8 pass.
- Full suite green.
- Real run, Apple M1 Max / macOS 26.5.2 / `osx-x64` under Rosetta 2, Dead Cells [PPSA15552]:
  - before: `Vulkan candidate: Apple M1 Max (IntegratedGpu) score=-100`
  - after: `Vulkan candidate: Apple M1 Max (IntegratedGpu) score=50`
  - same device selected, `VideoOut ready: 2560x1440`, first frame presented.

**Not verified**
- The lavapipe scenario itself — no integrated-only Linux host with a software ICD available here. That half is reasoned from the scoring code; the fix follows regardless, since it only reorders.
- Hybrid-graphics laptop (the #97 hardware). Argument that it's unaffected is arithmetic, not a test run.

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
+++ b/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
@@ -2219,6 +2219,47 @@ internal static unsafe class VulkanVideoPresenter
         target is not null &&
         (state.TestEnable || state.WriteEnable || state.ClearEnable);
 
+    /// <summary>
+    /// Ranks a Vulkan candidate; the highest score presents. SHARPEMU_VK_DEVICE
+    /// pins an adapter by name substring and overrides the ranking entirely.
+    /// </summary>
+    /// <remarks>
+    /// Discrete parts outrank integrated because an integrated AMD driver
+    /// access-violates inside vkCreateGraphicsPipelines while compiling some
+    /// translated guest shaders (#97). That penalty must not extend below
+    /// <see cref="PhysicalDeviceType.Cpu"/>, which is a software rasterizer
+    /// (Mesa's lavapipe, SwiftShader): rendering the guest on the CPU is slower
+    /// than any real GPU by orders of magnitude, so it stays the last resort.
+    /// MoltenVK also reports every Apple Silicon GPU as integrated, which would
+    /// otherwise lose to a software device on those hosts.
+    /// </remarks>
+    internal static int ScorePhysicalDevice(
+        PhysicalDeviceProperties properties,
+        string name,
+        string? deviceOverride)
+    {
+        if (!string.IsNullOrWhiteSpace(deviceOverride))
+        {
+            return name.Contains(deviceOverride, StringComparison.OrdinalIgnoreCase) ? 1000 : -1000;
+        }
+
+        var score = properties.DeviceType switch
+        {
+            PhysicalDeviceType.DiscreteGpu => 300,
+            PhysicalDeviceType.VirtualGpu => 100,
+            PhysicalDeviceType.IntegratedGpu => 50,
+            PhysicalDeviceType.Cpu => -100,
+            _ => 10,
+        };
+
+        if (properties.VendorID == NvidiaVendorId)
+        {
+            score += 500;
+        }
+
+        return score;
+    }
+
     private readonly record struct Presentation(
         byte[]? Pixels,
         uint Width,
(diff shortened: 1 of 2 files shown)


par274:

PR #338, opened 2026-07-17
Generalize the SSE4a EXTRQ patch to any scratch register (fixes #328)

### LINKED ISSUE #328: [CPU] Guest SSE4a EXTRQ raises SIGILL on any host without SSE4a (Dead Cells PPSA15552)
### Emulator Version / Commit

0.0.2-beta.2 / 24b82a7 (`main`). Reproduced with a self-contained `osx-x64` publish built exactly as CI does.

**Platform:** Apple M1 Max, macOS 26.5.2 (Darwin 25.5.0), `osx-x64` build under Rosetta 2
**Game:** Dead Cells [PPSA15552]

> **Note:** an earlier revision of this issue claimed Rosetta 2 also lacks AVX and AVX2, and concluded that recovering `extrq` was pointless because the following AVX block would fault anyway. That was wrong and is corrected below — see the comments. AVX/AVX2 execute correctly under Rosetta 2; SSE4a is the only extension that actually faults.

### Bug Description

The emulator boots, selects the GPU, presents a first frame, and then dies deterministically with **exit 132 (SIGILL)** while executing guest code. Same faulting address on every run:

```
[LOADER][TRACE] posix-signal#1: sig=4 rip=0x000000080030C6ED fault=0x0 access=0 rsp=0x00006FFFF01FF280
[LOADER][INFO]   Exception Address: 0x000000080030C6ED
[LOADER][INFO]   Host thread: managed=4 name='SharpEmu Emulation'
[LOADER][TRACE] posix-signal#1: recovered=False new_rip=0x000000080030C6ED
```

**Root cause:** the guest instruction at that address is AMD-only SSE4a, which the host does not implement.

Resolving the guest address through the SELF container (program header 0 maps to physical file offset `0xc070` via the SELF segment table — not the naive ELF offset) and disassembling at the exact faulting RIP:

```
0x80030C6ED: 66 0f 78 c1 28 00        extrq xmm1, 0x28, 0            groups: ['sse4a']
0x80030C6F3: c4 e3 79 02 c1 02        vpblendd xmm0, xmm0, xmm1, 2   groups: ['avx2']
0x80030C6F9: c5 fa e6 c0              vcvtdq2pd xmm0, xmm0           groups: ['avx']
```

`extrq` is SSE4a. **SSE4a is an AMD-only extension — Intel has never implemented it.** This is therefore not a macOS or Rosetta-specific bug: the PS5's Zen 2 cores have SSE4a and Sony's toolchain emits it, so the same guest code aborts on **any Intel host**, Windows and Linux included, while running fine on an AMD desktop. Rosetta 2 is simply another host without it.

Only the `extrq` faults. The AVX2/AVX instructions that follow it execute normally (verified by executing them directly — see comments).

### Existing handling

`DirectExecutionBackend.TryPatchSse4aExtrqBlend` already rewrites this at load time, replacing the `extrq`+`vpblendd` pair with equivalent SSE4.1 `pextrb`/`pinsrd`. Its comment already records that Rosetta does not implement EXTRQ.

It does not fire here because it matches a **fixed 12-byte sequence with `xmm2` hard-coded**:

```
66 0F 78 C2 28 00     extrq    xmm2, 0x28, 0
C4 E3 79 02 C2 02     vpblendd xmm0, xmm0, xmm2, 2
```

PPSA15552 emits the identical sequence against `xmm1` (`...C1...` rather than `...C2...`), so the pattern misses by one byte per instruction and the opcode reaches the CPU. The startup log shows `Patched ... 0 SSE4a EXTRQ blends`.

**Expected:** guest SSE4a executes, or is recovered, on hosts that lack the extension.
**Actual:** SIGILL; the signal bridge catches it, cannot recover, and the process dies with 132.

**Relationship to #194:** likely unrelated. #194 is a Rosetta *thread-resume* failure (`ThreadContext::resume failed 37`) with no SharpEmu-side exception. This is a clean caught #UD at an identifiable instruction.

### Steps to reproduce

1. `dotnet publish src/SharpEmu.CLI/SharpEmu.CLI.csproj -c Release -r osx-x64 --self-contained true -p:PublishDir=<dir>`
2. `scripts/fetch-macos-moltenvk.sh <dir>`
3. `./SharpEmu /path/to/PPSA15552/eboot.bin`
4. Process exits 132 after the first presented frame.

An Intel host should reproduce the same fault without Rosetta involved, though I have no Intel machine to confirm that.

### Log File

```shell
[LOADER][INFO] Patched 0 TLS loads, 164 TLS stores, 0 stack-canary accesses, 0 SSE4a EXTRQ blends
...
[LOADER][TRACE] posix-signal#1: sig=4 rip=0x000000080030C6ED fault=0x0000000000000000 access=0 rsp=0x00006FFFF01FF280
[LOADER][INFO]   Exception Address: 0x000000080030C6ED
[LOADER][INFO]   RIP: 0x000000080030C6ED
[LOADER][INFO]   Host thread: managed=4 name='SharpEmu Emulation'
[LOADER][TRACE] posix-signal#1: recovered=False new_rip=0x000000080030C6ED

exit code: 132 (SIGILL), reproduced identically across runs with no external signal sent.
```

6can:
## What this fixes

`DirectExecutionBackend.TryPatchSse4aExtrqBlend` rewrites the AMD-only `extrq` + `vpblendd` idiom that Sony's toolchain emits, because SSE4a isn't implemented on Intel hosts or under Rosetta and the raw opcode raises #UD there. #328 points out that the matcher hard-codes the scratch register to `xmm2`:

```
66 0F 78 C2 28 00     extrq    xmm2, 0x28, 0
C4 E3 79 02 C2 02     vpblendd xmm0, xmm0, xmm2, 2
```

Dead Cells (PPSA15552) emits the identical idiom but against `xmm1` (`...C1...`), so the pattern misses by one byte per instruction, the `extrq` reaches the CPU, and the game dies with SIGILL right after the first frame. The `Patched ... 0 SSE4a EXTRQ blends` line in the startup log is the giveaway.

## The change

I moved the pattern match and the replacement encoding out of the backend into a small pure class, `Sse4aExtrqRewriter`, that reads the register straight out of the ModRM byte instead of comparing it against a literal. For register N it emits:

```
66 0F 3A 14 (C0|N<<3) 04   pextrb eax, xmmN, 4
66 0F 3A 22 C0 01          pinsrd xmm0, eax, 1
```

The equivalence is the one the original patch already relied on: `extrq xmmN, 0x28, 0` keeps the low 40 bits of `xmmN` and zeroes the rest, so the second dword `vpblendd` folds into `xmm0` is just byte 4 of the original `xmmN`, zero-extended. `pextrb`/`pinsrd` reproduce exactly that, using EAX as scratch (the toolchain only uses `xmmN` as a throwaway for the extract, so nothing downstream reads it back).

A couple of deliberate limits:

- The extract field (`0x28` length, `0` index) is still matched literally rather than decoded, because the byte-4 offset in the replacement is only the right answer for that specific field. A different extract width would need different replacement math, so it's safer not to match it at all.
- Only the scratch register is generalized. Destination/first-source stay pinned to `xmm0` and the blend mask to `2`, and there's still no REX prefix, so this doesn't widen what gets rewritten beyond the one idiom the toolchain emits (xmm0–7) — it just stops caring which of those registers it landed in.

The unsafe `VirtualProtect`/write/`FlushInstructionCache` plumbing stays in the backend; only the decision moved out, which is what makes it testable.

## How I verified it

- New unit tests in `Sse4aExtrqRewriterTests`. They pin the xmm2 case to the exact bytes the old hand-written patch produced (so the refactor is provably byte-identical for the case that already shipped), cover the xmm1 regression and all eight registers, and check that mismatched / short / malformed input is rejected.
- Cross-checked the raw bytes through Iced's disassembler: the inputs decode to `extrq xmmN, 28h, 0` / `vpblendd xmm0, xmm0, xmmN, 2` and the outputs to `pextrb eax, xmmN, 4` / `pinsrd xmm0, eax, 1` for N = 1, 2, 7.
- `dotnet build` and the full `dotnet test` run are clean (226 tests in the Libs suite, +16 from this change).

What I could not do: I don't have the game or an Intel / Rosetta machine, so I haven't watched PPSA15552 boot past the fault end to end. If anyone who can reproduce #328 is able to confirm the log now shows a non-zero `SSE4a EXTRQ blends` count and the SIGILL is gone, that would close the loop.

**Note:** this hasn't been tested in-game. I don't have PPSA15552 or an Intel/Rosetta machine, so the verification is at the unit-test and instruction-encoding level only (see above). If someone who can reproduce #328 confirms the SIGILL is gone in-game, that closes the loop.

Fixes #328.

### DIFF
--- a/src/SharpEmu.Core/Cpu/Emulation/Sse4aExtrqRewriter.cs
+++ b/src/SharpEmu.Core/Cpu/Emulation/Sse4aExtrqRewriter.cs
@@ -1,0 +1,98 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+namespace SharpEmu.Core.Cpu.Emulation;
+
+/// <summary>
+/// Rewrites the AMD SSE4a EXTRQ idiom that Sony's toolchain emits for the PS5's Zen 2 cores into an
+/// equivalent SSE4.1 sequence at load time.
+///
+/// SSE4a is an AMD-only extension; Intel has never shipped it. When the direct-execution backend runs
+/// the guest natively, the <c>extrq</c> opcode raises #UD on any Intel host (and on Rosetta 2), while
+/// the AVX/AVX2 code around it runs fine. The particular shape the toolchain produces is a fixed-width
+/// bit extract followed immediately by a <c>vpblendd</c> that folds the extracted dword back into
+/// <c>xmm0</c>. This class recognises that pair and produces a <c>pextrb</c>/<c>pinsrd</c> replacement
+/// with the same observable effect, which the backend patches over the guest bytes before first run.
+///
+/// The matcher is pure and works on plain byte spans so the encoding can be unit-tested on its own; the
+/// unsafe page-protection and instruction-cache plumbing stays in the backend adapter.
+/// </summary>
+public static class Sse4aExtrqRewriter
+{
+    /// <summary>Byte length of both the recognised idiom and its replacement. They are the same size,
+    /// so the patch is an in-place overwrite with no relocation.</summary>
+    public const int SequenceLength = 12;
+
+    /// <summary>
+    /// Matches
+    /// <code>
+    ///   66 0F 78 /0 28 00      extrq    xmmN, 0x28, 0
+    ///   C4 E3 79 02 C0|N 02    vpblendd xmm0, xmm0, xmmN, 2
+    /// </code>
+    /// for any scratch register <c>xmmN</c> (N in 0..7) and, on a match, writes
+    /// <code>
+    ///   66 0F 3A 14 (N&lt;&lt;3) 04   pextrb eax, xmmN, 4
+    ///   66 0F 3A 22 C0 01           pinsrd xmm0, eax, 1
+    /// </code>
+    /// into <paramref name="replacement"/>.
+    ///
+    /// Why these two instructions are equivalent: <c>extrq xmmN, 0x28, 0</c> keeps the low 40 bits of
+    /// <c>xmmN</c> and zeroes the rest, so the second dword <c>vpblendd</c> then copies into
+    /// <c>xmm0</c> is just byte 4 of the original <c>xmmN</c>, zero-extended. <c>pextrb</c> reads that
+    /// same byte into EAX and <c>pinsrd</c> drops it into dword lane 1 of <c>xmm0</c>. The extract
+    /// length (0x28) and index (0) are matched exactly rather than decoded, because the replacement's
+    /// byte offset of 4 is only the right answer for that specific field.
+    ///
+    /// Unlike the original, the replacement clobbers EAX instead of <c>xmmN</c>. That trade already
+    /// existed in the hand-written patch this generalises; the toolchain only uses <c>xmmN</c> here as a
+    /// throwaway for the extract, so nothing downstream depends on it.
+    /// </summary>
+    /// <param name="source">Guest bytes at the candidate instruction. Only the first
+    /// <see cref="SequenceLength"/> bytes are inspected.</param>
+    /// <param name="replacement">Destination for the replacement encoding. Written only on a match, and
+    /// must be at least <see cref="SequenceLength"/> bytes.</param>
+    /// <returns><see langword="true"/> with <paramref name="replacement"/> filled when the idiom is
+    /// recognised; otherwise <see langword="false"/> with <paramref name="replacement"/> left alone.</returns>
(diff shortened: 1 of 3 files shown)


Spooks4576:

Tested `fix/sse4a-extrq-any-register` @ `ac9a6474` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-338

par274:

PR #375, opened 2026-07-17
[CPU] Generalize SSE4a extrq patch to any scratch register

### LINKED ISSUE #328: [CPU] Guest SSE4a EXTRQ raises SIGILL on any host without SSE4a (Dead Cells PPSA15552)

999sian:

## What this fixes

`DirectExecutionBackend.TryPatchSse4aExtrqBlend` rewrites the AMD-only `extrq` + `vpblendd` idiom the PS5 toolchain emits into equivalent SSE4.1, because SSE4a is not implemented on Intel hosts or under Rosetta and the raw opcode raises #UD there. As #328 reports, the matcher pinned the scratch register to `xmm2`:

```
66 0F 78 C2 28 00     extrq    xmm2, 0x28, 0
C4 E3 79 02 C2 02     vpblendd xmm0, xmm0, xmm2, 2
```

Dead Cells (PPSA15552) emits the same idiom against `xmm1` (`...C1...`), so the pattern missed by one byte per instruction, the `extrq` reached the CPU, and the game died with SIGILL right after the first frame. The `Patched ... 0 SSE4a EXTRQ blends` line in the startup log is the giveaway.

## The change

I read the register out of the ModRM byte instead of comparing it to a literal, so every register the toolchain uses (xmm0–xmm7) is handled. For register N it emits:

```
66 0F 3A 14 (C0|N<<3) 04   pextrb eax, xmmN, 4
66 0F 3A 22 C0 01          pinsrd xmm0, eax, 1
```

This is the same equivalence the original patch relied on: `extrq xmmN, 0x28, 0` keeps the low 40 bits of `xmmN` and zeroes the rest, so the dword the `vpblendd` folds into `xmm0` is byte 4 of the original `xmmN`, zero-extended. `pextrb`/`pinsrd` reproduce that using EAX as a throwaway.

- The extract field (`0x28` length, `0` index) is still matched literally, because the byte-4 offset in the replacement is only correct for that field. A different width would need different math, so it is safer not to match it.
- The destination stays `xmm0`, the blend mask stays `2`, and there is no REX prefix, so this does not widen what gets rewritten beyond the single idiom (xmm0–7) — it just stops caring which of those registers the toolchain picked.

I moved the match/encode decision into a small pure `Sse4aExtrqRewriter` so it can be tested directly; the unsafe `VirtualProtect`/write/`FlushInstructionCache` plumbing stays in the backend.

## Testing

- `Sse4aExtrqRewriterTests`: pins the `xmm2` case to the exact bytes the previous patch produced (so the refactor is byte-identical for the case that already shipped), covers `xmm1` and all eight registers, and rejects a mismatched extract field, a differing blend register, a non-`/0` reg field, unrelated bytes, and a short window.
- `dotnet build SharpEmu.slnx -c Release` is clean (0 warnings); `dotnet test SharpEmu.slnx -c Release` passes (249 in the Libs suite, +15 here; 33 in the source-generator suite).
- I disassembled the bytes to double-check: the inputs decode to `extrq xmm1, 0x28, 0` / `vpblendd xmm0, xmm0, xmm1, 2` and the outputs to `pextrb eax, xmm1, 4` / `pinsrd xmm0, eax, 1`.

I have not booted PPSA15552 past the fault myself, so verification here is at the unit-test, build, and instruction-encoding level.

## Checklist

Fixes #328.

### DIFF
--- a/src/SharpEmu.Core/Cpu/Emulation/Sse4aExtrqRewriter.cs
+++ b/src/SharpEmu.Core/Cpu/Emulation/Sse4aExtrqRewriter.cs
@@ -1,0 +1,96 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+namespace SharpEmu.Core.Cpu.Emulation;
+
+/// <summary>
+/// Rewrites the AMD SSE4a <c>extrq</c> idiom that Sony's compiler emits into an
+/// equivalent SSE4.1 sequence.
+///
+/// Guest PS5 code runs natively. The PS5's Zen 2 cores implement SSE4a, but Intel
+/// hosts and Rosetta 2 do not, so the raw <c>extrq</c> opcode raises #UD and the
+/// title aborts. The compiler emits a single fixed shape: extract the low 40 bits of
+/// a scratch register, then fold that register's second dword into <c>xmm0</c>.
+///
+/// <code>
+/// 66 0F 78 /0 28 00        extrq    xmmN, 0x28, 0
+/// C4 E3 79 02 /r 02        vpblendd xmm0, xmm0, xmmN, 2
+/// </code>
+///
+/// After <c>extrq xmmN, 0x28, 0</c> only the low 40 bits of <c>xmmN</c> survive, so
+/// the dword the blend takes (lane 1) is just byte 4 of the original register,
+/// zero-extended. <c>pextrb</c>/<c>pinsrd</c> reproduces exactly that, using EAX as a
+/// throwaway scratch:
+///
+/// <code>
+/// 66 0F 3A 14 /r 04        pextrb eax, xmmN, 4
+/// 66 0F 3A 22 C0 01        pinsrd xmm0, eax, 1
+/// </code>
+///
+/// The scratch register N is read out of the ModRM byte rather than compared against
+/// a literal, so every register the compiler picks (xmm0–xmm7) is covered. The
+/// extract field (length 0x28, index 0) and the xmm0 destination stay pinned: the
+/// byte-4 offset in the replacement is only correct for that field, so a different
+/// extract is deliberately left unmatched instead of rewritten wrongly.
+///
+/// The unsafe memory-protection and instruction-cache plumbing lives in the backend;
+/// this class only decides whether a byte window matches and what to write back, so
+/// the equivalence can be unit-tested in isolation.
+/// </summary>
+public static class Sse4aExtrqRewriter
+{
+    /// <summary>Length in bytes of both the matched idiom and its replacement.</summary>
+    public const int SequenceLength = 12;
+
+    /// <summary>
+    /// Matches the <c>extrq</c> + <c>vpblendd</c> idiom at the start of
+    /// <paramref name="source"/> and, when it matches, writes the SSE4.1 replacement
+    /// into <paramref name="replacement"/>.
+    /// </summary>
+    /// <param name="source">Instruction bytes at the faulting address.</param>
+    /// <param name="replacement">Receives the 12-byte replacement when the idiom matches.</param>
+    /// <returns><see langword="true"/> when the idiom matched and a replacement was produced.</returns>
+    public static bool TryGetReplacement(ReadOnlySpan<byte> source, Span<byte> replacement)
+    {
+        if (source.Length < SequenceLength || replacement.Length < SequenceLength)
+        {
+            return false;
+        }
+
+        // extrq xmmN, 0x28, 0 is the /0 form, so the ModRM byte is 0xC0|N (N = 0..7).
+        var modrm = source[3];
+        if (source[0] != 0x66 || source[1] != 0x0F || source[2] != 0x78 ||
+            (modrm & 0xF8) != 0xC0 || source[4] != 0x28 || source[5] != 0x00)
+        {
+            return false;
+        }
+
+        // vpblendd xmm0, xmm0, xmmN, 2: the same register, an xmm0 destination and a
+        // blend mask of 2 all fall out of requiring the ModRM byte to equal the extrq's.
+        if (source[6] != 0xC4 || source[7] != 0xE3 || source[8] != 0x79 || source[9] != 0x02 ||
+            source[10] != modrm || source[11] != 0x02)
+        {
+            return false;
+        }
+
+        var register = modrm & 0x07;
+
+        // pextrb eax, xmmN, 4: xmmN is the reg field, EAX (0) the r/m destination.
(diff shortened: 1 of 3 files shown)


Spooks4576:

Tested `fix/sse4a-extrq-scratch-register` @ `f51a4305` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-375

par274:
You have not boot Dead Cells, but fixed something? How?

999sian:
The crash in #328 is a single instruction, not game logic: `extrq` is AMD-only SSE4a and Intel CPUs don't implement it, so the fault is reproducible without the game. That's how I verified it.

On my Intel host (Core Ultra 7 155H, no SSE4a) I executed the exact bytes from the PPSA15552 fault address in an executable page:

```
original  extrq+vpblendd : SIGILL
original  extrq alone    : SIGILL
patched   pextrb+pinsrd  : executed
  xmm0 in       = ffffffffeeeeeeee ddddddddcccccccc
  xmm1 in       = 99aabbccddeeff00 1122334455667788
  xmm0 out      = ffffffffeeeeeeee 00000044cccccccc
  xmm0 expected = ffffffffeeeeeeee 00000044cccccccc (extrq+vpblendd semantics)
  equivalence   : MATCH
```

Same #UD the issue log shows, and the replacement leaves the exact register state the original pair produces on a CPU that has SSE4a (dword 1 of xmm0 = byte 4 of xmm1, zero-extended; everything else untouched). The unit tests also pin the xmm2 encoding byte-identical to what the patcher already ships, so the shipped case cannot regress.

What I don't have is a PPSA15552 dump, so I can't show the game booting past the fault myself. @chengboonrong reproduced this on Rosetta in #328 and offered to verify a fix; the macOS artifact from this PR's build is linked above if that verification is worth having before deciding.

PR #377, opened 2026-07-17
[CPU] Check BlockWakeHandler when a blocked thread has no BlockWaiter

999sian:

## What this fixes

A blocked guest thread's readiness predicate gets registered one of two ways: as an `IGuestThreadBlockWaiter` (`BlockWaiter`), or through the `resumeHandler`/`wakeHandler` compatibility bridge (`RequestCurrentThreadBlock`'s `Func<int>`/`Func<bool>` overload) that every `pthread_mutex_lock`, `pthread_cond_wait`/`timedwait`, `pthread_rwlock_wrlock`/`rdlock`, and `sceKernelWaitSema` wait uses. That bridge leaves `BlockWaiter` null and puts the readiness predicate in `BlockWakeHandler` instead — `RegisterBlockedGuestThreadContinuation`'s two overloads are mutually exclusive about which one gets set.

`WakeBlockedThreads` and `RunGuestThread`'s post-block exit-race check only tested `BlockWaiter`:

```csharp
if (thread.BlockWaiter is not null && !thread.BlockWaiter.TryWake())
{
    continue;
}
```

When `BlockWaiter` is null, this short-circuits to `false` and the thread is marked `Ready` unconditionally, without ever calling `BlockWakeHandler()`. Concretely: a thread blocked in `pthread_mutex_lock` wakes as soon as *any* thread touches the same `wakeKey` on unlock, whether or not `TryGrantBlockedMutexLock` actually granted it the lock — the guest can end up believing it holds a mutex it was never granted. The same gap applies to condition variables, rwlocks, and semaphores.

`RunGuestThread`'s resume path had the mirror problem: it calls `blockWaiter.Resume()` to compute the resumed RAX, but never called `BlockResumeHandler()`, so a thread woken through the handler bridge resumed with whatever RAX was left over from before it blocked instead of the actual completion result code (e.g. `CompleteBlockedMutexLock`'s granted/busy result).

Two other consumers in the same file already handle this correctly — `ResumeBlockedNestedGuestCallback` (checks `BlockWakeHandler`, calls the resume delegate) and `RestoreInterruptedGuestThread` (checks `BlockWakeHandler` before re-readying) — so this wasn't a design gap, just two call sites that predate the delegate-bridge path or were missed when it was added.

## The change

- Extracted the shared decision into `TryConsumeRegisteredWakeSignal(GuestThreadState)`: prefer `BlockWaiter.TryWake()` when a waiter is registered, fall back to `BlockWakeHandler()` when only that's set, and default to waking when neither is set (plain `wakeKey`-only waits like event flags/queues, which have no readiness predicate of their own).
- `WakeBlockedThreads` and the exit-race check in `RunGuestThread` both now call this helper instead of testing `BlockWaiter` directly.
- `RunGuestThread`'s resume path now captures `BlockResumeHandler` alongside `BlockWaiter` and calls it for the resumed RAX when there's no waiter.

## Testing

- New `GuestThreadWakeSignalTests` (7 tests) exercise `TryConsumeRegisteredWakeSignal` directly via reflection (the same pattern `ImportTrampolineAbiTests` already uses for this private nested type): waiter-only wake/no-wake, handler-only wake/no-wake (the regression this fixes), neither-set-always-wakes, and waiter-takes-precedence-when-both-are-set.
- `dotnet build SharpEmu.slnx -c Release` is clean (0 warnings). `dotnet test SharpEmu.slnx -c Release` passes (241 in the Libs suite on this branch's `main` base, +7 here; 33 in the source-generator suite).
- Smoke-tested the built CLI (`--help` and a bare invocation) to confirm the scheduler change doesn't break startup.

I have not reproduced this against a live game with contended pthread mutexes/condvars — I don't have a title that reliably exercises this path end to end. The fix is scoped to reading the existing `BlockWakeHandler`/`BlockResumeHandler` fields correctly rather than changing any locking semantics, and is unit-tested at the decision-function level, but I'd welcome confirmation from anyone who can reproduce a mutex-heavy title (e.g. the IL2CPP/Unity titles that hit `sceKernelWaitSema` heavily, like #254) that this changes observed behavior for the better.

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -3139,7 +3139,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 					continue;
 				}
 
-				if (thread.BlockWaiter is not null && !thread.BlockWaiter.TryWake())
+				if (!TryConsumeRegisteredWakeSignal(thread))
 				{
 					continue;
 				}
@@ -3168,6 +3168,24 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 		}
 
 		return wakeCount;
+	}
+
+	// A blocked thread's readiness predicate is registered either as an
+	// IGuestThreadBlockWaiter (BlockWaiter) or, for the delegate-pair
+	// compatibility bridge used by pthread mutex/cond/rwlock/semaphore waits
+	// (see RequestCurrentThreadBlock's resumeHandler/wakeHandler overload), as
+	// a bare BlockWakeHandler with BlockWaiter left null. Both must be checked
+	// here; skipping BlockWakeHandler would ready a mutex/cond waiter before
+	// its handle was actually granted. A thread with neither (a plain
+	// wakeKey wait, e.g. event flags/queues) always wakes on a key match.
+	private static bool TryConsumeRegisteredWakeSignal(GuestThreadState thread)
+	{
+		if (thread.BlockWaiter is { } waiter)
+		{
+			return waiter.TryWake();
+		}
+
+		return thread.BlockWakeHandler is not { } wakeHandler || wakeHandler();
 	}
 
 	public IReadOnlyList<GuestThreadSnapshot> SnapshotThreads()
@@ -4734,6 +4752,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 			LastError = null;
 			GuestCpuContinuation continuation = default;
 			IGuestThreadBlockWaiter? blockWaiter = null;
+			Func<int>? blockResumeHandler = null;
 			var resumeContinuation = false;
 			using (LockGate("RunGuestThread.block"))
 			{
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `fix/blocked-thread-wake-handler` @ `f5333b80` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-377

par274:
You fixed Unity deadlocks, but you have not test a real game.

999sian:
That's fair — I don't have a title that reliably exercises contended cooperative mutex/cond waits, and I said so in the description rather than claim otherwise.

The code-level claim doesn't need a game to check, for what it's worth: `RegisterBlockedGuestThreadContinuation` has two overloads that set either `BlockWaiter` or `BlockWakeHandler` (never both), and `WakeBlockedThreads` only tested `BlockWaiter` — so every pthread mutex/cond/rwlock/sema wait, which registers through the handler overload, woke without its readiness predicate ever being evaluated. The other two consumers in the same file already check `BlockWakeHandler`; these two call sites didn't.

If the bar is a before/after on a real game, the right test is probably someone with the #254 repro (Arcade Game Zone, IL2CPP threads deadlocked on sceKernelWaitSema). The branch stays up if anyone wants to run that against it.

PR #379, opened 2026-07-18
Fix/mem reserve and mutex exit cleanup

shadowbeat070:

## What this fixes

Three early-boot blockers that stopped SILENT HILL: The Short Message (PPSA10112) — an Unreal Engine PS5 title — from getting past process startup. Each one halted the guest with the stall watchdog (`Process exited with code 4`), one after the other.

**1. `sceKernelReserveVirtualRange` hung the loader at import #8.** The CRT reserves a 512 GiB virtual range early in libc init. `PhysicalVirtualMemory.TryAllocateAtExact` — the exact-address path the reservation search uses — always mapped with `MEM_RESERVE | MEM_COMMIT`, so it tried to * commit* 512 GiB. That far exceeds the host commit limit, the allocation failed, and the exact-address search loop ground through tens of thousands of doomed attempts until the watchdog fired:

```
[LOADER][TRACE] Import#8: libKernel:sceKernelReserveVirtualRange (7oxv3PPCumo)
[LOADER][ERROR] Execution stalled with no import progress for 20s (imports=256).
[LOADER][ERROR] Stall import-stub: ... nid=7oxv3PPCumo -> libKernel:sceKernelReserveVirtualRange
```

**2. A terminated guest thread could strand a mutex forever.** Nothing released the mutex ownership or dequeued the lock waiters a thread left behind when it exited — ownership was only ever cleared by an explicit unlock, and a waiter only removed when it was granted. Since the fast-acquire path refuses a mutex whose wait queue is non-empty, a stranded waiter wedged that mutex for every future locker. This is a latent correctness bug found while investigating the deadlock below; it is not the specific stall that hung this title, but it is the same class and worth closing.

**3. `pthread_cond_wait` failed to re-acquire a free mutex.** `pthread_cond_wait` releases the mutex, waits, and re-acquires it on wake. The signal path (`CompleteCondWaiterLocked`) moved the woken waitbut never granted it when the mutex was already free. POSIX permits signalling a condvar without holding its mutex, and Unreal's async-task queue does exactly that — so no later unlock hands the mutex over, the waiter is stranded on a free mutex, the woken thre a lock it does not, and its next unlock fails. That deadlocked the whole engine with the online task thread stuck at 94 imports. The per-mutex history (from the diagnostic below) shows a clean lock/unlock followed by an unlock of a now-free mutex that already

```
[LOADER][PTHREAD-DIAG] unlock-not-locked: ... owner=0x0 rec=0 waiters=1 sameObject=True
[LOADER][PTHREAD-DIAG]   #219311 lock   ... owner=T rec=1
[LOADER][PTHREAD-DIAG]   #219312 unlock ... owner=0 rec=0 waiters=0 result=0x00000000
[LOADER][PTHREAD-DIAG]   #219499 unlock ... owner=0 rec=0
```

## The change

**Reserve-only for large exact allocations** (`PhysicalVirtualMemory.TryAllocateAtExact`). For non-executable requests over the existing `LargeDataReserveThreshold` / `FullCommitRegionLimit`, reserve thd mark the region `IsReservedOnly` so pages commit on demand through the existing `EnsureRangeCommitted` fault path — the same thing the sibling `AllocateAt` already does. Small and executable allocations still commit up front, unchanged.

**Release synchronization state on thread exit.** Added ahreadExited` notification, raised from the worker-thread teardown in `DirectExecutionBackend` (which covers both a thread returning from its entry and `pthread_exit`, since both unwind to the same `Returned` point). `KernelPthreadCompatExports.ReleaseThreadSyncand, for the exiting thread, dequeues its mutex waiters, releases any mutex it still owned, and hands a now-free contended mutex to the next waiter. It fires after the scheduler gate is released, so it keeps the same lock ordering as `pthread_mutex_unlock`.

**Grant the cond-var re-acquire when the mutex is free.**`, after enqueuing the re-acquisition waiter, call `TryGrantMutexWaiterLocked`. That call is a no-op unless the mutex is free and this waiter is at the FIFO head, and it is idempotent with the wake-time grant, so held-mutex signalling is unaffected — it only cll time" hole.

- The reserve-only switch is gated on the same thresholdso it does not change small or executable allocations.
- The exit cleanup is scoped to mutexes. Condition-variable and semaphore waiters have the same "only removed on success" shape and are left as follow-ups rather than widened here.
- The cond-var grant does not change queue ordering or the held-mutex path; it only grants when the mutex is already free.

## Testing

Games: SILENT HILL: The Short Message (PPSA10112).

- New unit tests: `GuestMemoryAllocatorTests` (+2: a large exact request reserves instead of committing; a small one still commits), `PthreadMutexExitCleanupTests` (+3: an exiting waiter is dequeued anditing owner releases and hands off; an uncontended exiting
owner leaves the mutex free), `PthreadCondReacquireTests` (+2: a free mutex at signal time is granted to the waiter; a held mutex leaves the waiter queued).
- `dotnet test tests/SharpEmu.Libs.Tests` passes: 241 tests (+7 here).
- `dotnet build src/SharpEmu.CLI -c Release` is clean (0
- Verified against the real title, run to run: the reserve fix clears the import-#8 hang and lets all six modules load; the cond fix moves the online task thread from 94 imports (wedged in `pthread_mutehread_cond_timedwait`), and the `INVALID_ARGUMENT` unlock disappears from the log.

Honest scope: this does **not** boot the game. After these three fixes the title still stalls later — the engine spins up all its threads and then goes idle waiting on the network/NP (HTTP) subsystem, wks in a host call and never returns. That is a separate issue, not addressed here.

I also added an address-independent mutex diagnostic behind `SHARPEMU_LOG_PTHREAD_MUTEX_DIAG=1` — it dumps a mutex's recent operation history when an unlock hits an unexpected error, which is what pinneLR.

## Checklist

- [ ] I wrote this pull request description myself and dianations.

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -4765,6 +4765,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 			var exitReason = resumeContinuation
 				? ExecuteBlockedGuestThreadContinuation(thread.Context, continuation, thread.Name, out var blockReason)
 				: ExecuteGuestThreadEntry(thread.Context, thread.EntryPoint, thread.Name, out blockReason);
+			var guestThreadExited = false;
 			using (LockGate("RunGuestThread.exit"))
 			{
 				switch (exitReason)
@@ -4772,6 +4773,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 					case GuestNativeCallExitReason.Returned:
 						thread.ExitValue = thread.Context[CpuRegister.Rax];
 						thread.State = GuestThreadRunState.Exited;
+						guestThreadExited = true;
 						if (_logGuestThreads)
 						Console.Error.WriteLine(
 							$"[LOADER][INFO] Guest thread exited: name='{thread.Name}' " +
@@ -4798,6 +4800,15 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 						thread.BlockReason = blockReason;
 						break;
 				}
+			}
+
+			// Release any mutex ownership or queued lock waiters the thread left
+			// behind, now that the gate is released (the cleanup takes per-mutex
+			// locks and may wake blocked threads, mirroring pthread_mutex_unlock).
+			// A stranded owner or waiter would otherwise wedge that mutex forever.
+			if (guestThreadExited)
+			{
+				GuestThreadExecution.NotifyGuestThreadExited(thread.ThreadHandle);
 			}
 			if (_logGuestThreads)
 			{
--- a/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
+++ b/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
@@ -162,7 +162,14 @@ public sealed unsafe class PhysicalVirtualMemory : IVirtualMemory, IGuestMemoryA
         var alignedSize = (size + 0xFFF) & ~0xFFFUL;
         var protection = executable ? PAGE_EXECUTE_READWRITE : PAGE_READWRITE;
         var hostProtection = executable ? HostPageProtection.ReadWriteExecute : HostPageProtection.ReadWrite;
-        var result = _hostMemory.Allocate(desiredAddress, alignedSize, hostProtection);
+
+        var reserveOnly = !executable &&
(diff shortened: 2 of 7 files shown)


Spooks4576:

Tested `fix/mem-reserve-and-mutex-exit-cleanup` @ `9fc77724` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-379

par274:
Please use your own words when filling out the PR template.

PR #347, opened 2026-07-17
[AvPlayer] Resolve guest media within app0

peter-bonanni:
## What changed

Resolves AvPlayer media paths inside the mounted `app0` filesystem, including project-relative file URIs and unambiguous case-insensitive matches. Host paths, traversal, remote URIs, and symlink escapes are rejected.

## Why

Guest media references were treated as host paths, which prevented valid application-relative files from resolving and could allow paths outside the mounted title.

## Validation

- 7 focused AvPlayer path tests passed with temporary synthetic files.
- The full build and test suite passed.

## Testing

- Grand Theft Auto V (PS5) – Exercised on the combined integration branch containing this change.
- Automated, focused, and local validation: Retained in the original description below.

## Checklist

By submitting this pull request, you confirm that:

- [x] I have read and followed `CONTRIBUTING.md`.
- [x] I tested my changes or marked the testing section as `N/A`.
- [x] I wrote this pull request description myself and did not paste AI-generated explanations.
- [x] I listed the game(s) I tested (or marked the testing section as `N/A`).

PR #381, opened 2026-07-18
[Kernel] Match path cache comparisons to host filesystem case sensitivity

999sian:

## What this fixes

The negative-stat cache and the apr file-size cache in `KernelMemoryCompatExports` memoize the outcome of host filesystem probes, but both are keyed with `StringComparer.OrdinalIgnoreCase` while the probes themselves (`File.Exists` / `Directory.Exists` / `FileInfo`) are case-sensitive on Linux. A cache using a looser equivalence than the thing it memoizes serves wrong answers:

- `stat("/app0/DATA.BIN")` fails, the miss is cached, and a later `stat("/app0/Data.bin")` gets `NOT_FOUND` straight from the cache without ever touching the disk — even though that file exists and a real probe would succeed. From then on the game is permanently told an existing file is missing.
- `sceKernelAprResolveFilepathsToIdsAndFileSizes` returns the cached size of a case-distinct sibling file instead of the file's own size.

The registered-mount containment guard has the inverse problem: its ignore-case `StartsWith` accepts a `..` path that resolves into a sibling directory differing from the mount root only by case (`…/Save` vs `…/save`), so guest I/O can escape the mount on a case-sensitive host.

## The change

All three sites now compare with the host filesystem's own semantics: ordinal-ignore-case on Windows, ordinal on Linux/macOS. Windows behavior is unchanged (NTFS lookups are case-insensitive, so the ignore-case keys were correct there). This is the same host-case-sensitivity class as #218 and the app0 handling in #347.

- `_negativeStatCache` and `_aprFileSizeCache` comparers are host-dependent.
- The mount containment check in `TryResolveRegisteredGuestMount` uses the same comparison.

An ordinal cache on a case-insensitive host filesystem is strictly conservative — a differently-cased repeat lookup just re-probes instead of hitting the cache, it can never produce a wrong answer.

## Testing

Three new tests in `KernelPathCaseSensitivityTests` drive the real exports (`PosixStat`, `KernelAprResolveFilepathsToIdsAndFileSizes`, `ResolveGuestPath`) against real temp files on the actual host filesystem:

- a cached miss for one casing must not shadow the existing differently-cased file,
- two case-distinct host files must each report their own size,
- a `../save/…` path must not escape a `Save` mount root through the case-insensitive prefix match.

All three fail on current `main` and pass with this change. The case-specific sections detect the host filesystem's actual sensitivity at runtime and skip on case-insensitive hosts, so the suite stays green on Windows and macOS. Full solution: clean build, 268+33 tests pass.

This is host-side filesystem cache behavior, demonstrated end-to-end against the real filesystem; I did not reproduce it inside a specific title.

## Checklist

- [x] I tested my changes. (unit tests against the real host filesystem on Linux; red on main, green with the fix)

Testing games: `N/A` — host filesystem cache behavior, verified against the real filesystem as described above.

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
@@ -110,8 +110,18 @@ public static partial class KernelMemoryCompatExports
     private static readonly Dictionary<ulong, string> _mappedRegionNames = new();
     private static readonly Dictionary<string, string> _guestMounts = new(StringComparer.OrdinalIgnoreCase);
     private static readonly HashSet<string> _tracedStatResults = new(StringComparer.Ordinal);
-    private static readonly HashSet<string> _negativeStatCache = new(StringComparer.OrdinalIgnoreCase);
-    private static readonly ConcurrentDictionary<string, ulong> _aprFileSizeCache = new(StringComparer.OrdinalIgnoreCase);
+    // Both caches memoize host filesystem probe outcomes, so their key
+    // equivalence must match the host filesystem's: Windows resolves names
+    // case-insensitively, but Linux hosts are case-sensitive, and an
+    // ignore-case cache there aliases distinct paths — a cached miss for
+    // "/app0/DATA.BIN" keeps answering NOT_FOUND for "/app0/Data.bin" even
+    // though that file exists and a fresh probe would find it.
+    private static readonly StringComparer HostFsPathComparer =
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `fix/path-cache-case-sensitivity` @ `1fa04d64` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-381

par274:
Good work.

PR #382, opened 2026-07-18
Fix PS5 memory and audio startup

PaN1C27:
## What changed

- support exact sub-allocation mappings inside Windows allocation-granularity reservations
- add PS5 AJM codec-range and memory-registration compatibility
- add the PS5 ACM context lifecycle required during audio startup
- fill missing AudioOut2 initialization exports and correct its port-create ABI
- write both queue-level outputs as 32-bit values to avoid corrupting adjacent guest stack data
- add focused memory, AJM, ACM, and AudioOut2 regression coverage

## Why

PS5 titles could fail to map exact guest ranges on Windows, then abort audio initialization on missing or mismatched AJM, ACM, and AudioOut2 contracts. One AudioOut2 queue-level shim also wrote an eight-byte value into a four-byte guest output, overwriting the caller's stack canary.

These changes allow the tested PS5 title to complete memory and audio startup, create its rendering threads, and keep its game window running instead of exiting on an unresolved import, access violation, or stack-check failure.

## Validation

- `dotnet test SharpEmu.slnx -c Release --no-restore` — 329 tests passed
- `dotnet build src/SharpEmu.CLI/SharpEmu.CLI.csproj -c Release -r win-x64 --no-restore` — 0 warnings, 0 errors
- runtime validation with PPSA26344 reached an active Vulkan game window without unresolved imports, native exceptions, or stack-check failure


PaN1C27:
Closing because this branch was intended only for the fork, not as an upstream pull request.

PR #390, opened 2026-07-18
Reject undefined numeric LogLevel values

kadu04t:
## Description

Found a small bug in TryParseLevel: Enum.TryParse accepts any numeric literal valid for the underlying type, even if it's not a defined member of the enum. So a value like "999" would parse successfully into (LogLevel)999, which isn't a real LogLevel. This meant SHARPEMU_LOG_LEVEL could end up with an invalid LogLevel, silently disabling log filtering.

I added && Enum.IsDefined(level) to the existing Enum.TryParse check. Since the if condition can now be false even when TryParse itself succeeds, I also reset level back to default in that case, so it doesn't keep the invalid 999 value when the method returns false.

## Testing

I didn't run any games because this doesn't affect gameplay at all; it only changes how the program reads the log level, i tested it by running the automated tests, and all 296 passed, including the new ones I wrote for the bug.

## Checklist

### DIFF
--- a/src/SharpEmu.Logging/SharpEmuLog.cs
+++ b/src/SharpEmu.Logging/SharpEmuLog.cs
@@ -110,7 +110,8 @@ public static class SharpEmuLog
         }
 
         var normalized = text.Trim();
-        if (Enum.TryParse<LogLevel>(normalized, ignoreCase: true, out level))
+        if (Enum.TryParse<LogLevel>(normalized, ignoreCase: true, out level) &&
+            Enum.IsDefined(level))
         {
             return true;
         }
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `origin/log-level-numeric-validation` @ `4ad25085` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-390

PR #391, opened 2026-07-18
Fix/mem reserve and mutex exit cleanup

shadowbeat070:

## What this fixes

**1. `PhysicalVirtualMemory.TryAllocateAtExact` unconditionally used MEM_RESERVE|MEM_COMMIT.** When a guest calls sceKernelReserveVirtualRange to carve out hundreds of GiB of address space, that tried to commit hundreds of GiB — far beyond what the host can back — so every call failed. That far exceeds the host commit limit, the allocation failed, and the exact-address search loop ground through tens of thousands of doomed attempts until the watchdog fired:

Fix: I essentially copied the fix that 'AllocateAt' already had which was is to use MEM_RESERVE only for very large non-executable requests.

**2. A guest thread that exited or waited in it's queue would strand any mutex it was holding permanently.** Ownership of a mutex would only be cleared by an explicit unlock and waiter was only removed if it was granted the lock. If the thread dies then none of this happens.

Fix: created a guestThreadExited notifcation that's raised on worker thread teardown. KernelPthreadCompatExports now subscribes, dequeues the dead thread waiters, releases any mutexes it owned and hand the free mutex to the next waiter. Explicitly scoped to mutexes.

**3. `pthread_cond_wait` failed to re-acquire a free mutex.** `pthread_cond_wait` releases the mutex, waits, and re-acquires it on wake. The signal path (`CompleteCondWaiterLocked`) moved the woken wait but never granted it when the mutex was already free. POSIX permits signalling a condvar without holding its mutex, and Unreal's async-task queue does exactly that — so no later unlock hands the mutex over, the waiter is stranded on a free mutex, and its next unlock fails.

Fix: Grant the re-acquisition immediately at signal time, if the mutex is free right then.

## The change

**Reserve-only for large exact allocations** (`PhysicalVirtualMemory.TryAllocateAtExact`). For non-executable requests reserve the region `IsReservedOnly` so pages commit on demand through the existing `EnsureRangeCommitted` fault path — the same thing the sibling `AllocateAt` already does. Small and executable allocations still commit up front, unchanged.

**Release synchronization state on thread exit.** Added a `GuestThreadExited` notification, raised from the worker-thread teardown in `DirectExecutionBackend` (which covers both a thread returning from its entry and `pthread_exit`, since both unwind to the same `Returned` point). `KernelPthreadCompatExports.ReleaseThreadSynchronizationState`, for the exiting thread, dequeues its mutex waiters, releases any mutex it still owned, and hands a now-free contended mutex to the next waiter. It fires after the scheduler gate is released, so it keeps the same lock ordering as `pthread_mutex_unlock`.

**Grant the cond-var re-acquire when the mutex is free.**, after enqueuing the re-acquisition waiter, call `TryGrantMutexWaiterLocked`. That call is a no-op unless the mutex is free and this waiter is at the FIFO head, and it is idempotent with the wake-time grant, so held-mutex signalling is unaffected — it only cll time" hole.

- The reserve-only switch is gated on the same thresholdso it does not change small or executable allocations.
- The exit cleanup is scoped to mutexes. Condition-variable and semaphore waiters have the same "only removed on success" shape and are left as follow-ups rather than widened here.
- The cond-var grant does not change queue ordering or the held-mutex path; it only grants when the mutex is already free.

## Testing

Games: SILENT HILL: The Short Message (PPSA10112).

- New unit tests: `GuestMemoryAllocatorTests` (+2: a large exact request reserves instead of committing; a small one still commits), `PthreadMutexExitCleanupTests` (+3: an exiting waiter is dequeued anditing owner releases and hands off; an uncontended exiting
owner leaves the mutex free), `PthreadCondReacquireTests` (+2: a free mutex at signal time is granted to the waiter; a held mutex leaves the waiter queued).
- `dotnet test tests/SharpEmu.Libs.Tests` passes: 241 tests (+7 here).
- `dotnet build src/SharpEmu.CLI -c Release` is clean (0

I also added an address-independent mutex diagnostic behind `SHARPEMU_LOG_PTHREAD_MUTEX_DIAG=1` — it dumps a mutex's recent operation history when an unlock hits an unexpected error, which is what pinneLR.

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -4765,6 +4765,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 			var exitReason = resumeContinuation
 				? ExecuteBlockedGuestThreadContinuation(thread.Context, continuation, thread.Name, out var blockReason)
 				: ExecuteGuestThreadEntry(thread.Context, thread.EntryPoint, thread.Name, out blockReason);
+			var guestThreadExited = false;
 			using (LockGate("RunGuestThread.exit"))
 			{
 				switch (exitReason)
@@ -4772,6 +4773,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 					case GuestNativeCallExitReason.Returned:
 						thread.ExitValue = thread.Context[CpuRegister.Rax];
 						thread.State = GuestThreadRunState.Exited;
+						guestThreadExited = true;
 						if (_logGuestThreads)
 						Console.Error.WriteLine(
 							$"[LOADER][INFO] Guest thread exited: name='{thread.Name}' " +
@@ -4798,6 +4800,15 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 						thread.BlockReason = blockReason;
 						break;
 				}
+			}
+
+			// Release any mutex ownership or queued lock waiters the thread left
+			// behind, now that the gate is released (the cleanup takes per-mutex
+			// locks and may wake blocked threads, mirroring pthread_mutex_unlock).
+			// A stranded owner or waiter would otherwise wedge that mutex forever.
+			if (guestThreadExited)
+			{
+				GuestThreadExecution.NotifyGuestThreadExited(thread.ThreadHandle);
 			}
 			if (_logGuestThreads)
 			{
--- a/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
+++ b/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
@@ -24,8 +24,7 @@ public sealed unsafe class PhysicalVirtualMemory : IVirtualMemory, IGuestMemoryA
     private const ulong GuestAllocationArenaAddress = 0x00006000_0000_0000;
     private const ulong GuestAllocationArenaSize = 0x0100_0000;
     private const ulong GuestAllocationArenaStartOffset = PageSize;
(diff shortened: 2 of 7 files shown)


Spooks4576:

Tested `fix/mem-reserve-and-mutex-exit-cleanup` @ `7066d043` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-391

Spooks4576:
This does not get Silent Hill to boot successfully 
I would like verification on that claim as im getting a pthread issue

 This should not be merged until verifying your claim? the mutex issues dont appear which is true but id like more clarification if this is supposed to fix the game or if you only fixed it partially 

```pthread_lock:
  mutex=0x000002341F3E4520
  current=0x00000233D5BBD0C0
  result=0x00000000

pthread_cond_wait-enter:
  cond=0x000002341F3E4528
  mutex=0x000002341F3E4520
  waiters=1
  timed=True
  result=0x00000000

pthread_unlock:
  mutex=0x000002341F3E4520
  result=0x00000000

pthread_cond_wait-block:
  cond=0x000002341F3E4528
  mutex=0x000002341F3E4520
  waiters=1
  timed=True
  result=0x00000000```

shadowbeat070:
Ah no, this absolutely does not make the game boot. I apologize if i was unclear. This currently is still only a partial fix and i am currently investigating further issues. Would it be better if i kept this pr as a draft until i can get the game to boot?

par274:
"Three early-boot blockers that stopped SILENT HILL: The Short Message (PPSA10112) that stopped the game from starting up. each issue caused the emulatos to crash/stop with exit code 4."

Need evidence for real game.

PR #392, opened 2026-07-18
[CPU] Patch SSE4a EXTRQ blend for any source xmm register

sullivanlb:
TryPatchSse4aExtrqBlend only recognized the extrq+vpblendd sequence when the compiler picked xmm2 as the source register. I've added compatibilty for xmm0 to xmm7.

Fixes #328

## Before submitting

Please read our contribution guidelines before opening a pull request:

➡️ [**CONTRIBUTING.md**](https://github.com/sharpemu/sharpemu/blob/main/CONTRIBUTING.md)

By opening this pull request, you confirm that you have read and agree to follow the contribution guidelines.

## Testing

No game reproduction, this bug only triggers on hosts without SSE4a. My dev machine is AMD (Zen 3), which supports SSE4a natively, so the original SIGILL never occurs there even with the affected instruction unpatched.

Instead, I verified directly against the byte-pattern matcher itself using the exact instruction bytes from the issue:

- Confirmed the bug: the current code matches the `xmm2` pattern but silently fails to match the `xmm1` pattern PPSA15552 emits
- Confirmed the fix: all 8 possible source registers (xmm0-xmm7) now match and patch correctly
- Confirmed no regression: the original `xmm2` produce same output as before
- Confirmed error case: a sequence with mismatched registers between `extrq` and `vpblendd` is correctly rejected
- Added a permanent unit test (`Sse4aExtrqBlendPatchTests.cs`) covering all of the above, plus the full project test suite (308 tests) passes

## Checklist

By submitting this pull request, you confirm that:

- [x] I have read and followed `CONTRIBUTING.md`.
- [x] I tested my changes or marked the testing section as `N/A`.
- [x] I wrote this pull request description myself and did not paste AI-generated explanations.
- [x] I listed the game(s) I tested (or marked the testing section as `N/A`).

PR #399, opened 2026-07-18
[Kernel/Memory] Fix sceKernelReserveVirtualRange stall and pthread mutex/cond-var synchronization bugs

ftre120:
## Before submitting

## What this fixes

Two related boot/synchronization bugs, both surfaced while investigating why Unreal Engine PS5 titles stall deep in process init:

**1. `sceKernelReserveVirtualRange` could stall the loader forever.** The CRT reserves a huge (e.g. 512 GiB) virtual range early in libc init. `PhysicalVirtualMemory.TryAllocateAtExact` always committed the full request up front (`MEM_RESERVE|MEM_COMMIT`), which fails past the host's commit limit — the exact-address search loop then retried at different addresses forever until the stall watchdog killed the process.

**2. A guest thread that exits while owning or waiting on a mutex stranded it forever**, and **a condition-variable waiter could resume believing it held a mutex it was never granted.** Both come from the same shape of bug: a mutex waiter is only removed when granted the lock, and ownership is only cleared by an explicit unlock — neither happens if the owning thread just dies, or if the mutex was already free when a condvar signalled it (which POSIX permits, and which Unreal's async task queue does).

## Credit

The diagnosis and original fix for these are [shadowbeat070](https://github.com/shadowbeat070)'s, from #379, found while getting SILENT HILL: The Short Message running. That PR was closed for not following the AI-assisted contribution guidelines, not because the fix was wrong. I read through the full diff, understood each of the three changes (memory reserve, thread-exit mutex cleanup, cond-var re-acquire), and independently re-verified them against a different title before resubmitting.

## The change

**Reserve-only for large exact allocations** (`PhysicalVirtualMemory.TryAllocateAtExact`): for non-executable requests over the existing large-allocation threshold, reserve the address range only and mark it `IsReservedOnly`, so pages commit on demand through the existing `EnsureRangeCommitted` fault path — mirroring what the sibling `AllocateAt` already does. Small and executable allocations are unchanged.

**Release synchronization state on thread exit**: `GuestThreadExecution` raises a `GuestThreadExited` event from `DirectExecutionBackend`'s worker-thread teardown. `KernelPthreadCompatExports.ReleaseThreadSynchronizationState` dequeues the exiting thread's mutex waiters, releases any mutex it still owned, and hands a now-free contended mutex to the next waiter.

**Grant the cond-var re-acquire when the mutex is free**: `CompleteCondWaiterLocked` now calls `TryGrantMutexWaiterLocked` right after enqueueing the re-acquisition waiter. That call is a no-op unless the mutex is free and the waiter is at the FIFO head, so the held-mutex handoff on unlock is unchanged.

## Testing

- The Invincible (PPSA06426): before this change, boot stalls with `sceKernelReserveVirtualRange` returning `ORBIS_GEN2_ERROR_PERMISSION_DENIED`/stalling and repeated `ORBIS_GEN2_ERROR_DEADLOCK` around thread pool teardown; after this change, both are gone and the loader proceeds well past process init into asset streaming and rendering. (Note: the title still doesn't reach gameplay — it hits a separate, unrelated slowdown further in, not addressed by this PR.)
- `dotnet test tests/SharpEmu.Libs.Tests -c Release`: 246 passed (7 new: `PthreadMutexExitCleanupTests` x3, `PthreadCondReacquireTests` x2, `GuestMemoryAllocatorTests` x2).
- `dotnet build sharpemu.sln -c Release`: 0 warnings, 0 errors.

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -2493,30 +2493,58 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 		Console.Error.WriteLine($"[LOADER][INFO] Patched {num3} TLS loads, {num9} TLS stores, {num4} stack-canary accesses, {sse4aPatchCount} SSE4a EXTRQ blends");
 	}
 
-	private unsafe bool TryPatchSse4aExtrqBlend(nint address, byte* source)
+	// Neither Rosetta nor non-AMD host CPUs implement SSE4a EXTRQ. Compilers
+	// emit this exact two-instruction idiom to mask a scratch xmm register to
+	// its low 40 bits and blend the resulting second dword into xmm0; the
+	// scratch register varies by call site (whichever xmm happens to be
+	// free), so it is decoded from the ModRM byte rather than hardcoded.
+	// PEXTRB/PINSRD provides the same observable result in 12 bytes: extract
+	// source byte 4 into a scratch GPR and insert it into xmm0's lane 1.
+	internal static bool TryComputeSse4aExtrqBlendReplacement(ReadOnlySpan<byte> source, Span<byte> replacement)
 	{
-		// Rosetta does not implement AMD SSE4a EXTRQ. This exact sequence masks
-		// xmm2 to its low 40 bits, then copies the resulting second dword into
-		// xmm0. PEXTRB/PINSRD provides the same observable result in 12 bytes:
-		// extract source byte 4 and insert the zero-extended value into lane 1.
-		ReadOnlySpan<byte> pattern =
-		[
-			0x66, 0x0F, 0x78, 0xC2, 0x28, 0x00,
-			0xC4, 0xE3, 0x79, 0x02, 0xC2, 0x02,
-		];
-		for (var i = 0; i < pattern.Length; i++)
+		if (source.Length < 12 || replacement.Length < 12)
 		{
-			if (source[i] != pattern[i])
-			{
-				return false;
-			}
+			return false;
 		}
 
-		ReadOnlySpan<byte> replacement =
-		[
-			0x66, 0x0F, 0x3A, 0x14, 0xD0, 0x04,
-			0x66, 0x0F, 0x3A, 0x22, 0xC0, 0x01,
-		];
+		// 66 0F 78 /0 28 00   EXTRQ   xmmR, 0x28, 0x00
+		// C4 E3 79 02 /r 02   VBLENDD xmm0, xmm0, xmmR, 0x02
+		// Both ModRM bytes are register-direct (mod=11) with the scratch
+		// register R in the low 3 bits, so they must agree on R.
+		if (source[0] != 0x66 || source[1] != 0x0F || source[2] != 0x78 ||
+			source[4] != 0x28 || source[5] != 0x00 ||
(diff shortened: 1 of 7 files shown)


Spooks4576:

Tested `fix/reserve-virtual-range-and-mutex-exit-cleanup` @ `0c8fe051` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-399

kostyaff (approved):
Three well-scoped fixes targeting real boot/sync bugs, with strong testing evidence (The Invincible past process init, Silent Hill deadlock in UE init). Approving with comments.

## Per-fix assessment

**Fix 1 — `sceKernelReserveVirtualRange` stall (PhysicalVirtualMemory.cs)**
Correct: large non-exec reservations (≥1 GiB AND >4 GiB) use `Reserve` instead of `Allocate`, avoiding the `MEM_COMMIT` failure that drove the exact-address search loop into the stall watchdog. Constants `LargeDataReserveThreshold`/`FullCommitRegionLimit` already exist in the file. `CommitLimitedHostMemory` fake is a clean testing approach. Two tests cover the large-reserve and small-commit paths.

**Fix 2 — Thread-exit cleanup (DirectExecutionBackend + GuestThreadExecution + PthreadCompat)**
`GuestThreadExited` event + `NotifyGuestThreadExited` invoked after the `RunGuestThread.exit` gate releases (correct — cleanup takes per-mutex locks and may wake blocked threads). `ReleaseThreadSynchronizationState` dedupes by `ReferenceEqualityComparer` on the state object, drops owner + dequeues waiters, hands mutex to the new head. `ConcurrentDictionary.Values` iteration + per-state `lock(state)` is race-safe with concurrent unlock paths. Three tests cover exiting waiter / exiting owner / uncontended owner.

**Fix 3 — Cond-var reacquire (PthreadCompat:1715)**
`TryGrantMutexWaiterLocked` call after enqueueing on the mutex wait queue — if the mutex is already free, grant immediately instead of stranding the waiter. Matches POSIX signal-without-lock semantics. Two tests cover free-mutex-grants and held-mutex-leaves-queued.

## Concerns / suggestions

1. **Overlap with PR #392 (SSE4a EXTRQ blend).** Both this PR and #392 modify `TryPatchSse4aExtrqBlend` in `DirectExecutionBackend.cs` to support any source xmm register (fix #328). They will conflict on merge. #399 adds `TryComputeSse4aExtrqBlendReplacement` (instance method, returns replacement bytes); #392 adds `TryMatchSse4aExtrqBlend` (out-param). Recommend the maintainer pick one — #399 is currently preferable since #392 has a red `Build linux-x64` check, but either approach is fine once reconciled.

2. **`_diagPthreadMutex` ring buffer (512 records).** Allocations per op only when `SHARPEMU_LOG_PTHREAD_MUTEX_DIAG=1` (default off) — zero overhead in normal runs. Fine for a debug aid.

3. **Reflection in tests (`PthreadCondReacquireTests` / `PthreadMutexExitCleanupTests`).** Brittle to renames of nested types (`PthreadMutexState`, `PthreadCondWaiter`, etc.), but necessary for white-box testing of private sync primitives. An `InternalsVisibleTo` + internal test-helpers approach (as done elsewhere in the codebase) would be less brittle, but not blocking.

Overall: strong PR, real game progress, good test coverage of the edge cases. Approving.

PR #406, opened 2026-07-18
Fix Massive Bug Preventing UE5 Titles From Booting

Spooks4576:
Fixes a Memcpy bug that effects all ue5 titles this specific fix fixes a memcpy corruption that would happen under the old memcpy implementation this fixes that

### DIFF
--- a/src/SharpEmu.CLI/packages.lock.json
+++ b/src/SharpEmu.CLI/packages.lock.json
@@ -4,9 +4,9 @@
     "net10.0": {
       "Microsoft.NET.ILLink.Tasks": {
         "type": "Direct",
-        "requested": "[10.0.3, )",
-        "resolved": "10.0.3",
-        "contentHash": "0B6nZyCHWXnvmlB559oduOspVdNOnpNXPjhpWVMovLPAsDVG7A4jJR9rzECf67JUzxP8/ee/wA8clwIzJcWNFA=="
+        "requested": "[10.0.8, )",
+        "resolved": "10.0.8",
+        "contentHash": "dVbSXGIFNR5nZcv2tOLoWI+a9T4jtFd77IYjuND+QVe360qWgAF7H0WtoopYhRw/+SgpGUTyrkrh+65+ClNnfw=="
       },
       "Avalonia.Angle.Windows.Natives": {
         "type": "Transitive",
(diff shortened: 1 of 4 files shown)


Spooks4576:
image
image

Spooks4576:

Tested `main` @ `ea62611b` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-406

kostyaff:
## Build osx-x64 FAILURE — test bug, not a fix bug

`Build osx-x64` is red because of one test:

```
SharpEmu.Libs.Tests.Cpu.MemcpyHleRoutingTests.TryCreateNativeImportIntrinsic_DoesNotClaimMemcpy [FAIL]
---- System.OutOfMemoryException : Failed to allocate Win64->SysV thunk page
```

### Root cause
`TryCreateNativeImportIntrinsic_DoesNotClaimMemcpy` invokes `TryCreateNativeImportIntrinsic` via reflection on a `RuntimeHelpers.GetUninitializedObject`-created `DirectExecutionBackend`. On macOS that path tries to allocate a Win64→SysV thunk page (Windows-specific memory API) and throws `OutOfMemoryException` **before** the `IsHlePreferredNid` guard at the top of `TryCreateNativeImportIntrinsic` is reached. Ironically, the test is asserting that the intrinsic does *not* claim memcpy — but the intrinsic path crashes on an unrelated Windows-only codepath before it can decline.

### Fix bug itself looks correct
- Adding `Q3VBxCXhUHs` to `IsHlePreferredNid` (unconditional, no OS branch) — right call, the old OS gating left Linux/macOS on the LLE stub.
- `ctx[CpuRegister.Rax] = destination` in `KernelMemoryCompatExports.Memcpy` — semantically correct (memcpy returns dest).
- The IL-scan test (`IsHlePreferredNid_DoesNotBranchOnHostOperatingSystem`) is a nice guard against reintroducing the OS condition.

### Suggested fix for the failing test
Guard the test with `[PlatformSpecificTest` / `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)`, or skip the `TryCreateNativeImportIntrinsic` call on non-Windows (the `IsHlePreferredNid`-only assertions already pass on macOS — the `DoesNotClaimMemcpy` one is the only OS-dependent case). The four other tests in the file are green on macOS.

Not blocking on the fix logic — just the test harness needs a platform guard.

PR #411, opened 2026-07-18
CPU: run native guest workers on Linux

999sian:
## Summary

The pooled raw-thread guest executor (`NativeGuestExecutor`) previously ran only on Windows; every other host fell back to inline guest execution because the run loop was emitted for the Win32 wait/signal/exit API and created with `CreateThread`. This enables the same dedicated raw-thread execution on Linux:

- The emitted run loop's wait/signal calls are bound on Linux to Win64-ABI thunks over the existing POSIX worker-event semaphores, and its stop path returns cleanly from the pthread start routine instead of calling `ExitThread`.
- The worker thread is created through the existing stack-sized `PosixHostStubs.CreateWorkerThread` helper.
- Teardown confirms the worker reached its stop path via the done-event it already signals, then reaps it with a blocking `pthread_join` (new `JoinWorkerThread`). A timed `pthread_kill(tid, 0)` probe is not used: on current glibc it reports a returned-but-unjoined thread as live indefinitely, whereas `pthread_join` reaps it immediately. If the worker never signals within the timeout it is left detached and its run loop leaked, matching the existing Windows wedged-worker behavior.

Windows behavior is unchanged. Unsupported hosts and `SHARPEMU_DISABLE_NATIVE_GUEST_WORKERS=1` still return the inline fallback, and every initialization-failure path frees its resources without signalling an uninitialized semaphore.

## Testing

- `dotnet build SharpEmu.slnx -c Release`
- `dotnet test tests/SharpEmu.Libs.Tests/SharpEmu.Libs.Tests.csproj -c Release` — 266 passed, including a new Linux test that drives the done-event handshake and blocking join used by teardown.
- Real-game testing was not performed.

## Checklist

- [x] I read and followed `CONTRIBUTING.md`.
- [x] I tested the change as described above.
- [x] I wrote the pull request description myself and did not paste AI-generated explanations.
- [x] I listed game(s) I tested (or marked the testing section as `N/A`).

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.NativeWorker.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.NativeWorker.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using System.Threading;
 using SharpEmu.HLE;
+using HlePosixHostStubs = SharpEmu.HLE.Host.Posix.PosixHostStubs;
 
 namespace SharpEmu.Core.Cpu.Native;
 
@@ -92,10 +93,8 @@ public sealed partial class DirectExecutionBackend
 
 	private NativeGuestExecutor? RentNativeGuestExecutor()
 	{
-		// NativeGuestExecutor emits a Win32 wait loop and creates it with
-		// kernel32!CreateThread. POSIX hosts use the established inline entry
-		// path until the worker loop has a pthread/eventfd implementation.
-		if (!OperatingSystem.IsWindows() || NativeGuestWorkersDisabled)
+		if ((!OperatingSystem.IsWindows() && !OperatingSystem.IsLinux()) ||
+			NativeGuestWorkersDisabled)
 		{
 			return null;
 		}
@@ -186,6 +185,8 @@ public sealed partial class DirectExecutionBackend
 		private static nint _waitForSingleObjectAddress;
 		private static nint _setEventAddress;
 		private static nint _exitThreadAddress;
+		private static nint _posixWaitWorkerEventThunk;
+		private static nint _posixSignalWorkerEventThunk;
 
 		private readonly DirectExecutionBackend _backend;
 		// Windows uses AutoResetEvent (its SafeWaitHandle is a real kernel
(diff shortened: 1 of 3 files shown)


Spooks4576:

Tested `linux/posix-native-workers` @ `2d603bff` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-411

kostyaff (approved):
Reviewed from a POSIX-correctness angle across three areas; all hold.

**(1) Win64→SysV thunk boundary** — reuses the existing `CreateWin64ToSysVThunk` for the new wait/signal worker-event trampolines. Register shuffle is correct (rcx→rdi, rdx→rsi, r8→rdx, r9→rcx), rdi/rsi saved+restored (non-volatile in Win64), rax result forwarded. Stack alignment: entry RSP≡8 (mod 16) → push rdi → 0 → push rsi → 8 → sub rsp,8 → 0 → call → 8 inside callee (SysV convention). Correct for both `WaitWorkerEvent(nint,uint)→int` (2 args) and `SignalWorkerEvent(nint)→int` (1 arg).

**(2) Worker thread lifecycle** — stop path emits `add rsp,0x28; xor eax,eax; ret`, cleanly returning from the pthread start routine (canonical exit, vs Windows` ExitThread). Dispose waits on the done-event (1000ms) then reaps via blocking `pthread_join` (`JoinWorkerThread`, with `holder[1]=1` guarding double-join); wedged worker falls through to `CloseWorkerThreadHandle` (detach + leak), matching the existing Windows wedged-worker behavior. The if/else ensures join XOR close, never both. Rationale for avoiding `pthread_kill(tid,0)` polling (glibc reports returned-but-unjoined as live) is sound. Race where Dispose hits a worker mid-guest-code → wedged → detach is documented and symmetric.

**(3) Windows behavior untouched** — gate `(!IsWindows && !IsLinux) || disabled` preserves the Windows path; `EnsureWorkerLoopExports` Windows branch is verbatim the old `EnsureKernel32Exports`; stop-emit, CreateThread, and Dispose Windows branches are unchanged; the new `if (_workSemaphore != 0)` guard in `SignalWorkAvailable` is a no-op on Windows (early-returns via `_workAvailable` first).

CI green (linux-x64 + osx-x64 + Build Archive pass). Test `PosixNativeWorkerTests.WorkerSignalsDoneEventAndJoins` is Linux-guarded and exercises the done-event → blocking join handshake that is the core teardown invariant.

Non-blocking nits: two stray blank lines after `SignalWorkerEvent`, and the file mixes the `HlePosixHostStubs` alias with bare `PosixHostStubs.` references — cosmetic only.

PR #412, opened 2026-07-18
Threading rework

galvesribeiro:
## Why

Four titles (Lunar Lander Beyond, Hades, Astro Bot, Silent Hill f) stalled in early boot the same way: threads waiting on signals (mutex/cond/sema/event-flag/equeue) that never arrived. Investigation traced this to the guest-thread scheduling layer itself, not the games:

- Every guest thread already owns a **dedicated host thread** (`GuestExecutionRunner`, strictly 1:1), yet blocking was implemented as a hand-rolled cooperative layer on top: capture a CPU continuation, exit guest code, register a wake key, and wait for a producer to find the registration and re-schedule the runner.
- That split **block-and-wake into two racy steps**. Measured in Hades: 514 of 611 `pthread_cond_timedwait` re-acquire hand-offs lost their wake in one run, permanently wedging a mutex that was observably free (`owner==0`).
- Silent Hill f froze with its main thread parked at a `scePthreadGetspecific` import stub — i.e. the host thread was wedged inside the scheduling machinery (`usleep` pumped the scheduler on every poll iteration).

## What

Replace cooperative continuation/wake-key scheduling with **in-place blocking**: when a guest thread blocks, its host thread blocks right there, inside the HLE call, on a host OS primitive. The primitive owns block-and-wake atomically, so the lost-wakeup class dies by construction. No continuation capture, no wake keys, no ready queue, no pump. Semantics sourced clean-room from FreeBSD `libthr`/`ksem` and public kqueue/kevent references (cited per commit).

**One commit per phase**, each verified against the 8-game matrix (void Terrarium + Dream Sara must-not-regress; the four stall games + Dead Cells watched for progress):

1. **Phase 1** `de45605` — pthread mutex + condvar block in place (−416/+104). Hades ~2× faster.
2. **Phase 2** `19dc71b` — semaphores + event flags. Dead Cells 5.2M→10.3M imports.
3. **Phase 3** `ca67322` — event queues (unify infinite + timed waits).
4. **Phase 4** `c17c6ed` — rwlocks (last cooperative primitive) + park-state diagnostics. Nothing calls `RequestCurrentThreadBlock` after this.
5. **Phase 5** `5099816` — guest-exception delivery to parked threads via a SA_RESTART-style wait-loop checkpoint (keeps IL2CPP stop-the-world working; dormant fast-path).
6. **Phase 6** `3d4a880` — **the purge** (−1172/+51): delete `Pump`, `WakeBlockedThreads`, the ready queue, continuation capture, the nested-callback `Thread.Sleep(1)` spin, and the parked-exception hijack.

Plus `2d72d0e` (remove the stale NuGet lock file — nothing enforced it and it broke locked-mode restores on the pinned SDK) and `22a7ca9` (resolve the 5 imports Lunar reached once it ran further).

## Results

The purge didn't just hold the line — it moved it, because two stuck games had their main threads wedged inside the deleted machinery:

| Game | Before | After |
|---|---|---|
| **Silent Hill f** | main thread wedged at 2.4M imports, presenter never initialized | presenter initializes, runs to 4.6M — still no visible content (next wall: a TLS-slot poll) |
| **Lunar Lander** | frozen at 800K imports | 6.7M imports (~8×), 0 unresolved NIDs — still splash-stalled (next wall: an exception-unwind loop) |
| Hades | 24.9M / 60s | 36.5M / 40s |
| Dead Cells | 5.2M | 10.7M |
| void / Dream Sara | render + play | unregressed |
| Astro Bot | splash → black (0 flips) | unregressed |

Zero stall watchdogs and zero unrecovered crashes across all eight; 349 Libs tests pass. To be explicit about scope: no game went from broken to playable in this PR — the stalled titles all progress further and then hit their next, non-scheduler wall. The win is that those walls are now ordinary game-progress bugs (missing produced values, GPU pipeline gaps), not lost-wakeup races in the emulator.

## Net

Roughly **−1500 lines** of concurrency machinery removed; the model is now 1:1 host-thread-per-guest-thread blocking on real OS primitives, which is simpler to reason about and eliminates the lost-wakeup class by construction.


Spooks4576:

Tested `gpu-bootstrap-stall` @ `37b3e787` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-412

kostyaff (changes requested):
Scoped review — not a full line-by-line pass. Posting what I verified, what I did not, and one merge-blocker.

## Merge-blocker (must fix before merge)

CI on the merge-base run (29671941693) fails to compile:

```
src/SharpEmu.Libs/Kernel/KernelSyncOnAddressCompatExports.cs(74,34): error CS0117: 'GuestThreadExecution' does not contain a definition for 'RequestCurrentThreadBlock'
src/SharpEmu.Libs/Kernel/KernelSyncOnAddressCompatExports.cs(121,40): error CS1061: 'IGuestThreadScheduler' does not contain a definition for 'WakeBlockedThreads'
```

`KernelSyncOnAddressCompatExports.cs` is not in this PR's diff, but it was added to main by #422 (sceKernelSyncOnAddressWait/Wake, merged 2026-07-18) after this PR's branch point, and it still uses the cooperative `RequestCurrentThreadBlock`/`WakeBlockedThreads` API that Phase 6 here deletes. The PR HEAD run is green because it sits on a pre-#422 base; merging into current main breaks the build. The fix is to rebase onto current main and migrate `sceKernelSyncOnAddress{Wait,Wake,Cancel}` to the in-place `GuestThreadBlocking` model the way Phase 1–4 did for mutex/cond/sema/eventflag/equeue/rwlock. (Also worth confirming #399's reserve/mutex fix doesn't collide with the Phase 1 mutex rewrite.)

## What I reviewed (in scope)

**`GuestThreadBlocking.cs` (new, 116 lines) — the new model's core.** Sound. `WaitSliceMilliseconds=50` slicing lets parked threads observe `ShutdownRequested` promptly while `Monitor.Pulse` wakes still land immediately. `Checkpoint` releases the gate (`Monitor.Exit`) before running the interrupt deliverer and re-acquires it in `finally` — correct re-entrancy guard, since the handler may re-enter the same primitive. The `_interrupted`/`_blockDescriptions` dictionaries are `ConcurrentDictionary`; `RequestInterrupt`/`NoteBlocked`/`NoteUnblocked` are thread-safe. `DeliverInterruptForCurrentThread` being a parameterless static `Action?` is fine given it is wired to `DeliverPendingGuestExceptionInPlaceForCurrentThread` which resolves the target via current-thread context.

**Phase 1 — pthread mutex/cond (`KernelPthreadCompatExports.cs`).** The in-place rewrite is correct:
- Mutex contended path: `WaiterCount++` under `lock(state)` → `while(OwnerThreadId!=0){Checkpoint; Monitor.Wait(state,slice)}` → acquire inside the loop → `finally{WaiterCount--; NoteUnblocked}`. `Monitor.Wait` atomically releases the lock and parks, so an unlock's `PulseAll` cannot be missed — the lost-wakeup class is gone by construction. Shutdown returns `ORBIS_GEN2_ERROR_TRY_AGAIN`.
- Unlock pulses only when `WaiterCount != 0` (no spurious pulses).
- Cond wait: `Waiters++` under `SyncRoot` **before** `PthreadMutexUnlockCore` → POSIX atomicity holds (a signal issued the instant the mutex unlocks counts us and lands in `SignalsPending`). `SignalsPending` only increments when an unserved waiter exists (`SignalsPending < Waiters`), so stale signals cannot accumulate; the `if (SignalsPending > Waiters) SignalsPending = Waiters` cleanup after `Waiters--` drops phantom signals left by a timed-out waiter. Mutex is re-acquired on every return path via `PthreadMutexLockCore`.
- Lock ordering is consistent: cond wait takes `SyncRoot → mutexState`; mutex path takes `mutexState` only; signal takes `SyncRoot` only. No reverse ordering → no deadlock.
- One subtle point (not a bug): a timed-out waiter that exits the loop can still consume a signal that arrived before the break — but this is all under `lock(SyncRoot)`, so the race is resolved deterministically and matches POSIX (signal-before-timeout wins).

**Consistency across primitives.** The same `NoteBlocked` → try/finally `NoteUnblocked` + `Checkpoint` + `Monitor.Wait(gate, slice)` + shutdown-check pattern is applied uniformly to sema, event flag, equeue, and rwlock (verified via the diff hunks).

**Removed test.** `GuestThreadBlockWaiterRepresentationTests.cs` is deleted — correct, since it reflected over the deleted cooperative API (`RegisterBlockedGuestThreadContinuation`, `TryConsumeCurrentThreadBlock`, `IGuestThreadBlockWaiter`). Gap: `GuestThreadBlocking.cs` has no unit test replacing it; the PR relies on the 8-game matrix for verification, which is stronger for a concurrency primitive but leaves the diagnostics/interrupt helpers unexercised in the test suite.

## What I did NOT review (out of scope for this pass)

- Phase 2–5 semantics for sema/eventflag/equeue/rwlock in depth (only pattern-consistency checked, not per-primitive correctness of count/bitwise/kqueue semantics).
- Phase 5 guest-exception delivery wiring in `DirectExecutionBackend.cs` beyond the `RequestInterrupt`/`RequestShutdown`/`DeliverInterruptForCurrentThread` call sites.
- The Lunar Lander import resolution (`22a7ca9`) and Metal HUD commit (`c583728`) — unrelated to the threading rework, did not look at them.
- The lock-file deletion (`2d72d0e`) — already done on main by #419, so this will likely conflict on rebase; drop that commit on rebase.

## CI

PR HEAD run (29671940641): pass. Merge-base run (29671941693): fail (the SyncOnAddress errors above). `mergeable: MERGEABLE` per the API but `mergeStateStatus: BLOCKED` — consistent with a stale base.

Requesting changes for the rebase + SyncOnAddress migration. Happy to re-review those two things once updated.

kostyaff:
Follow-up to the review — three load-bearing findings in `GuestThreadBlocking.cs` / `DirectExecutionBackend.cs`, verified against the diff.

**1. `_shutdownRequested` is never reset (in-process re-execution bug).** It is a static volatile bool, only ever set by `RequestShutdown()` (called from `RequestHostShutdown`). Nothing in the PR clears it. For the one-game-per-process CLI this is invisible, but in any in-process re-execution scenario — GUI "restart game", or xUnit running multiple `TryExecute`s in one process — the second run starts with `ShutdownRequested == true` and every in-place wait (`pthread_mutex_lock`, `cond_wait`, `sema`, `eventflag`, `equeue`, `rwlock`) immediately returns `ORBIS_GEN2_ERROR_TRY_AGAIN` without blocking. The TryExecute init block already saves/restores `previousGuestThreadScheduler` (line ~752) but does not reset `_shutdownRequested`. Suggest resetting it at the top of `TryExecute` (or in a `ResetForNewExecution()` called there), symmetric with how `Scheduler` is re-bound.

**2. `DeliverInterruptForCurrentThread` is set without save/restore.** Line ~754 sets it unconditionally at TryExecute entry, but unlike `GuestThreadExecution.Scheduler` immediately above (which saves `previousGuestThreadScheduler` and restores in finally), this static `Action?` is overwritten with no restore. Same in-process re-execution concern: a second backend instance clobbers the first's deliverer, and if the first is still unwinding parked threads mid-teardown they will see the wrong (or null) deliverer at their next `Checkpoint`. Suggest the same save/restore pattern: `var previousDeliverer = GuestThreadBlocking.DeliverInterruptForCurrentThread; … finally { GuestThreadBlocking.DeliverInterruptForCurrentThread = previousDeliverer; }`.

**3. `Checkpoint` consumes the interrupt flag even when the deliverer is null (interrupt loss).** In `GuestThreadBlocking.Checkpoint`:

```csharp
if (_interrupted.IsEmpty || guestThreadHandle == 0 || !_interrupted.TryRemove(guestThreadHandle, out _))
    return;
var deliver = DeliverInterruptForCurrentThread;
if (deliver is null)
    return;   // flag already removed → interrupt lost
```

`_interrupted.TryRemove` runs before the `deliver is null` check, so if the deliverer is not wired (e.g. a non-guest host-side compatibility caller, or the race window in #2 where the static was clobbered to null), the interrupt flag is consumed and the queued guest exception is silently dropped. Suggest null-checking the deliverer before `TryRemove`, or re-adding the flag in the null branch.

These are the kind of issues that only surface under re-execution or the null-deliverer race, so they will not show up in a single-game test pass — worth fixing now rather than after they bite a GUI/restart path.

galvesribeiro:
Re: the review — the merge-blocker was verified and is fixed in `528985c`:

- Merged current main into the branch. (Squash-merge target, so merge-vs-rebase yields the same squashed result on main; a merge avoids force-pushing 60+ commits mid-review.)
- `sceKernelSyncOnAddressWait/Wake` (#422) migrated to the in-place `GuestThreadBlocking` model per phases 1–4: per-address gate, generation check under the gate (bump-before-pulse keeps the register-vs-park race closed), sliced wait with shutdown unwind + exception checkpoint, 100 ms self-heal bound retained. The separate host-fallback path collapsed into the single in-place path. Wake-one degrades to wake-all; resumed waiters re-evaluate their condition (futex semantics, as the original comment already assumed).
- #419's lock-file removal merges cleanly with ours; confirmed main did **not** touch `KernelPthreadCompatExports` since the fork point, so no #399 collision.
- #424's new pthread semaphore exports sit on our in-place `KernelWaitSema` unchanged — notably its new `PthreadSemaphoreSemanticsTests` suite now runs against the reworked implementation and passes.

Full solution builds on the merged tree; 431 tests pass (371 Libs incl. the #424 suite, 27 Metal golden, 33 source-gen); void Terrarium + Hades smoke-runs clean. On the test-gap note: agreed `GuestThreadBlocking` deserves direct unit coverage — happy to add a small suite (Checkpoint gate release/reacquire, NoteBlocked/DescribeBlock, interrupt consume-once) as a follow-up commit if you want it in this PR rather than after.

kostyaff:
Re-review at `37b3e78` following the respin. Short version: the merge-blocker is verified fixed — thank you for the careful SyncOnAddress migration. The three load-bearing findings from the previous review are still open, so the request-changes stands; all three are cheap fixes.

**Merge-blocker — verified resolved ✅**

- Zero references to `RequestCurrentThreadBlock` / `WakeBlockedThreads` / `IGuestThreadBlockWaiter` remain in the PR tree (main has 8). No dangling references after a squash-merge.
- The SyncOnAddress migration is correct: capturing the generation outside the gate is safe because the wait loop re-checks it under the gate, and the waker's bump-then-pulse ordering (pulse under the same gate) closes the register-vs-park window. The sliced wait + `ShutdownRequested` + `Checkpoint` is strictly better than main's single-shot host fallback. Futex-tolerant simplifications (no compare value, wake-one→wake-all) are documented and match prior behavior.
- Deleting `GuestThreadBlockWaiterRepresentationTests` is legitimate — it pinned the removed API.

**The three findings are unchanged at `37b3e78`:**

1. **`_shutdownRequested` is still never reset** (`GuestThreadBlocking.cs:25,34` — set-only). There is already a precedent in the tree: `HostSessionControl.ResetShutdownRequest()` exists ("Starts a fresh session after the previous guest has fully left its execution backend") — though note it currently has no callers either. `TryExecute` is itself multi-call in-process (per module initializer), and any future in-process re-execution leaves every in-place wait (mutex/cond/sema/eventflag/equeue/sync-on-address) permanently unwinding. A reset next to where the scheduler is saved closes the whole class.

2. **`DeliverInterruptForCurrentThread` is set without save/restore** (`DirectExecutionBackend.cs:1150`; the `finally` restores only `GuestThreadExecution.Scheduler`). The asymmetry with `Scheduler` is exactly the issue: the stale delegate outlives `Execute`, and in-process re-execution would deliver through a delegate captured by a previous backend instance.

3. **`Checkpoint` still consumes the interrupt flag when the deliverer is null** (`GuestThreadBlocking.cs:94→98`): `TryRemove` runs before the `deliver is null` check, so an interrupt raised during a deliverer-less window (pre-first-`Execute` HLE wait, tests) is silently lost. For an IL2CPP stop-the-world suspend that's a lost-wakeup-class hang. One-line fix: read the deliverer first, `TryRemove` only when non-null.

**Test gap:** since the old waiter-representation suite was removed with its API, `GuestThreadBlocking` now has zero direct coverage. Your offer of a small suite (Checkpoint gate release/reacquire, NoteBlocked/DescribeBlock, interrupt consume-once) would be very welcome **in this PR** — a consume-once test pins finding 3's semantics directly.

**Non-blocking notes:**

- Scope since the respin grew: `37b3e78` (SHA-1 software emulation under Rosetta) and `de0816e` (Hades AGC DCB builders) are independent features landing under a "Threading rework" title; a squash-merge would bury them in one commit. The SHA-1 code itself looks correct (MSG1/MSG2/NEXTE/RNDS4 semantics and constants match the Intel spec, and it has tests). Whether to split is the maintainer's call.
- Minor: `_addressGates` / `_wakeGenerations` grow unboundedly across unique addresses (same as main's `_hostAddressGates`, so not a regression — flagging while the file is open).

CI is green and the branch is mergeable; with the three fixes + the blocking suite this has my approve.

Hog185:
Doom  + doom ii crashes a bit later
brings powerwash simulator to booting
No other changes in my testing

shadowbeat070:
What is the current status with this PR?

PR #414, opened 2026-07-18
[CPU] Scan TLS access pattern at final code offset

Dafenxz0:
## What changed

`FindTlsAccessPatterns` stopped one byte range too early because its loop treated the last valid pattern start as an exclusive bound. A TLS load placed at the end of a guest code block was therefore left unpatched.

This changes that bound to inclusive and adds a regression test with the TLS sequence at the final valid offset.

## Why

`StubManager.PatchTlsAccess` relies on this scan before native guest code runs. Missing the final sequence can leave guest TLS access using the unpatched instruction instead of SharpEmu's TLS stub. The fix is generic and does not change the generated stub or any unrelated CPU path.

## Testing

Game testing: N/A. This is a deterministic scanner boundary case.

- Added a regression test that failed before the fix and passes after it.
- `dotnet test SharpEmu.slnx --no-restore` — 352 passed, 0 failed.

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/JitStubs.cs
+++ b/src/SharpEmu.Core/Cpu/Native/JitStubs.cs
@@ -252,7 +252,7 @@ public static unsafe class JitStubs
         var pattern = TlsAccessPattern;
         var end = start + length - pattern.Length;
 
-        for (var ptr = start; ptr < end; ptr++)
+        for (var ptr = start; ptr <= end; ptr++)
         {
             if (MatchesPattern(ptr, pattern))
             {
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `agent/fix-terminal-tls-pattern-scan` @ `616cc488` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-414

PR #410, opened 2026-07-18
Gpu runtime stalls

par274:
## Before submitting

Please read our contribution guidelines before opening a pull request:

➡️ [**CONTRIBUTING.md**](https://github.com/sharpemu/sharpemu/blob/main/CONTRIBUTING.md)

By opening this pull request, you confirm that you have read and agree to follow the contribution guidelines.

## Testing

* Demons Souls
* Dreaming Sarah
* Void Terrarium

## Checklist

By submitting this pull request, you confirm that:

- [x] I have read and followed `CONTRIBUTING.md`.
- [x] I tested my changes or marked the testing section as `N/A`.
- [x] I wrote this pull request description myself and did not paste AI-generated explanations.
- [x] I listed the game(s) I tested (or marked the testing section as `N/A`).

## Fixes regression

* Fix memory leak: 19 GB to 600 MB, on Demon's Souls. This is just boot.

### On Demon's Souls;

* Because the CPU waited for the GPU fence and memory writeback after each compute dispatch, the GPU queue was running in full concurrent order. Fixed this. FPS is still very low, but it's now hovering around 3–4 FPS. That's still progress, the game boots much faster and reaches the first image sooner, which also makes debugging a lot less time-consuming.
* Fixed regression at #283 on Windows.
* It also improved frame pacing in many games. For example, Void Terrarium now runs at a stable 60 FPS.

resim

## General fixes

* Bink videos never skipped, because this is not true emulation.

PR #416, opened 2026-07-18
[CPU] Skip idle guest-exception work on HLE imports

Dafenxz0:
## What changed

This is a follow-up for #410 and targets `gpu-runtime-stalls`, not `main`.

Every HLE import currently builds a full `GuestCpuContinuation` and enters `_guestThreadGate` to discover that no guest exception is pending. The Dead Cells evidence for #410 reaches more than eight million import dispatches during its 60-second run, so that idle check sits on a very hot path.

This commit keeps an atomic count synchronized with the pending-exception dictionary. The normal path now returns after one volatile read. Register capture and the existing lock/delivery path run only while at least one exception is queued.

Queueing, replacement, follow-up delivery, removal and shutdown clear all update the count under the same lock that already protects the dictionary. A concurrent exception raised immediately after a zero read is observed at the next HLE boundary, which is the same boundary-based delivery contract used today.

## Expected impact

- removes one uncontended `Monitor` acquisition from every ordinary HLE import;
- avoids capturing the full guest register continuation when there is no exception;
- leaves exception delivery and HLE return behavior unchanged.

I have not claimed an FPS number yet. The purpose of this dependent PR is to let the existing Dead Cells and Dreaming Sarah runner compare #410 with and without this commit on the same hardware.

## Testing

- `dotnet test SharpEmu.slnx` — 412 passed across the three test projects, 0 failed.
- Manual game testing: N/A; awaiting the automated Linux gameplay runner.

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Imports.cs
@@ -532,7 +532,8 @@ public sealed partial class DirectExecutionBackend
 			}
 			DeliverPendingGuestExceptionAtSafePoint(
 				cpuContext,
-				CaptureImportBoundaryContinuation(cpuContext, argPackPtr, num7));
+				argPackPtr,
+				num7);
 			StoreImportVectorReturn(cpuContext, argPackPtr);
 			if (dispatchResolved &&
 				orbisGen2Result == OrbisGen2Result.ORBIS_GEN2_OK &&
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `agent/pr410-guest-exception-fast-path` @ `c60d6016` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-416

par274:
#410 already merged.

Send a PR for main branch, then I test it.

PR #417, opened 2026-07-18
[VideoOut] Detect and recover stuck presentation frame fences

xnetcat:
## Before submitting

I have read and agree to follow the contribution guidelines.

## What this fixes

On macOS/MoltenVK, a frame submitted against a swapchain acquire whose drawable never arrives (for example when the window is occluded: MoltenVK recycles presentable images only from the on-screen presented-drawable callback) leaves that frame slot's fence pending forever. `RenderCore` then skips every subsequent frame before reaching the guest work drain or another acquire/present, so the ordered guest queue jams at capacity, the window freezes on the last presented image permanently, and the guest keeps running blind. Un-occluding the window does not recover because the only code path that could re-acquire is gated behind the stuck fence.

## What this does

- Tracks consecutive frame-slot misses and warns with per-slot fence status once the streak passes two seconds (`vk.frame_slot_stuck` / `vk.frame_slot_recovered`).
- Past a recovery threshold (default 8 s, `SHARPEMU_FRAME_STUCK_RECOVERY_MS`, 0 disables) destroys the swapchain first — aborting MoltenVK's pending drawable chain — retires the unblocked frame and guest-submission fences under bounded waits, then rebuilds presentation resources.
- If the queue still does not drain, falls back to the existing device-lost drain so the guest keeps running.

## Testing

- Diagnosed and reproduced with SILENT HILL: The Short Message (PPSA10112) on an Apple M4 (Rosetta, MoltenVK 1.4.0); the wedge signature and recovery behavior verified live.
- Full SharpEmu.Libs test suite passes.

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
+++ b/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
@@ -2736,6 +2736,13 @@ internal static unsafe class VulkanVideoPresenter
         private TranslatedDrawResources?[] _frameTranslatedResources = [];
         private GuestImageResource?[] _frameGuestImageVersions = [];
         private int _currentFrameSlot;
+        // Wall-clock diagnostics for a frame slot whose fence stays unsignaled
+        // across consecutive Render() calls. A short streak is a normal GPU
+        // backlog; a long one means the present path is wedged (e.g. a frame
+        // submitted against an acquire semaphore whose drawable never arrived),
+        // which otherwise freezes the window silently forever.
+        private long _frameSlotStuckSinceTicks;
+        private long _frameSlotStuckLastWarnTicks;
         // Monotonic submission/completion counters across every queue submit
         // (guest batches, compute chunks and presents). Fences on a single
         // queue signal in submission order, so "timeline <= completed" means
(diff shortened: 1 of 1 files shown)


Spooks4576:

Live status: https://pr.sharpemu.app/#pr-417

kostyaff (approved):
Reviewed at `ca83fe4` (diff vs `a60bfc9`, +197/−0, single file). CI green (Build Archive / linux-x64 / osx-x64 pass).

The diagnosis and the recovery mechanism check out:

- The trigger path is clean: `ReportFrameSlotStuck` runs from the `TryWaitFrameSlot` early-return in `Render()` with no lock held; `_gate` is only taken briefly to snapshot `_pendingGuestWorkCount`. Recovery itself runs without `_gate`.
- The unblock mechanism is spec-sound, not just MoltenVK folklore: `vkDestroySwapchainKHR` forces pending acquire semaphores to complete, which lets the submitted batch retire and its fence signal — verified against the fence-wait + drain ordering here, with bounded waits throughout (2s per fence, 2s drain deadline).
- The failure path integrates with the existing device-lost flow (`_deviceLost = true`, Render's drain-and-return branch) instead of inventing a parallel state.
- Stuck-state tracking is consistent: slot fences on a single queue signal in submission order, so a wedged oldest fence blocks all later ones — the single shared `_frameSlotStuckSinceTicks` can't flap between slots.
- Warn cadence is bounded (2s first warn, 10s re-warn), thresholds are env-tunable, and recovery is disablable (`SHARPEMU_FRAME_STUCK_RECOVERY_MS=0`).

One finding worth a respin (minor, but user-visible):

**Deferred recovery resets its own throttle → WARN spam every frame.** In `RecoverStuckPresentation`, the zero-framebuffer (miniaturized) branch logs `vk.frame_stuck_recovery_deferred` and then resets `_presentationRecoveryLastAttemptTicks = 0`. The very next `Render()` (≈16 ms later) finds `stuckMs` still past the recovery threshold with the throttle cleared, so it re-enters, defers, and logs again — an identical WARN line per rendered frame for as long as the window stays miniaturized. Keeping the throttle across the defer (just drop the reset) preserves the intended 30 s retry cadence.

Two accepted tradeoffs, noting for the record (not asking for changes): recovery blocks the render/main thread for a few seconds worst-case (per-slot fence waits + drain loop) — a beachball during explicit recovery is still strictly better than the permanent freeze it replaces; and destroying a live swapchain is inherently driver-risky, but the trigger (8 s unsignaled frame fence) only fires in states where the presentation is already wedged, with device-lost as the floor.

Approving; the defer-throttle nit doesn't gate the value.

PR #418, opened 2026-07-18
[AGC/Vulkan] Mirror guest-image DMA copies on the GPU

xnetcat:
## Before submitting

I have read and agree to follow the contribution guidelines.

## What this fixes

A CP DMA copy whose source is a live guest image was mirrored into the destination's Vulkan image by re-reading the destination's guest CPU memory after the CPU-side copy. When the source is a render target (SILENT HILL: The Short Message feeds its title-movie plane by DMA from the NV12 conversion pass's render target), the source's freshest pixels exist only in its Vulkan image until a writeback lands — so the mirror overwrote the destination image with stale black CPU bytes and every later composite sampled black.

## What this does

When both DMA endpoints are registered guest images with matching extents and formats, mirror the copy image-to-image on the GPU (transfer-layout round trip for both images inside the batched guest command stream), preserving the unified-memory semantics the guest relies on. All other cases keep the existing CPU-bytes mirror.

## Testing

- Root-caused from an AGC trace of PPSA10112's title flow (rt_writer NV12 conversion draws into rotating buffers, then `dma_data` into the composited movie plane, presented black).
- Builds clean against main; SharpEmu.Libs test suite passes on the development branch carrying the same change.

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -3630,7 +3630,8 @@ public static partial class AgcExports
                         ctx,
                         destinationAddress,
                         byteCount,
-                        immediateFill ? (uint)sourceAddress : null);
+                        immediateFill ? (uint)sourceAddress : null,
+                        immediateFill ? 0 : sourceAddress);
                 }
 
                 if (tracePacket)
@@ -4129,7 +4130,8 @@ public static partial class AgcExports
         CpuContext ctx,
         ulong destinationAddress,
         ulong byteCount,
-        uint? fillValue)
+        uint? fillValue,
+        ulong sourceAddress = 0)
     {
         var hasImage = GuestGpu.Current.TryGetGuestImageExtent(
             destinationAddress,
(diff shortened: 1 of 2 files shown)


Spooks4576:

Live status: https://pr.sharpemu.app/#pr-418

kostyaff (approved):
Reviewed at `4712cd1` (diff vs `a60bfc9`, +165/−3, two files). CI green (Build Archive / linux-x64 / osx-x64 pass).

The core design is right and the Vulkan usage checks out:

- The stale-source diagnosis is sound: when the DMA source is a live render target, its freshest pixels exist only GPU-side until a writeback, so mirroring the destination's CPU bytes was actively wrong for exactly the reported title-movie flow.
- Fallback structure is conservative in the right direction: `SubmitGuestImageCopy` returns false for closed/unregistered/size-mismatched endpoints and the caller drops through to the existing CPU mirror unchanged.
- The partial-copy concern doesn't apply: the pre-existing `byteCount < imageBytes` early-return in `MirrorDmaWriteToGuestImage` means the mirror (and hence this GPU path) only runs on full-image coverage.
- The barrier round-trip is correct: src `ColorAttachmentWrite|ShaderRead → TransferRead` gated on `ColorAttachmentOutput|FragmentShader → Transfer`, dst first-write handled via `Undefined`/`Initialized`, both images restored to `ShaderReadOnlyOptimal` — consistent with the file's guest-image layout invariant, all inside the batched guest command stream (no queue-order violation vs. draws and writebacks).

Two edge findings, both cheap (not blocking):

1. **Format mismatch is discovered only at execute time, after the fallback is gone.** `SubmitGuestImageCopy` gates on width/height; `ExecuteGuestImageCopy` additionally rejects `source.Format != target.Format` — but by then `MirrorDmaWriteToGuestImage` has already returned, so a same-size/different-format image-to-image DMA (plausible in video paths, e.g. RGBA→NV12) gets neither the GPU copy nor the CPU mirror and the destination silently keeps its previous content. Carrying the guest format in `_guestImageExtents` would let the enqueue side reject up front and fall back to the CPU path.

2. **Self-copy (`sourceAddress == destinationAddress`) passes the enqueue check.** Same extents trivially match, and the execute path would then put one image into two different layouts in a single barrier pair — a validation error. A one-line `sourceAddress == destinationAddress` guard returning false (CPU mirror harmlessly re-uploads identical bytes) closes it.

Approving on the strength of the main path; both items are edge hardening.

PR #422, opened 2026-07-18
[Kernel] Implement sceKernelSyncOnAddressWait/Wake

NickPom:
## What this adds

libKernel's futex-style address-wait primitives `sceKernelSyncOnAddressWait`
(NID `Hc4CaR6JBL0`) and `sceKernelSyncOnAddressWake` (NID `q2y-wDIVWZA`) were
not implemented. A guest thread that parks on an address returned from the
wait immediately, so runtimes that build their own spinlocks and queues on
top of these primitives busy-spin forever with no forward progress.

Observed in Juicy Realm (PPSA19268): the wait was dispatched roughly 15.8
million times in a few seconds while the wake fired only a handful of times,
the process pinned several cores at around 650% CPU, and boot never advanced.

## How it works

Both are implemented over the existing cooperative block scheduler
(`RequestCurrentThreadBlock` / `IGuestThreadScheduler.WakeBlockedThreads`),
keyed on the guest address:

- Wait parks the calling guest thread on `sceKernelSyncOnAddress:<addr>` until
  woken. A per-address wake generation is captured on entry and the wake
  predicate stays unsatisfied until a wake bumps it. This is required: an
  always-satisfied predicate is treated as an immediate late-arrival by the
  dispatcher and never yields, which reproduces the original spin.
- Wake bumps the address generation, which also closes the
  register-versus-park race, and releases waiters through the same key.
- The real primitive takes a compare value that decides whether to sleep. That
  value is not recovered here, so each wait is given a bounded self-heal
  deadline and treated as a spurious-wakeup-tolerant park, which futex callers
  already handle. A non-cooperative caller (the host main thread) takes an
  equivalent bounded host wait.

No new dependencies, no ABI or registry changes, and no game-specific logic:
this is generic to any title that uses these primitives.

## Testing

macOS (Apple Silicon, Rosetta 2), native CPU backend:

- Juicy Realm (PPSA19268) - before: busy-spins at boot, roughly 15.8M wait
  calls, around 650% CPU, zero forward progress. After: the spin is gone
  (around 3% CPU) and boot advances through IL2CPP initialization to the GC
  stop-the-world stage.
- Dreaming Sarah (PPSA02929) - unaffected (does not use the primitive), around
  7,900 draw submissions, no regression.
- Arcade Game Zone (PPSA19015) - unaffected (does not use the primitive),
  behavior unchanged.

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelSyncOnAddressCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelSyncOnAddressCompatExports.cs
@@ -1,0 +1,140 @@
+// Copyright (C) 2026 SharpEmu Emulator Project
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+using System.Collections.Concurrent;
+using System.Threading;
+using SharpEmu.HLE;
+
+namespace SharpEmu.Libs.Kernel;
+
+// libKernel's address-wait primitives (sceKernelSyncOnAddress*) are the PS5's
+// futex-style wait/wake: a thread parks on a guest address until another thread
+// wakes that address. Guest runtimes (seen driving Juicy Realm, PPSA19268)
+// build their own spinlocks/queues on top of it and call the wait in a hot
+// loop; left unimplemented, every wait returns immediately and the runtime
+// busy-spins forever (millions of calls, no forward progress).
+//
+// This implements wait/wake over the existing cooperative-block scheduler,
+// keyed on the address. The real primitive takes a compare value so the wait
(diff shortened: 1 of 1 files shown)


Spooks4576:

Tested `feat/sync-on-address` @ `a382cc39` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-422

PR #430, opened 2026-07-19
Gui Settings Null list Entries

kadu04t:
## Description

GuiSettings properties are declared non-nullable, but that doesn't stop the deserializer from assigning null from the JSON. The fallback in Load() doesn't catch this because it only returns new GuiSettings() when the entire deserialization result is null - not when individual properties inside it are null. Since no exception is thrown during deserialization, the catch block in Load() never runs either.

While fixing this, I found the same issue inside lists: "EnvironmentToggles": [null] deserializes fine and the null value flows straight into Environment.SetEnvironmentVariable later, throwing an exception. I fixed both cases.

## Testing

N/A. No game-specific testing was applicable. Dotnet test - all 438 passed 

## Checklist

### DIFF
--- a/src/SharpEmu.GUI/GuiSettings.cs
+++ b/src/SharpEmu.GUI/GuiSettings.cs
@@ -71,7 +71,7 @@ public sealed class GuiSettings
             if (File.Exists(SettingsPath))
             {
                 var json = File.ReadAllText(SettingsPath);
-                return JsonSerializer.Deserialize<GuiSettings>(json, SerializerOptions) ?? new GuiSettings();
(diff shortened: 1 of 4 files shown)


Spooks4576:

Tested `gui-settings-null-list-entries` @ `0f1b54f6` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-430

PR #433, opened 2026-07-19
[Kernel] Implement sceKernelMapDirectMemory2

NickPom:
## What this adds

`sceKernelMapDirectMemory2` (NID `BQQniolj9tQ`) was unimplemented while its v1
counterpart `sceKernelMapDirectMemory` (NID `L-Q3LEjIbgA`) already exists. A
title that calls the v2 variant therefore hit an unresolved import that
returned an error value the guest then treated as the mapped address.

## How it works

The "2" variant inserts a `memoryType` argument (rdx) ahead of v1's protection,
shifting protection, flags and directMemoryStart down one register each and
pushing alignment onto the stack (the 7th argument, at `[rsp + 8]`, above the
return address). v1's body is extracted into a shared `MapDirectMemoryCore`
and both exports route through it. v2 reads its shifted arguments plus the
stack alignment and accepts the memoryType, which only selects cache/GPU
access attributes this HLE does not model per mapping, so it does not affect
placement. The ABI shift was confirmed against real call arguments (the value
in rcx is a valid protection mask, which only makes sense in v2's layout).

## Testing

macOS (Apple Silicon, Rosetta 2), native CPU backend:

- Gex Trilogy (PPSA28542) - calls `sceKernelMapDirectMemory2` during init.
  Before: unresolved import returning an error. After: it maps like v1 and
  returns a valid address. (The title progresses to a later spin on separate
  still-unimplemented AGC/GPU driver functions, so this alone does not make it
  playable, but the mapping call is now handled with no crash.)
- Dreaming Sarah (PPSA02929) - no regression from the shared-core refactor
  (it exercises v1); around 1,800 draw submissions.

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
@@ -2827,12 +2827,51 @@ public static partial class KernelMemoryCompatExports
         LibraryName = "libKernel")]
     public static int KernelMapDirectMemory(CpuContext ctx)
     {
-        var inOutAddressPointer = ctx[CpuRegister.Rdi];
-        var length = ctx[CpuRegister.Rsi];
-        var protection = unchecked((int)ctx[CpuRegister.Rdx]);
-        var flags = ctx[CpuRegister.Rcx];
-        var directMemoryStart = ctx[CpuRegister.R8];
-        var alignment = ctx[CpuRegister.R9];
+        return MapDirectMemoryCore(
+            ctx,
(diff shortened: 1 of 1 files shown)


Spooks4576:

Tested `feat/map-direct-memory2` @ `9205c7c0` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-433

PR #435, opened 2026-07-19
[CPU] Skip idle exception work at HLE import boundaries

Dafenxz0:
## Before submitting

Please read our contribution guidelines before opening a pull request:

➡️ [**CONTRIBUTING.md**](https://github.com/sharpemu/sharpemu/blob/main/CONTRIBUTING.md)

By opening this pull request, you confirm that you have read and agree to follow the contribution guidelines.

## What changed

Every HLE import currently captures the guest continuation and enters `_guestThreadGate` to check for a pending guest exception. Most imports have no exception waiting, so that work is repeated on the normal path for no result.

This adds a small atomic pending count. When the count is zero, the import returns from the exception check after one volatile read. When an exception is queued, the existing locked path still removes it, captures the continuation, and delivers it as before. Queue replacement, follow-up delivery, removal, and shutdown keep the count synchronized with the dictionary.

An exception queued just after the zero check is observed at the next import boundary, which preserves the existing boundary-based delivery model.

## Why

The Dead Cells trace from #410 recorded more than eight million HLE imports in a 60-second run. Removing a continuation capture and an uncontended monitor acquisition from that common path reduces CPU work without changing GPU memory, command ordering, or emulation state.

This is the same focused change previously submitted as #416 against the #410 branch. That PR was closed after #410 merged, with a request to submit it against `main` for normal game testing. This branch is rebased on the current `main`.

I am not attaching an FPS claim from local testing. The improvement needs to be measured by the gameplay runner on the same real hardware used for other performance PRs.

## Testing

- `dotnet test SharpEmu.slnx -c Release --no-restore -m:1`
  - SharpEmu.Libs.Tests: 391 passed
  - SharpEmu.ShaderCompiler.Metal.Tests: 27 passed
  - SharpEmu.SourceGenerators.Tests: 33 passed
  - Total: 451 passed, 0 failed
- The earlier #416 run passed Dead Cells and Dreaming Sarah without gameplay regressions. This rebased `main` version still requires a fresh runner comparison.

## Checklist

- [x] I have read and followed `CONTRIBUTING.md`.
- [x] I tested my changes or marked the testing section as `N/A`.
- [x] I wrote this pull request description myself and did not paste AI-generated explanations.
- [x] I listed the game(s) I tested (or marked the testing section as `N/A`).

PR #443, opened 2026-07-19
[AGC] Apply color-target component swaps to pixel exports

Dafenxz0:
## Why

The first version of this draft reinterpreted the image at VideoOut presentation time. Hog185's Dead Cells test showed no visual change, so that implementation has been removed completely.

The component order is lost earlier. SharpEmu already decodes `CB_COLOR_INFO.FORMAT` and `NUMBER_TYPE`, but it ignored bits 11–12, `COMP_SWAP`. Those bits define how pixel-shader export lanes map into a color target. For a four-component target, the alternate mode is B, G, R, A; treating it as the standard R, G, B, A order can produce the red/cyan scene seen in the recording.

This matches the color-target model and export handling used by [shadPS4](https://github.com/shadps4-emu/shadPS4/blob/a9e825ed3f3d1fb5a03bcdabf68216781452ed1c/src/video_core/amdgpu/regs_color.h#L263-L313) and its [shader export translator](https://github.com/shadps4-emu/shadPS4/blob/a9e825ed3f3d1fb5a03bcdabf68216781452ed1c/src/shader_recompiler/frontend/translate/export.cpp#L120-L132).

## What changed

- Decode `CB_COLOR_INFO.COMP_SWAP` for every active color target.
- Carry component swap, component count, and the guest data format in the backend-neutral pixel-output binding.
- Apply all four hardware mappings—standard, alternate, standard reverse, and alternate reverse—while emitting both SPIR-V and MSL.
- Preserve partial and compressed export masks after remapping.
- Handle the host packed-format remap for guest `11_11_10` and the host red lane used by single-component formats.
- Include format and component-swap state in the graphics shader-cache key so a shader compiled for one target layout is never reused for another.
- Add a permanent `mrt-swap` ShaderDump module, mapping-table coverage, a Vulkan compilation test, and an MSL source test.

The old VideoOut snapshot-format changes are no longer part of the PR. Standard-swap targets generate the same component order as before; only targets that explicitly request another hardware mapping change behavior.

## Testing

- `dotnet build SharpEmu.slnx -c Release --no-restore` — succeeded.
- `dotnet test SharpEmu.slnx -c Release --no-build` — 471 passed.
- ShaderDump generated 28 synthetic modules, including `mrt-swap-ps.spv`.
- All 28 modules pass `spirv-val --target-env vulkan1.3`.
- Disassembly of the alternate RGBA export confirms the output is constructed from `v2, v1, v0, v3` (B, G, R, A).

## Real-game validation

The earlier “no change” result applies to commit `54b3dba`, which used the removed presentation-time approach. The replacement is commit `3ec6c71`.

Dead Cells still needs to be rerun on the new artifact, ideally together with #435 while that boot/progress change is not yet on `main`. The expected result is corrected color placement; I am deliberately not marking that result as confirmed until the recording is available.

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -6824,6 +6824,7 @@ public static partial class AgcExports
                 (attrib2 & 0x3FFFu) + 1,
                 (info >> 2) & 0x1Fu,
                 (info >> 8) & 0x7u,
+                // CB_COLOR_INFO.COMP_SWAP occupies bits 11..12.
                 (info >> 11) & 0x3u,
                 (attrib3 >> 14) & 0x1Fu));
         }
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -206,7 +206,7 @@ public static partial class AgcExports
     // shares _submitTraceGate with tracing).
     private static readonly ConcurrentDictionary<
         (ulong Es, ulong EsState, ulong Ps, ulong PsState, ulong OutputLayout,
-         uint OutputCount, uint Attributes, uint PsInputEna, uint PsInputAddr,
+         ulong OutputSwizzles, uint OutputCount, uint Attributes, uint PsInputEna, uint PsInputAddr,
          ulong AliasAlignment),
         (IGuestCompiledShader Vertex, IGuestCompiledShader Pixel)> _graphicsShaderCache = new();
     private static readonly ConcurrentDictionary<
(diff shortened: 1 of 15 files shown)


Hog185:
no change in dead cells

j92580498-max:
https://docs.amd.com/v/u/en-US/rdna2-shader-instruction-set-architecture https://docs.discoverer.bg/zen2.html

Dafenxz0:
Closing this after gameplay validation. The component-swap path is useful shader state to model, but this change produced no visible difference in Dead Cells, so it is not the cause of the red/pink output I was trying to fix.

Comparing the runner captures made the actual issue clear: the affected frames are linear-float VideoOut sources being presented to an UNORM swapchain without the display transfer conversion. #448 addresses that path directly and its Dead Cells recording shows the scene with the expected colors.

Keeping this PR open would add unrelated shader complexity without evidence that it fixes a real title. Thanks @Hog185 for testing it.

PR #447, opened 2026-07-19
Refresh CPU-rewritten guest textures by write generation

xnetcat:
## Problem
A guest texture whose backing memory the CPU rewrites after its first upload (video frame planes, streamed font/subtitle atlases) never re-uploads: `IsGuestImageUploadKnown` only checks address availability, so the draw path keeps shipping empty pixels and the image stays frozen at its first contents.

## Fix
`GuestImageWriteTracker` gains a monotonic per-range write generation (bumped once per armed write-fault lifetime, carried across range replacement, surviving `ConsumeDirty`). The draw path ships the generation captured at texel-read time; the presenter records the uploaded generation per CPU-backed image and `IsGuestImageUploadKnown` returns false when the tracker has a newer generation — so the existing fingerprint-compared refresh path receives fresh texels exactly when the CPU rewrote the memory. Content-identical rewrites advance the recorded generation so restaging stops after one frame.

Default-on, no env vars, no title-specific logic. On Windows the tracker is disabled, `TryGetWriteGeneration` returns false, and behavior is unchanged.

## Validation
- Superliminal (PPSA06084): movie NV12 planes previously froze on the first (black) frame and subtitle/font atlases rendered as damaged fragments; with this change fresh planes upload every frame and subtitles render perfectly (visually verified).
- New unit tests for the tracker generation semantics (4 cases); full SharpEmu.Libs.Tests suite green (362/362).
- `dotnet publish` osx-x64 builds clean.

### DIFF
--- a/src/SharpEmu.HLE/GuestImageWriteTracker.cs
+++ b/src/SharpEmu.HLE/GuestImageWriteTracker.cs
@@ -32,6 +32,7 @@ public static unsafe class GuestImageWriteTracker
         public int Armed;
         public int FirstCpuWriteSeen;
         public int PendingFirstCpuWrite;
+        public long WriteGeneration;
         public bool TraceLifetime;
         public long SourceSequence;
         public long FirstCpuWriteTraceSequence;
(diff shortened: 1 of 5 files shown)


Spooks4576:

Tested `fix/cpu-texture-write-generation` @ `a1b51508` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-447

PR #448, opened 2026-07-19
Encode linear-float VideoOut flips to sRGB at present

xnetcat:
## Problem
PS5 float VideoOut buffers (R16G16B16A16Sfloat flips) hold linear scRGB values; hardware scan-out applies the display transfer function. Presenting them with a raw `vkCmdBlitImage` into a UNORM swapchain performs only numeric conversion, so a dim linear scene (e.g. ~0.02-0.13 linear) lands at a handful of 8-bit steps and reads as black while UI whites stay visible.

## Fix
When the presented source is a linear-float format and the swapchain format has an sRGB counterpart, blit through a cached swapchain-sized sRGB intermediate (the sRGB store performs the linear→sRGB encode) and then raw-copy (`vkCmdCopyImage`, same compatibility class) into the swapchain image. Non-float sources and swapchains without an sRGB counterpart keep the existing direct blit. The intermediate is recreated on extent change and destroyed with the swapchain resources.

Default-on, no env vars, no title-specific logic; pure Vulkan format/blit logic identical across platforms.

## Validation
- Superliminal (PPSA06084): dim linear-float frames (loading screens, watermark overlays) previously invisible became correctly visible (visually verified).
- New unit tests for the format helpers (16 cases); VideoOut test namespace green; full SharpEmu.Libs.Tests suite green.
- `dotnet publish` osx-x64 builds clean. Cost: one RGBA8 swapchain-sized image (~8 MB at 1080p) plus one extra blit+copy per frame, only for float-flip titles.

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
+++ b/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
@@ -1684,6 +1684,21 @@ internal static unsafe class VulkanVideoPresenter
         return checked((ulong)width * height * bytesPerPixel);
     }
 
+    // Maps a UNORM swapchain format to the sRGB view of the same bit layout,
+    // or Undefined when no counterpart exists. Used to encode linear-float
+    // guest flips on their way into a UNORM swapchain.
+    internal static Format GetSrgbCounterpart(Format format) => format switch
+    {
+        Format.B8G8R8A8Unorm => Format.B8G8R8A8Srgb,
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `fix/float-flip-srgb-present` @ `f7ab28cc` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-448

PR #453, opened 2026-07-19
PerGameSettings Null toggles

kadu04t:
## Description

I had already fixed a similar issue in GuiSettings, and found the same pattern in PerGameSettings. EnvironmentToggles is a List<string>? (nullable). the list itself isn't null, but its elements can be, perGame.EnvironmentToggles = [null], and in that case the ?? operator in Resolve() never kicks in, since the list reference is non-null. So the [null] flows straight into EnvironmentToggles and breaks the same way as before, at Environment.SetEnvironmentVariable.

Since i had already fixed this pattern once, i applied the same fix to PerGameSettings and wrote a test in Tests SharpEmu.Libs.Tests/GUI, which I had already created for the GuiSettings fix.

One difference from the GuiSettings fix, here i didn't normalize a null list into an empty one. A null EnvironmentToggles list has a real meaning in this file, it tells Resolve() to inherit the global setting instead of overriding it. So I only filter out null/empty entries inside an existing list, and leave a null list untouched. 

## Testing

N/A. Dotnet test - all 508 passed

## Checklist

### DIFF
--- a/src/SharpEmu.GUI/PerGameSettings.cs
+++ b/src/SharpEmu.GUI/PerGameSettings.cs
@@ -49,7 +49,7 @@ public sealed class PerGameSettings
             var path = PathFor(titleId);
             if (File.Exists(path))
             {
-                return JsonSerializer.Deserialize<PerGameSettings>(File.ReadAllText(path), SerializerOptions);
+                return NormalizeFromJson(File.ReadAllText(path));
             }
         }
         catch (Exception)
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `pergame-settings-null-toggles` @ `40f9f95a` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-453

PR #455, opened 2026-07-19
[Kernel] Fix event-flag ABBA deadlock, add GC suspend bridge, recover Crunch decoder crash

Bouh:
## Before submitting

- [x] I have read the CONTRIBUTING guidelines.
- [x] My change follows the existing code style of the project.
- [x] I have tested my change locally.

## What does this PR do?

Fixes a set of hangs/crashes hit while getting Cult of the Lamb (PPSA06465) further into its boot sequence under SharpEmu:

- **ABBA lock inversion in `sceKernelWaitEventFlag`**: the wait path held the event flag's own gate while entering the scheduler gate, inverted relative to every wake path (scheduler gate first, then the flag's gate). This intermittently froze the whole scheduler, and the freeze vanished under tracing (which shifts timing), making it look non-reproducible.
- **Pluggable block-waiter describer registry** (`GuestThreadExecution`): each HLE sync primitive (event flag, mutex, semaphore) can now expose its object's state (bits/count/owner/waiters) in guest-thread snapshots, instead of relying on a single hardcoded describer.
- **Enriched mutex profiler** (`SHARPEMU_PROFILE_MUTEX`): owner thread, recursion count and waiter count per hot mutex, plus a time-based dump trigger so a stall (which collapses lock traffic) still produces a dump.
- **Stop-the-world suspend handshake bridge**: Unity's GC coordinator waits on `SuspendSemaphore` for one ack per GC-unsafe thread, but a thread parked in an unrelated HLE wait can never reach its safepoint to post one. A watchdog force-posts the missing acks once the count stalls ~1.5s (`SHARPEMU_SUSPEND_BRIDGE=0` to disable) — a thread blocked in HLE cannot touch the guest heap, so counting it as suspended is safe.
- **`sceHttpUriParse`**: real implementation (was unresolved, causing a null `SceHttpUriElement` dereference and an access violation around t=99s).
- **Crunch decoder null-mip recovery**: Cult of the Lamb's Unity Crunch texture decoder (`DoDeCruncherJob`) writes a decoded entry through a per-mip output pointer read from a local array on its own stack frame, and that slot is null for some textures (a guest-side data/logic issue — confirmed not a cross-thread race, since the array lives on the faulting call's own private stack). The VEH now redirects the write to a scratch buffer and lets the same instruction re-execute, discarding decoded texel data this title never had a real destination for (`SHARPEMU_DISABLE_CRUNCH_NULL_MIP_RECOVERY=1` to disable).
- `/dev/urandom` and `/dev/random` as virtual CSPRNG devices (mono/Unity opens them at boot for entropy).

## Testing

- Reproduced the ABBA deadlock and the GC-suspend stall in Cult of the Lamb before the fix; confirmed both no longer occur across repeated build+run cycles after the fix.
- Root-caused the Crunch decoder crash via Ghidra static analysis of the guest eboot (rebased at `0x800000000`) plus live diagnostic tracing of the faulting register/stack window, then validated the recovery crash-free across 3 heavy-trace runs (with the recovery counter firing and no further access violations at that RIP).
- Full solution build (`dotnet build SharpEmu.slnx -c Release`): clean.
- `dotnet test tests/SharpEmu.Libs.Tests`: 444/444 passing.

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works (existing suite covers the touched HLE surfaces; the Crunch/deadlock fixes were validated via repeated emulator runs against the reproducing title as described above).
- [x] New and existing unit tests pass locally with my changes.
- [x] I have updated the documentation accordingly (n/a — internal HLE/CPU backend behavior, no user-facing docs affected).

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Exceptions.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Exceptions.cs
@@ -128,6 +128,11 @@ public sealed partial class DirectExecutionBackend
 			{
 				return -1;
 			}
+			if (exceptionCode == 3221225477u &&
+				TryRecoverCrunchNullMipOutputWrite(exceptionRecord, contextRecord, rip))
+			{
+				return -1;
+			}
 			if (exceptionCode == StatusIllegalInstruction &&
 				TryRecoverIllegalInstruction(contextRecord, rip))
 			{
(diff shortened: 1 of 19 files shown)


Spooks4576:

Tested `fix/cotl-sync-deadlocks-and-crunch-crash` @ `95e40294` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-455

par274:
Split PR.

PR #420, opened 2026-07-18
[AGC] Implement v_pk_fma_f16 (VOP3P)

Peisr25:
## Before submitting

Please read our contribution guidelines before opening a pull request:

➡️ [**CONTRIBUTING.md**](https://github.com/sharpemu/sharpemu/blob/main/CONTRIBUTING.md)

By opening this pull request, you confirm that you have read and agree to follow the contribution guidelines.

### What this does

Implements `v_pk_fma_f16` (VOP3P). Right now the opcode hits the
unimplemented-emit path. Picks up the follow-up left in #145.

Multiplying the two f16 inputs as f32 gives an exact product, so only the
add rounds. I take the plain sum and recover the exact residual with 2Sum.
When the sum is inexact and lands on an even significand, I round it to odd
(one ulp toward the true value). Then I pack it back to f16 with RNE, which
reproduces the fused result bit for bit.

src2 modifiers (op_sel, neg) reuse the existing path. clamp is still
rejected.

Every op in the chain is marked NoContraction. Without it the AMD driver
contracts the sequence on its own and the double rounding comes back. Ran
into this before adding the marks.

## Testing

No game reached this opcode yet on my setup, so testing was done with the
repo's own tools:

- ShaderDump, new `pk-f16` program: all 5 packed opcodes decode and emit.
  Pinned case `fma(0x4100, 0x7522, 0x04EA) = 0x7A6B`.
- GpuConformance on an RX 7700 XT: pinned case plus its negated twin
  (`0x7A6B7A6B` / `0x7A6A7A6A`, one on each side of the same f16 midpoint).
  All values match.
- 34M cases checked locally against an exact Int128 reference (every finite
  f16 is m·2⁻²⁴, so a·b+c is an exact multiple of 2⁻⁴⁸). 0 mismatches.
- Full repo suite: 234 tests passing.

## Checklist

By submitting this pull request, you confirm that:

- [x] I have read and followed `CONTRIBUTING.md`.
- [x] I tested my changes or marked the testing section as `N/A`.
- [x] I wrote this pull request description myself and did not paste AI-generated explanations.
- [x] I listed the game(s) I tested (or marked the testing section as `N/A`).

PR #460, opened 2026-07-19
[AGC] Support the clamp modifier on packed f16 VOP3P ops

Peisr25:
### What this does
The VOP3P emitter rejected any packed op that had the clamp bit set,
including the `v_pk_fma_f16` from #420. This adds clamp support so those
ops emit instead of failing.

Clamp saturates each f16 output half to [0,1], and NaN becomes 0 (same as
RDNA hardware).

I apply the saturation to each lane's f32 result, before narrowing to f16.
This is bit-exact: 0.0 and 1.0 are both representable in f16 and clamp is
monotonic, so clamping in f32 and then rounding gives the same value as
clamping the f16 result directly. For fma the pre-narrowing value is the
round-to-odd f32, which keeps that property. NaN falls out to 0 through
ordered compares, so no separate IsNan check is needed.

## Testing
No game hits this opcode on my setup, so I verified with the repo tools:
- ShaderDump pk-f16: add and fma with clamp both decode and emit.
- GpuConformance on an RX 7700 XT: pinned fma with clamp saturates to
  `0x3C003C00`, matches.
- Local exact reference: add/mul/fma with clamp against a reference over the
  f16 domain (NaN to 0, range [0,1]), 24M cases plus boundary values,
  0 divergences.

## Checklist

### DIFF
--- a/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
+++ b/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
@@ -1015,12 +1015,6 @@ public static partial class Gen5SpirvTranslator
                 return false;
             }
 
-            if (control.Clamp)
-            {
-                error = $"unsupported vop3p modifiers (clamp) for {instruction.Opcode}";
-                return false;
-            }
-
             var sourceCount = instruction.Opcode == "VPkFmaF16" ? 3 : 2;
             for (var index = 0; index < sourceCount; index++)
             {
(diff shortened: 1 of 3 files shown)


Spooks4576:

Tested `agc-vpk-f16-clamp` @ `87a1ee9f` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-460

PR #461, opened 2026-07-19
kernel: return -1/errno from POSIX file syscalls on failure

### LINKED ISSUE #459: Harvest Days PPSA14677
### Game Name

Harvest Days

### Title ID

PPSA14677

### Game Version

01.002.001

### Current Status

Nothing happens

### Description

Game doesn't boot with following crash error.

### Log File

[log.txt](https://github.com/user-attachments/files/30171729/log.txt)

### Last Milestone / First Error

need to implement hle

### Operating System

Windows 11

### CPU

Intel i7-11800H

### GPU

RTX 3060

### RAM

16 GB

### Emulator Version / Commit

0b83b34

### Renderer

Vulkan

### Additional Notes

I am planning a fix.

slick-daddy:
Fixes #459
## Summary

The POSIX-named file exports `open`, `fstat`, `close`, `read`, and `write`
forwarded the raw `sceKernel*` core result straight to their callers. On
failure those cores report an `OrbisGen2Result` sentinel (`0x8002xxxx`) in
the return value, but libc callers follow the POSIX ABI and expect `-1`
with `errno` set. A non-negative sentinel gets treated as success.

This wraps those five exports so failures return `-1`/`errno`, mirroring the
convention already used by `PosixStat` and `PosixLseek`.

## The crash this fixes

Harvest Days (Unity/IL2CPP) crashed with `0xC0000005` on boot. The IL2CPP
file layer probes `/app0/Media/il2cpp.usym`, a Unity crash-symbol file that
genuinely isn't on disk, so `open` correctly failed with `NOT_FOUND`
(`0x80020002`). But because that was returned as the raw sentinel rather
than `-1`, the guest stored `0x80020002` as a valid fd and passed it back
into `fstat` (visible in the import trace: `fstat(rdi=0x0000000080020002)`).
It kept using the bogus handle and eventually dereferenced a null pointer
(`vmovups xmm0,[rdi]`, `rdi=0`) deep inside

## Changes

- Add a shared `PosixFailure` helper that maps a failed `OrbisGen2Result`
  to the libc convention: set `errno` (`EINVAL`/`EFAULT`/`EACCES`, else
  `EBADF` for fd-based calls or `ENOENT` for path-based), set `RAX` to
  `ulong.MaxValue`, return `-1`.
- Route `PosixStat`, `PosixOpen`, `PosixFstat`, `PosixClose`, `PosixRead`,
  and `PosixWrite` through it. On success these already write the real value
  (fd / byte count) into `RAX`, which the import bridge prefers over the
  return value, so returning `0` is correct.
- The `sceKernel*` siblings are unchanged and still return the raw sentinel.

## Testing
Game: Harvest Days

- Added regression tests for the missing-file `open`, the
  misused-sentinel-fd `fstat`, and bad-fd `close`/`read`/`write`.
- Corrected a socket test that had locked in the old raw-sentinel contract
  for a double `close`.
- Full `SharpEmu.Libs.Tests` suite passes (473 tests); full solution builds
  clean.

## Follow-up (not in this PR)

The same sentinel leak exists in the sibling POSIX exports in
`KernelFileExtendedExports.cs` (`pread`, `pwrite`, `ftruncate`, `truncate`,
`fsync`, `rename`, `dup`, `dup2`, `fcntl`). They aren't on this crash path,
so they're intentionally left for a separate PR using the same
`PosixFailure` wrapper.
## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelExports.cs
@@ -362,7 +362,7 @@ public static class KernelExports
         ExportName = "open",
         Target = Generation.Gen4 | Generation.Gen5,
         LibraryName = "libKernel")]
-    public static int Open(CpuContext ctx) => KernelMemoryCompatExports.KernelOpenUnderscore(ctx);
+    public static int Open(CpuContext ctx) => KernelMemoryCompatExports.PosixOpen(ctx);
 
     [SysAbiExport(
         Nid = "1G3lF1Gg1k8",
@@ -376,7 +376,7 @@ public static class KernelExports
         ExportName = "fstat",
         Target = Generation.Gen4 | Generation.Gen5,
         LibraryName = "libc")]
-    public static int Fstat(CpuContext ctx) => KernelMemoryCompatExports.KernelFstat(ctx);
+    public static int Fstat(CpuContext ctx) => KernelMemoryCompatExports.PosixFstat(ctx);
 
     [SysAbiExport(
         Nid = "hcuQgD53UxM",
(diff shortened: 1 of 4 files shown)


Spooks4576:

Tested `harvest-days` @ `218db46e` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-461

PR #464, opened 2026-07-20
fix(cpu): implement native exception pre-filtering and restore native worker threads

### LINKED ISSUE #212: ReversePInvokeBadTransition fail-fast on the pooled native guest worker (UnmanagedCallersOnly fatal)
### Emulator Version / Commit

`7b86a91` (current main), Release build, Windows 10, Ryzen 5 5600X / RTX 2060 / 16 GB.

### Bug Description

The runtime fail-fasts with a CLR fatal while a game is still booting:

```
Fatal error.
Invalid Program: attempted to call a UnmanagedCallersOnly method from managed code.
```

Despite the wording this is not managed code calling `RunPrologue` directly. It is the `ReversePInvokeBadTransition` fail-fast the comment in `DirectExecutionBackend.cs` already describes: the worker enters `RunPrologue` while the thread is already in cooperative mode. It looks like the same class of bug as #81, on a different path.

**Where it happens.** Astro Bot (PPSA21564 v01.007.000), right as the engine spawns its `WebApiJobWorker` guest threads. Those are created back to back and are the only threads with `entry=0x8002B9400` (every other guest thread uses `0x800001090`), `priority=0`, `host_priority=Highest`.

**What the built-in tracer says.** With `SHARPEMU_LOG_THREAD_MODE=1`:

```
[THREADMODE] entry_setup name='WebApiJobWorker' entry=0x00000008002B9400 stub=0x000001CEB4F70000 cycle=39 tid=17324 managed=16
[THREADMODE] worker_enter guest=0x000001CEB6214B10 stub=0x000001CEB4F70000 cycle=0 tid=28176 managed=23
[THREADMODE] worker_exit  guest=0x000001CEB6214B10 result=0x00000000 yield=True cycle=0 tid=28176 managed=23
[THREADMODE] cycle_end   name='WebApiJobWorker' state=Blocked imports=2 cycle=39 tid=17324 managed=16
[THREADMODE] cycle_start name='WebApiJobWorker' guest=0x000001CEB6215B20 reason=pthread_create cycle=40 tid=25404 managed=13
[THREADMODE] entry_setup name='WebApiJobWorker' entry=0x00000008002B9400 stub=0x000001CEB4F70000 cycle=40 tid=25404 managed=13
[THREADMODE] worker_enter guest=0x000001CEB6215B20 stub=0x000001CEB4F70000 cycle=0 tid=28176 managed=23
[THREADMODE] worker_exit  guest=0x000001CEB6215B20 result=0x00000000 yield=True cycle=0 tid=28176 managed=23
[THREADMODE] cycle_end   name='WebApiJobWorker' state=Blocked imports=2 cycle=40 tid=25404 managed=13
[THREADMODE] cycle_start name='WebApiJobWorker' guest=0x000001CEB6212AF0 reason=pthread_create cycle=41 tid=21820 managed=12
Fatal error.
Invalid Program: attempted to call a UnmanagedCallersOnly method from managed code.
```

Every cycle runs through the same pooled native worker (`tid=28176`, `managed=23`, so it is CLR-attached), driven by a different orchestrator thread each time (`managed=16`, `13`, `12`). It dies on cycle 41. The historical symptom in the comments was cycle-count dependent too (the audio_output_thread "7th cycle" crash), so this may be the same shape.

**What I ruled out**, by instrumenting `RunGuestEntryStub` locally:

- Not the inline fallback. `RentNativeGuestExecutor()` never returns null in a normal run, so `CallNativeEntry` is never reached. Zero occurrences.
- Not worker creation failure. The "Failed to create a native guest worker thread" warning never fires.
- Not the teardown race documented around `_guestTeardownRequested`. No teardown anywhere near the crash.
- Not the `_workAvailable` / `_workCompleted` handshake as far as I can tell. The emitted loop does `RunEpilogue` then `SetEvent`, in that order, and the `AutoResetEvent` latches a Set that arrives while the worker is still unwinding.

So the worker path is being used correctly and it still ends up in a bad transition. I could not get further than that without a debugger attached to the worker thread at the transition, and I would rather report it than guess at a fix in a race.

**Useful workaround for anyone debugging this:** `SHARPEMU_DISABLE_NATIVE_GUEST_WORKERS=1` forces the inline path, which crashes the same way but on a CLR thread, so you get a real managed stack trace:

```
   at SharpEmu.Core.Cpu.Native.DirectExecutionBackend.CallNativeEntry(Void*)
   at SharpEmu.Core.Cpu.Native.DirectExecutionBackend.RunGuestEntryStub(Void*, UInt64)
   at SharpEmu.Core.Cpu.Native.DirectExecutionBackend.ExecuteEntry(...)
   at SharpEmu.Core.Cpu.Native.DirectExecutionBackend.TryExecute(...)
   at SharpEmu.Core.Cpu.CpuDispatcher.DispatchEntryCore(...)
```

**Reproducing.** One caveat worth being upfront about: on plain `main` Astro Bot dies earlier, at its own engine assert, because the save data memory family (`sceSaveDataSetupSaveDataMemory2` and friends) is unresolved. You need those implemented to get far enough to hit this. I have that working locally and am happy to open a PR for it if it helps make this reproducible, just say the word.

### Log File

Full `SHARPEMU_LOG_THREAD_MODE=1` run available on request (1.4 MB). The excerpt above is the tail, which is the part that matters.

aadi-varc:
Fixes #212. Implements native pre-filtering in CreateExceptionHandlerTrampoline to bypass managed code for non-managed exceptions and restores native worker threads.

### DIFF
--- a/src/SharpEmu.CLI/Program.cs
+++ b/src/SharpEmu.CLI/Program.cs
@@ -607,7 +607,7 @@ internal static partial class Program
             nint jobHandle = 0;
             Environment.SetEnvironmentVariable(MitigatedChildEnvironment, "1");
             var created = CreateProcessW(
-                processPath,
+                null,
                 cmdLineBuilder,
                 0,
                 0,
@@ -1433,7 +1433,7 @@ internal static partial class Program
     [DllImport("kernel32.dll", EntryPoint = "CreateProcessW", SetLastError = true, CharSet = CharSet.Unicode)]
     [return: MarshalAs(UnmanagedType.Bool)]
     private static extern bool CreateProcessW(
-        string applicationName,
+        string? applicationName,
         StringBuilder commandLine,
         nint processAttributes,
         nint threadAttributes,
--- a/src/SharpEmu.CLI/SharpEmu.CLI.csproj
+++ b/src/SharpEmu.CLI/SharpEmu.CLI.csproj
@@ -32,6 +32,7 @@ SPDX-License-Identifier: GPL-2.0-or-later
          that collision and keeps the emulator's required mappings available. -->
     <ServerGarbageCollection>false</ServerGarbageCollection>
     <ConcurrentGarbageCollection>true</ConcurrentGarbageCollection>
+    <GCRetainVM>false</GCRetainVM>
     <TieredPGO>true</TieredPGO>
   </PropertyGroup>
 
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -2414,6 +2414,42 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 
 		byte* code = (byte*)ptr;
 		int offset = 0;
+
+		if (OperatingSystem.IsWindows())
+		{
+			// Native pre-filter: these exception codes are raised while the thread can be in
+			// cooperative GC mode (a C# throw is RaiseException(0xE0434352) on the throwing
+			// thread; FailFast/stack-overflow arrive mid-runtime-failure). Entering the managed
+			// handler then trips the CLR's reverse-P/Invoke check and kills the process with
+			// "Invalid Program: attempted to call a UnmanagedCallersOnly method from managed
+			// code" — this is why no managed throw (even one with a catch handler) ever
+			// survived inside the emulator. Continue the handler search without touching
+			// managed code; the CLR's own VEH handles its exceptions. MSVC C++ exceptions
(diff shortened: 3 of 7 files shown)


Spooks4576:

Tested `fix/reverse-pinvoke-crash` @ `5a60631a` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-464

par274:
This PR does not meet our contribution guidelines and has been closed. Please review the CONTRIBUTING.md and submit a new PR that follows the project's requirements.

PR #465, opened 2026-07-20
Shader: test a wave mask consumed as a per-lane predicate at the lane bit

xnetcat:
## Summary

A VCC/EXEC wave mask consumed as a **per-lane predicate** (the condition of `V_CNDMASK_B32`, a VCC/EXEC branch, or the derived `_vcc` / `_exec` bool) was tested in single-lane emulation with a whole-word non-zero test (`IsNotZero64(mask)`) instead of the **current lane's bit**.

That is correct for comparison results — a compare only ever sets the lane's own bit — so the single-lane path historically took the cheaper whole-word test. But it is wrong for the **bitwise-complement wave-mask idioms** (`S_NOT` / `S_ORN2` / `S_ANDN2` / `S_NAND` / `S_NOR` on a 64-bit mask), which set the unused upper 63 bits. A whole-word non-zero test then reports the lane **active even when this lane's bit is clear**.

The fix makes `IsWaveMaskActive` always extract the current lane's bit (`mask & lane_bit`) in both single-lane and subgroup modes, exactly as the hardware evaluates a per-lane predicate.

## Root cause / evidence

Unity's built-in PostProcessing **NaN killer** pass does exactly this idiom. Per channel it computes `isNaN = NLT AND NGT AND NEQ` (against 0), then combines the channels as `anyNaN OR NOT(v3-is-finite)` — lowered to **`S_ORN2_B64`**. The bitwise complement set the upper mask bits, so under the whole-word test **every valid pixel read as NaN** and was replaced with `0`, zeroing the whole HDR scene before Bloom/Uber/tone-mapping.

Observed in Superliminal (PPSA06084): the 3D scene rendered **black behind the menu** while the UI (which never hits this pass) survived. Render-graph tracing showed a healthy HDR scene buffer (mean ≈ 67) collapse to **exact 0** across the NaN-killer's single-texture-in / HDR-out pass, then propagate black through every downstream pass. With this fix the storage room renders behind the menu with natural exposure and no forced values.

## Why it is safe / no-op for unaffected shaders

In single-lane mode `CurrentLaneBit()` is the constant `1` (bit 0), so `IsCurrentLaneSet(mask)` reduces to `(mask & 1) != 0`. For any wave mask produced by a compare — where only bit 0 is ever set — this is **byte-identical** to the previous `IsNotZero64(mask)`. Only the complement idioms (which set the upper 63 bits) change behavior, and there the new result matches the hardware. The subgroup path already used `IsCurrentLaneSet`, so it is unchanged.

## Test

Added `Gen5WaveMaskSpirvTests.WaveMaskPredicate_IsTestedAtCurrentLaneBit`: it compiles a shader that writes VCC at run time (`V_CMP_EQ_F32`) and asserts the emitted SPIR-V tests the wave mask at the current lane bit (`mask & lane_bit`) rather than as a whole-word non-zero test. The test fails against the previous `IsNotZero64(mask)` path and passes with the fix.

Diff: `Gen5SpirvTranslator.cs` (13 insertions, 3 deletions) plus the focused regression test. `dotnet build` clean; `SharpEmu.Libs.Tests` 471/471 green.

### DIFF
--- a/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.cs
+++ b/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.cs
@@ -5290,10 +5290,20 @@ public static partial class Gen5SpirvTranslator
                 UInt(0x108));
         }
 
+        // A wave-mask SGPR (VCC/EXEC) consumed as a per-lane predicate — the
+        // condition of VCndmask, a VCC/EXEC branch, or the derived _vcc/_exec
+        // bool — must be tested at the CURRENT lane's bit, exactly as the
+        // hardware does, not as "the 64-bit value is non-zero". The two coincide
+        // for comparison results (only the lane's own bit is ever set), so the
+        // single-lane path historically used a cheaper whole-word non-zero test.
+        // But bitwise-complement wave-mask idioms (S_NOT/S_ORN2/S_ANDN2/S_NAND/
+        // S_NOR on a 64-bit mask) set the unused upper 63 bits; a whole-word test
+        // then reports "lane active" even when this lane's bit is clear. Unity's
+        // PostProcessing NaN killer does exactly this (`anyNaN | ~allFinite`),
+        // which made every valid pixel read as NaN and get replaced with 0 —
+        // zeroing the whole scene before tonemap. Extract the lane bit always.
         private uint IsWaveMaskActive(uint mask) =>
-            _subgroupInvocationIdInput == 0
-                ? IsNotZero64(mask)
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `fix/wave-mask-lane-bit` @ `530f85bd` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-465

PR #466, opened 2026-07-20
Shader: lower VOP3P V_FMA_MIX_F32/LO/HI (was dropping Unity HDR shaders)

xnetcat:
## Problem

The VOP3P decoder recognises the three mix opcodes — `0x20` **V_FMA_MIX_F32**, `0x21` **V_FMA_MIXLO_F16**, `0x22` **V_FMA_MIXHI_F16** — but leaves them opaque as `Vop3pRaw20/21/22`. At SPIR-V emission they fall through the vector-ALU switch to the `default` case and fail with `unsupported vector opcode Vop3pRaw20`. Because a single unhandled instruction fails the whole compile, **any shader that uses `fma_mix` is dropped entirely**.

Unity's built-in-RP / PostProcessing Stack v2 HDR, tone-mapping and auto-exposure passes emit `V_FMA_MIX_F32`, so those shaders never translated. This is what kept Superliminal's auto-exposure luminance chain from running.

## Root cause

Decoded-but-not-lowered → `default` error → `return false` → shader dropped. The decode side already handles VOP3P (opcode `0x33`) and trailing literals; only the emission-side lowering for the mix ops was missing.

## Fix

- Name `0x20/0x21/0x22` in `DecodeVop3p` (`VFmaMixF32` / `VFmaMixloF16` / `VFmaMixhiF16`), the same way the packed `v_pk_*` ops are named.
- Lower them in `Gen5SpirvTranslator.Alu.cs`. Each mix op computes a single **f32** `fma(a, b, c)` where every source is read *independently* as either a full f32 register/constant or one f16 half widened to f32:
  - `op_sel_hi[i]` selects **f16 vs f32** for operand `i`;
  - `op_sel[i]` picks **which f16 half** feeds the f32;
  - `neg_hi[i]` is repurposed as the **absolute-value** modifier and `neg[i]` negates, applied **abs-then-neg**.
  
  This reuses the VOP3P `op_sel`/`op_sel_hi`/`neg`/`neg_hi` bit layout with the **mix-specific meaning**, not the packed-math (low/high-lane) meaning.
- `V_FMA_MIX_F32` writes the scalar f32; `_MIXLO`/`_MIXHI` narrow the f32 result back to f16 (exact round-to-nearest-even via the existing `EmitFloatToHalf`) and write it into the low/high 16 bits of `vdst`, preserving the other half. The clamp modifier saturates to `[0, 1]` consistently with the other VOP3P ops.

The per-operand F16/F32 select and the abs/neg modifiers follow **shadPS4's `GetSrcMix`**, the authoritative reference for the mix semantics.

## Tests

Adds `Gen5FmaMixSpirvTests`: assembles `V_FMA_MIX_F32` (with a representative `op_sel`/`op_sel_hi`/`neg`/`abs`) and `V_FMA_MIXLO_F16` compute shaders and asserts they translate to GPU SPIR-V **without hitting the drop path** and emit a GLSL.std.450 `Fma` (and an `FAbs` for the `neg_hi` modifier). Both tests fail against the pre-fix tree with `unsupported vector opcode Vop3pRaw20/21`.

- Full solution build: 0 errors, 0 warnings.
- `SharpEmu.Libs.Tests`: 472/472 green (470 baseline + 2 new).

## Failure mode fixed

HDR / tone-mapping / auto-exposure shaders (any pass using `V_FMA_MIX_F32`) now translate instead of being dropped. Leaving cross-title validation to the gameplay bot.

### DIFF
--- a/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
+++ b/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
@@ -960,6 +960,15 @@ public static partial class Gen5SpirvTranslator
                     }
 
                     break;
+                case "VFmaMixF32":
+                case "VFmaMixloF16":
+                case "VFmaMixhiF16":
+                    if (!TryEmitFmaMix(instruction, destination, out result, out error))
+                    {
+                        return false;
+                    }
+
+                    break;
                 default:
                     error = $"unsupported vector opcode {instruction.Opcode}";
                     return false;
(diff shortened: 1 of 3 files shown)


Spooks4576:

Tested `fix/vop3p-fma-mix` @ `641393cc` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-466

PR #474, opened 2026-07-20
fix: roll back earlier host allocations on later gap failure in TryBackFixedRange (#472)

### LINKED ISSUE #472: [Memory] Roll back partial fixed mappings when a later gap cannot be backed
## Summary

`PhysicalVirtualMemory.TryBackFixedRange` can leave a partially applied fixed mapping when a requested range contains multiple free runs and backing a later run fails.

## Reproduction

At upstream `25d741b35b93f7b6a818c6156517afbca6665321`, use a synthetic host-memory implementation that reports this page layout:

```text
committed | free | committed | free
   0x1000   0x1000   0x1000   0x1000
```

Make the first free-run allocation succeed and the second one fail. `TryBackFixedRange` returns `false`, but the successful first allocation is not freed. In the current implementation, `PhysicalVirtualMemory.cs:453-506` allocates and inserts each gap as it walks; the failure returns at lines 455 or 478 without undoing earlier allocations or region bookkeeping.

A focused synthetic regression test (`GuestMemoryAllocatorTests.TryBackFixedRangeRollsBackEarlierGapsWhenLaterGapCannotBeBacked`) reproduced this locally:

- unpatched upstream behavior: expected `Free(firstGap)` but observed no `Free` call;
- transactional local fix: focused test passed, along with all 8 `GuestMemoryAllocatorTests`.

## Impact

The caller treats `false` as failure and may attempt another allocation path, while the host mapping and `MemoryRegion` entry from an earlier gap remain. That can leak address space and leave the guest/internal allocator state inconsistent after a failed fixed mapping.

## Suggested direction

Stage successful host allocations while walking the range, free all staged allocations on any later query/allocation failure, and insert `MemoryRegion` entries only after every free sub-range has been backed successfully. A fully occupied range should continue to return `false` without changing state.

This report uses only synthetic fixtures and public source. No firmware, game data, proprietary code, or non-public reverse-engineering material was used. I did not open a competing PR because active PR #399 also changes `PhysicalVirtualMemory.cs` and `GuestMemoryAllocatorTests.cs`, although for a separate reservation issue.

kostyaff:
## Summary

`TryBackFixedRange` leaks earlier host allocations when a later free gap in the range cannot be backed. This change stages all successful host allocations during the range walk and inserts `MemoryRegion` entries only after **every** gap has been backed. On any failure, all staged allocations are freed and no region is inserted — the address space is left untouched.

## Problem

When a fixed mapping request spans multiple free runs, `TryBackFixedRange` previously allocated and inserted `MemoryRegion` entries gap-by-gap. If the second gap's `Allocate` failed, the first gap's host mapping and `MemoryRegion` entry remained — leaking address space and leaving guest/internal allocator state inconsistent.

## Changes

**`src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs`** — rewritten `TryBackFixedRange` as a transactional operation:

- **Staging:** successful host allocations are collected into `stagedAllocations` during the walk; no `MemoryRegion` is inserted yet
- **Commit:** if every gap backs successfully, all `MemoryRegion` entries are inserted in one batch under the write lock
- **Rollback:** if any gap fails (`Query` false, `runEnd <= cursor`, `Allocate` returned a different address), all staged allocations are freed via `_hostMemory.Free(gapAddress)`, no regions are inserted, and `false` is returned
- A fully occupied range still returns `false` with zero side effects

**`tests/SharpEmu.Libs.Tests/Memory/GuestMemoryAllocatorTests.cs`** — new test double `GappedHostMemory` and regression test:

- `GappedHostMemory` models a 4-page layout `[committed | free | committed | free]`: the first free gap's `Allocate` succeeds, the second returns `0`
- `TryBackFixedRangeRollsBackEarlierGapsWhenLaterGapCannotBeBacked` asserts that the first gap is allocated, the second is attempted, and the first gap is freed during rollback — nothing leaks

## Testing

- `dotnet build SharpEmu.slnx -c Release --no-restore` — 0 errors, 0 warnings
- `dotnet test --filter FullyQualifiedName~GuestMemoryAllocator` — all 8 tests pass (7 existing + 1 new)

Fixes #472

### DIFF
--- a/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
+++ b/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
@@ -446,13 +446,20 @@ public sealed unsafe class PhysicalVirtualMemory : IVirtualMemory, IGuestMemoryA
         // us over whole free or occupied stretches. Only free stretches get backed;
         // stretches already reserved or committed by another allocation are left as
         // they are, which is exactly what a fixed mapping does on hardware.
+        //
+        // Because backing may span several disjoint free runs, allocations are
+        // staged: host pages are reserved/committed first, and the corresponding
+        // MemoryRegions are inserted only once every gap in the range has been
+        // backed. If any gap fails to back, every earlier host allocation is freed
+        // and no region is inserted, so the address space is left untouched.
+        var stagedAllocations = new List<(ulong Address, ulong Size)>();
+
         var cursor = start;
-        var backedAny = false;
         while (cursor < end)
         {
             if (!_hostMemory.Query(cursor, out var info))
             {
-                return false;
+                goto Rollback;
             }
 
             var queriedEnd = info.RegionSize > ulong.MaxValue - info.BaseAddress
@@ -461,7 +468,7 @@ public sealed unsafe class PhysicalVirtualMemory : IVirtualMemory, IGuestMemoryA
             var runEnd = Math.Min(end, queriedEnd);
             if (runEnd <= cursor)
             {
-                return false;
+                goto Rollback;
             }
 
             if (info.State == HostRegionState.Free)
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `fix/trybackfixedrange-rollback-472` @ `501e1a93` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-474

PR #475, opened 2026-07-20
[VideoPresenter] Accept padded row pitch in guest image uploads

NickPom:
## What this fixes

Dead Cells (PPSA15552) freezes during a loading transition. The guest
sends texture upload data whose row length is padded to a hardware
alignment wider than the image width. UploadGuestImageInitialData only
accepted uploads whose byte count matched the tightly packed
width * height * bytes-per-pixel exactly, so these padded uploads were
rejected and the texture stayed blank, stalling the pipeline behind it.

## What this does

Added a check that recovers the real row length when the byte count
matches a common padding alignment (8/16/32/64/128/256 texels) and
passes it as BufferRowLength on the Vulkan copy, instead of hardcoding
0. Anything that doesn't match a known padding alignment is still
rejected as before.

## Testing

- Dead Cells (PPSA15552): before the fix, draw calls froze at a fixed
  count with the upload rejected in the trace log. After the fix, the
  game proceeds past the load screen and stays stable past 1,000,000
  draw calls with no rejected uploads.
- Dreaming Sarah (PPSA02929): unaffected, still renders normally,
  confirming the tightly packed (non-padded) upload path is unchanged.

## Checklist

- [x] I tested my changes (see Testing above).
- [x] I listed the game(s) I tested.

### DIFF
--- a/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
+++ b/src/SharpEmu.Libs/VideoOut/VulkanVideoPresenter.cs
@@ -11097,6 +11097,65 @@ internal static unsafe class VulkanVideoPresenter
             target.Initialized = true;
         }
 
+        // Returns the source row length in texels when the upload is a linear
+        // image whose rows are padded to a wider hardware pitch, or 0 when the
+        // data is an exact tightly packed match or not a recognisable padded
+        // layout (in which case the caller keeps rejecting it). Only a pitch
+        // equal to the width rounded up to a common alignment is accepted, so an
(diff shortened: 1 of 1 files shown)


Spooks4576:

Tested `fix/dead-cells-padded-texture-upload` @ `f22a0ec5` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-475

PR #478, opened 2026-07-20
Kernel FS: close guest→host sandbox escapes in the path resolver

slick-daddy:
## Summary

`KernelMemoryCompatExports.ResolveGuestPath` maps every guest file syscall
(`open`, `stat`, `unlink`, `mkdir`, `rmdir`, `rename`, `chmod`, APR resolve)
onto a host path, and callers pass that path straight to `FileStream` /
`File.*` / `Directory.*`. A security audit found the resolver was
**default-allow**: several classes of guest path escaped the mount sandbox
and reached arbitrary host files. Since guest paths originate from an
untrusted game dump, these are attacker-controlled.

This PR closes the escapes and hardens the resolver to fail closed.

## Vulnerabilities fixed

1. **Absolute-path passthrough (critical).** Any guest path matching no mount
   prefix was returned verbatim as a host path, so `open("/etc/passwd")` or
   `open("C:\Windows\System32\drivers\etc\hosts")` hit the real host file.
   The resolver now default-denies: unmapped paths resolve to empty, which
   callers map to `NOT_FOUND`.

2. **Windows drive-letter injection (critical).** The built-in mount branches
   combined the guest tail onto the mount root with `Path.Combine`, but a
   surviving `C:` segment made `Path.Combine` discard the root and yield a raw
   host path (`download0/C:/Windows/Temp/evil.dll` → arbitrary host write).
   All built-in branches now route through `CombineWithinMount`, which
   re-resolves with `Path.GetFullPath` and re-checks containment — the same
   guard the registered-mount path already used.

3. **Symlink/junction escape (high).** Lexical containment does not follow
   reparse points, so a dump that plants `link -> /` inside a mount let
   `open("/app0/link/etc/passwd")` escape. `EscapesMountViaReparsePoint` now
   walks each existing component below the mount root and rejects reparse
   points, mirroring the existing `AvPlayerExports` model.

## Hardening

- The containment guards call `Path.GetFullPath` / `File.GetAttributes` on
  untrusted input, which throw on crafted over-long or invalid-char paths.
  Because `ResolveGuestPath` runs outside the syscalls' try blocks, that was a
  guest-triggerable crash. All three resolution paths now **fail closed**:
  an access/format failure is treated as a denial, never propagated.
- The three syscalls whose catch blocks don't cover `ArgumentException`
  (`open`, `truncate`, `rename`) get an explicit empty-path guard so a denied
  path returns `NOT_FOUND` instead of throwing.
## Testing

No games tested.

- New `KernelSandboxEscapeTests` covering absolute-path denial, drive-letter
  injection, reparse-point escape, malformed-path fail-closed, plus positive
  tests confirming legitimate deep/nested mount access still resolves.
- Updated `AprStreamingContractTests`, which previously relied on the
  absolute-path passthrough being fixed — it now addresses the file through a
  registered mount, as a real guest would.

## Verification

- Build clean on .NET 10.
- Full `SharpEmu.Libs.Tests` suite passes (494 tests).

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelFileExtendedExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelFileExtendedExports.cs
@@ -267,6 +267,11 @@ public static partial class KernelMemoryCompatExports
         }
 
         var hostPath = ResolveGuestPath(guestPath);
+        if (string.IsNullOrEmpty(hostPath))
+        {
+            return (int)OrbisGen2Result.ORBIS_GEN2_ERROR_NOT_FOUND;
+        }
+
         try
         {
             using var stream = new FileStream(hostPath, FileMode.Open, FileAccess.Write, FileShare.ReadWrite);
@@ -310,6 +315,11 @@ public static partial class KernelMemoryCompatExports
 
         var fromHost = ResolveGuestPath(fromGuest);
         var toHost = ResolveGuestPath(toGuest);
+        if (string.IsNullOrEmpty(fromHost) || string.IsNullOrEmpty(toHost))
+        {
+            return (int)OrbisGen2Result.ORBIS_GEN2_ERROR_NOT_FOUND;
+        }
+
         try
         {
             if (Directory.Exists(fromHost))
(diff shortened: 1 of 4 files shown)


Spooks4576:

Tested `opsec` @ `8cd4905d` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-478

PR #482, opened 2026-07-20
[CPU] Recover SSE4a EXTRQ/INSERTQ faults on Linux via the POSIX XMM bridge

999sian:
## Summary

The general SSE4a fault-time fallback (`DirectExecutionBackend.Amd64Compat.cs`) only ran on Windows. On POSIX the EXTRQ/INSERTQ half was deliberately disabled because the signal bridge populated the CONTEXT scratch buffer with the 17 general-purpose registers only — emulating an XMM instruction there would compute a result from zeroed bytes and silently discard the write. So a guest EXTRQ/INSERTQ that the load-time `Sse4aExtrqBlendPatch` idiom does not cover still kills the process with an unrecovered SIGILL on any Linux host without SSE4a, i.e. every Intel CPU (#328, Dead Cells PPSA15552).

This bridges the XMM registers on Linux:

- `TryHandlePosixFault` now copies XMM0–15 from the mcontext's FXSAVE image into the CONTEXT `FltSave` slots before running the recovery chain, and copies them back on a recovered fault so `sigreturn` restores the emulated result. The fpstate pointer (gregs+184) and the XMM offset inside the FXSAVE image (+160) are kernel sigcontext ABI; glibc and musl overlay the same layout, and the FXSAVE image is byte-compatible with the Win64 CONTEXT `FltSave` area, so the copy is a straight 256-byte block both ways.
- The EXTRQ/INSERTQ recovery is gated on that bridge (`Windows || XMM state bridged`) instead of on Windows. A frame that carries no fpstate declines exactly as before, and Darwin keeps declining until its XMM area is bridged too.

MONITORX/MWAITX behavior is unchanged (already worked on every OS), Windows is untouched, and the load-time patch still avoids the fault-and-recover cost on the hot idiom it was built for.

## Testing

- `dotnet build SharpEmu.slnx -c Release`
- `dotnet test SharpEmu.slnx -c Release` — 558 passed, including three new tests that fabricate the exact Linux signal frame (kernel-layout gregs + FXSAVE image) and drive `TryHandlePosixFault` over real EXTRQ/INSERTQ encodings: XMM state is captured, emulated, written back, RIP advances, and a frame without fpstate declines.
- Standalone harness on a real SSE4a-less host (Intel Meteor Lake, Arch Linux): executing `extrq xmm0, 0x10, 0x08` and `insertq xmm0, xmm1, 0x10, 0x08` from raw code pages raises live SIGILLs that this recovery resumes with bit-exact results (verified against the AMD64-manual semantics from `Sse4aBitFieldEmulator`, on both a dedicated thread and a threadpool thread; on unmodified main the same harness dies on the first EXTRQ). The harness occasionally crashes while *constructing* the backend, but it does so identically on unmodified main under heap churn, so that instability pre-exists and is unrelated to this change.
- Real-game testing was not performed (no PS5 dumps here). Dead Cells on an Intel Linux host is the scenario this targets; the gameplay bot covers the no-regression side.

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Amd64Compat.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Amd64Compat.cs
@@ -51,12 +51,13 @@ public sealed partial class DirectExecutionBackend
         // round-trips through the real ucontext, so it works on every supported OS. EXTRQ/
         // INSERTQ additionally read and write an XMM register: on Windows contextRecord is the
         // live CONTEXT the OS resumes the thread from, so touching the Xmm0.. slots is visible
-        // to the guest, but on POSIX contextRecord is a CONTEXT-shaped scratch buffer that the
-        // bridge only populates with the 17 general-purpose registers - the XMM region is never
-        // read from or written back to the real mcontext/ucontext. Running this on POSIX would
-        // silently compute a result from stale/zeroed XMM bytes and then discard whatever it
-        // "wrote", so keep it Windows-only, matching Kyty's own scope for the identical fix.
-        return OperatingSystem.IsWindows() && TryRecoverSse4aExtractInsert(contextRecord, rip);
+        // to the guest, and on Linux the bridge copies the mcontext's FXSAVE image into the
+        // Xmm0.. slots and writes them back through sigreturn (_posixXmmContextBridged). On
+        // Darwin the XMM area is still a zeroed scratch buffer - running this there would
+        // silently compute a result from stale bytes and then discard whatever it "wrote", so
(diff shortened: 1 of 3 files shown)


Spooks4576:

Tested `linux/sse4a-posix-xmm-bridge` @ `bf7e8560` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-482

par274:
Are you test this Linux?

999sian:
Yes the dev machine is Linux.

PR #484, opened 2026-07-21
[Shader] Reuse zero storage for unbound buffer fallbacks

Dafenxz0:
## Before submitting

I read `CONTRIBUTING.md` and kept this change focused on one shader-evaluation fallback.

## What changed

A decoded buffer descriptor can temporarily point at memory SharpEmu cannot read. Non-strict mode intentionally keeps the pass alive by binding zero-filled storage, but that path allocated a fresh managed array on every evaluation. One fallback could allocate up to 16 MiB, and the same warning was formatted and written for every draw.

This PR:

- rents synthetic zero buffers from the existing bounded guest transfer pool;
- clears the active range before every use, so data from an earlier binding cannot leak;
- keeps synthetic storage shader-writable while explicitly disabling guest-memory writeback;
- applies the same ownership rules to null descriptors; and
- reports an unreadable descriptor once per shader instruction instead of once per draw.

This is not a persistent guest-buffer cache. Valid descriptors are still read from guest memory for every evaluation, so CPU writes cannot be hidden by stale cached contents.

## Why

The fallback is useful for optional resources and temporarily stale material descriptors, but it should be cheap after warmup. Repeated large `byte[]` allocations feed the LOH/GC and can become expensive in draw-heavy scenes. Pooling only the synthetic storage removes that pressure without changing which guest bytes a valid binding observes.

## Testing

- `dotnet restore SharpEmu.slnx --locked-mode`
- `dotnet build SharpEmu.slnx -c Release --no-restore` — 0 warnings, 0 errors
- `SharpEmu.Libs.Tests` — 512 passed
- `SharpEmu.SourceGenerators.Tests` — 33 passed
- `SharpEmu.ShaderCompiler.Metal.Tests` — 27 passed
- Added a regression test that repeatedly evaluates an unmapped 1 MiB descriptor. It dirties the returned array before reuse, verifies the next binding is fully zeroed and non-writeback, and keeps twelve warmed evaluations below 1 MiB of managed allocation.
- Real-game testing: N/A on my hardware. The automated Dead Cells/Dreaming Sarah run should provide the first gameplay regression check; an allocation-rate A/B in a title that exercises this fallback would be useful.

## Checklist

- [x] I tested my changes as described above.
- [x] I reviewed and wrote the final pull request description myself.
- [x] I listed game testing as N/A and stated the validation still needed.

### DIFF
--- a/src/SharpEmu.ShaderCompiler/Gen5ShaderScalarEvaluator.cs
+++ b/src/SharpEmu.ShaderCompiler/Gen5ShaderScalarEvaluator.cs
@@ -33,8 +33,9 @@ public static class Gen5ShaderScalarEvaluator
             Environment.GetEnvironmentVariable("SHARPEMU_STRICT_BUFFER_LOAD"),
             "1",
             StringComparison.Ordinal);
-    private static readonly object _scalarFallbackTraceGate = new();
+    private static readonly object _fallbackTraceGate = new();
     private static readonly HashSet<(ulong Shader, uint Pc)> _tracedScalarFallbacks = [];
+    private static readonly HashSet<(ulong Shader, uint Pc)> _tracedBufferFallbacks = [];
 
     // Uniform forward branches select material/resource bodies that remain
     // statically present in the translated shader. Discover the skipped body's
@@ -457,11 +458,12 @@ public static class Gen5ShaderScalarEvaluator
                             bufferMemory.ScalarResource,
                             0,
                             new List<uint> { instruction.Pc },
-                            new byte[sizeof(uint)],
+                            RentZeroedGlobalMemory(sizeof(uint)),
                             sizeof(uint),
-                            DataPooled: false)
+                            DataPooled: true)
                         {
                             Writable = writable,
+                            WriteBackToGuest = false,
                         };
                         globalMemoryByAddress.Add(nullKey, binding);
                         globalMemoryBindings.Add(binding);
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `perf/pool-unbound-shader-buffers` @ `b92f99d5` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-484

kostyaff (approved):
The pooling approach is sound. `GuestDataPool.Shared` is bounded (256 MiB, 8 arrays per bucket), so repeated Rent/Return cycles stay in a fixed memory footprint — verified in `ReturnPooledEvaluationArrays` that all evaluation paths do return pooled arrays.

Data clearing in `RentZeroedGlobalMemory` is correct: it clears exactly `dataLength` bytes, matching what `TryReadGlobalMemory` will actually read (the shrinking loop guarantees `dataLength == candidateSize`). Tail bytes beyond that aren't consumed, so partial clearing is safe.

`WriteBackToGuest = false` on the fallback preserves the previous semantic — the old code already had `dataPooled = false` → `WriteBackToGuest = false`, so synthetic zeros were never written back to guest memory. The change makes it explicit.

The per-(shader, pc) trace dedup is the right pattern — same as the existing `_tracedScalarFallbacks`.

Build 0/0, suite 512/512 (the new `RepeatedUnmappedBufferFallbacksReuseZeroedTransferStorage` test asserts both the zeroing guarantee and sub-1MiB allocation over 12 pooled 1MiB fallbacks). Gameplay bot green. LGTM.

PR #485, opened 2026-07-21
optimization: single-pass region-offset resolution on the memory access hot path

slick-daddy:
## Summary

The redundancy

FindRegion(address, size) is the lookup behind every guest memory read, write, and compare in the emulator. To validate that a candidate region actually contains the requested range, it called TryResolveRegionOffset(address, size, candidate, out _) — and discarded the computed offset. Then every caller (TryRead, TryWrite, TryCompare, TryCopy, TryReadExclusive, TryWriteExclusive) turned around and called TryResolveRegionOffset again with the same arguments to get the offset it needed.

So each memory access did the offset resolution twice — once to validate, once to actually use. TryCopy did it four times (two regions × two calls).

The fix

Added a FindRegion(address, size, out ulong offset) overload that returns the offset it already computes, with the old signature delegating to it. Threaded the offset through all seven call sites. This also let the caller-side region is not null && TryResolveRegionOffset(...) checks collapse into a plain region is not null — so the code got shorter and faster.

## Testing

N/A

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
+++ b/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
@@ -931,13 +931,8 @@ public sealed unsafe class PhysicalVirtualMemory : IVirtualMemory, IGuestMemoryA
         _gate.EnterReadLock();
         try
         {
-            var region = FindRegion(virtualAddress, (ulong)destination.Length);
-            if (region is not null &&
-                TryResolveRegionOffset(
-                    virtualAddress,
-                    (ulong)destination.Length,
-                    region,
-                    out var offset))
+            var region = FindRegion(virtualAddress, (ulong)destination.Length, out var offset);
+            if (region is not null)
             {
                 var srcPtr = (void*)(region.VirtualAddress + offset);
                 if (destination.IsEmpty)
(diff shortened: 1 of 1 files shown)


Spooks4576:

Tested `oopsie-2` @ `dbcec78b` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-485

kostyaff (approved):
The refactor is semantically sound — `TryResolveRegionOffset` is unchanged, and the new `FindRegion` overload just surfaces the offset it was already computing internally. All seven call sites collapse from `region is not null && TryResolveRegionOffset(...)` to a plain null check, which is both shorter and eliminates the duplicate walk.

Callers verify `region is null` before touching `offset`, so the `offset = 0` initialization on the failure path is correct (no observable change).

One missed opportunity (non-blocking): `GetPointer` at line 1305 also uses the old `FindRegion(addr, 1)` and could benefit from the new overload, since it immediately uses the pointer. It's not on the hot path like TryRead/TryWrite/TryCopy, so a follow-up is fine.

Build 0/0, full suite 511/511 pass locally (win-x64, Release). Gameplay bot green. LGTM.

PR #486, opened 2026-07-21
optimization(metal): memoize objc selector and class handle lookups

slick-daddy:
## Summary
The problem
Every objc_msgSend in the Metal presenter needs a selector pointer, obtained via Selector("someMethod:"). The old implementations P/Invoked straight into the objc runtime on every call:

public static nint Class(string name) => objc_getClass(name);
public static nint Selector(string name) => sel_registerName(name);

Each call marshals the managed string into a native UTF-8 buffer, crosses the managed→native boundary, and hashes the name inside the objc runtime. But selector and class pointers are registered once and stable for the entire process — the result never changes. Meanwhile the draw path resolves the same small set of selectors ("release", "commit", "setRenderPipelineState:", "renderCommandEncoderWithDescriptor:", …) over and over: Draws.cs alone has 112 call sites, and they fire hundreds of times per frame across every draw. That's a per-draw pile of redundant string marshalling and runtime lookups for values that are effectively constants.

The fix
Memoize both through a ConcurrentDictionary<string, nint>. A repeat resolve becomes a dictionary hit instead of a managed→native round trip. The runtime calls are idempotent and thread-safe, so the concurrent cache needs no extra locking.

## Testing
N/A

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Gpu/Metal/MetalNative.cs
+++ b/src/SharpEmu.Libs/Gpu/Metal/MetalNative.cs
@@ -1,6 +1,7 @@
 // Copyright (C) 2026 SharpEmu Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
+using System.Collections.Concurrent;
 using System.Runtime.InteropServices;
 
 namespace SharpEmu.Libs.Gpu.Metal;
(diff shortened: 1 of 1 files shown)


Spooks4576:

Tested `oopsie-3` @ `7923bfb0` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-486

slick-daddy:
CI flake is fixed in https://github.com/sharpemu/sharpemu/pull/488

PR #501, opened 2026-07-21
[Memory] Roll back partial fixed-range backings when a later gap cannot be backed

h4sht:
## What this does

Fixes a memory leak in `TryBackFixedRange` where partially-backed pages were left orphaned when a later gap in the same range could not be backed.

**Before:** Each successfully backed page run was inserted into `_regions` (the region tracking table) during the walk. When a later gap failed (host `Query` error or allocation failure), the already-backed runs remained in `_regions` and host memory — invisible to the caller, never freed.

**After:** Backed runs are collected in a temporary list. On success, they are inserted into `_regions` in a single batch under the write lock. On any failure, every backed run is freed from host memory before returning `false` — no pages leak, no orphaned regions.

## Why it\'s needed

`KernelVirtualRangeAllocator.TryReserve` calls `TryBackFixedRange` (`backPartialOverlap` path) without any failure cleanup. When `TryBackFixedRange` partially succeeded and then failed, those pages leaked permanently — corrupting the VM state and wasting host memory. The loader path (`SelfLoader.LoadCore`) worked around this by calling `Clear()` on failure, but that is a heavy hammer that resets the entire VM.

This fix eliminates the leak at the source so all callers are safe.

## How it was verified

- Existing tests `TryBackFixedRangeFillsOnlyTheFreePagesOfAPartiallyOccupiedRange` and `TryBackFixedRangeReturnsFalseWhenRangeIsFullyOccupied` continue to pass (semantics unchanged on the success path).
- New test `TryBackFixedRangeRollsBackPartialAllocationsOnQueryFailure`: a mock reports free pages for the first Query (so one gap is backed), then returns `false` on the next Query. The test asserts that the backed allocation was freed during rollback.
- New mock `IntermittentQueryFailureHostMemory` tracks both allocations and frees.

## Testing

### DIFF
--- a/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
+++ b/src/SharpEmu.Core/Memory/PhysicalVirtualMemory.cs
@@ -446,13 +446,19 @@ public sealed unsafe class PhysicalVirtualMemory : IVirtualMemory, IGuestMemoryA
         // us over whole free or occupied stretches. Only free stretches get backed;
         // stretches already reserved or committed by another allocation are left as
         // they are, which is exactly what a fixed mapping does on hardware.
+        //
+        // Backed runs are collected and only inserted into _regions after the
+        // entire range succeeds. If a later gap cannot be backed, every previously
+        // backed run is freed so no orphaned pages leak into the region table.
+        var backedRuns = new List<(ulong Address, ulong Size)>();
         var cursor = start;
-        var backedAny = false;
-        while (cursor < end)
+        var failed = false;
+        while (!failed && cursor < end)
         {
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `fix/trybackfixedrange-rollback` @ `4391d34f` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-501

PR #504, opened 2026-07-21
[Kernel/Semaphore] Close race between sceKernelWaitSema and sceKernelSignalSema

h4sht:
## What this does

Adds a re-check inside `sceKernelWaitSema` after `RequestCurrentThreadBlock` sets the thread-static block flags but before yielding to the scheduler. If the semaphore count is now sufficient (a signal arrived during the registration window), the thread consumes the tokens, cancels the pending block, and returns without ever yielding.

## Why it's needed

`sceKernelWaitSema` follows this sequence when the count is insufficient:

1. Under `lock(semaphore.Gate)`: increment `WaitingThreads`, release lock
2. `RequestCurrentThreadBlock()` — sets `[ThreadStatic]` pending-block flags
3. Return to HLE dispatch → scheduler calls `RegisterBlockedGuestThreadContinuation()`
4. Thread exits to scheduler → exit handler sets state to Blocked → re-checks `TryWake()`

`sceKernelSignalSema` does:

a. Under `lock(semaphore.Gate)`: increment `Count`, release lock
b. `WakeBlockedThreads(wakeKey)` — iterates threads, checks `HasBlockedContinuation && BlockWakeKey == wakeKey`, calls `TryWake()`

**The race:** A signal (step a-b) can arrive between steps 2 and 3. `WakeBlockedThreads` finds no registered waiter (`HasBlockedContinuation` is still false) and returns 0. The signal's count increment is lost — no thread was woken, and the waiter registers its block metadata *after* the wake iteration.

The scheduler's exit handler (step 4) *does* re-check `TryWake()` after setting the state to Blocked, which catches this race in most cases. However, this re-check requires the thread to fully exit to the scheduler and back — an unnecessary round-trip. The earlier re-check introduced here catches the signal while still inside the HLE, avoiding the scheduler yield entirely.

## How it was verified

- Code review confirms the re-check follows the same pattern as the exit handler's existing re-check (`DirectExecutionBackend.cs` line 4876) and the GC suspension race handler (line 4118)
- The `semaphore.Gate` lock ensures the re-check is atomic with respect to `KernelSignalSema`'s count increment
- `TryConsumeCurrentThreadBlock(out _)` properly clears the ThreadStatic flags set by `RequestCurrentThreadBlock`, so the scheduler never sees a pending block
- `WaitingThreads` is decremented using the same `Math.Max(0, ...)` pattern used in `WakePredicate` and `ResumeWait`

## Testing

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelSemaphoreCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelSemaphoreCompatExports.cs
@@ -191,6 +191,27 @@ public static class KernelSemaphoreCompatExports
                 WakePredicate,
                 deadline))
         {
+            // A signal may have arrived between releasing the semaphore gate
+            // (after incrementing WaitingThreads) and the scheduler registering
+            // this block.  When that happens WakeBlockedThreads cannot find the
+            // waiter yet and the exit-handler re-check runs later; a re-check
+            // here keeps the thread from yielding to the scheduler at all when
+            // the count is already sufficient.
+            lock (semaphore.Gate)
+            {
+                if (semaphore.Count >= needCount)
+                {
+                    semaphore.Count -= needCount;
+                    semaphore.WaitingThreads = Math.Max(0, semaphore.WaitingThreads - 1);
+                    GuestThreadExecution.TryConsumeCurrentThreadBlock(out _);
+                    if (_traceSema)
+                    {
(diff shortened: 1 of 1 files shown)


Spooks4576:

Tested `fix/semaphore-wait-race` @ `ccc35eb0` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-504

PR #538, opened 2026-07-22
[CPU] Preserve guest return values across TLS lookup

Dafenxz0:
## What changed

- Preserve the guest's `RAX` value while the return stub calls `TlsGetValue`.
- Keep the TLS slot pointer in `R11` while restoring the original guest result.
- Add an ABI regression test for the emitted return-stub instructions.

## Why

`CreateGuestReturnStub` receives the guest function result in `RAX`, then calls
`TlsGetValue` to recover the host stack. That host call also returns in `RAX`, so
the guest result was replaced by the TLS slot pointer before control returned to
the caller.

This preservation previously existed in #104, but was lost when the CPU backend
was rewritten by #216. The change restores only that ABI guarantee while keeping
the current XMM restoration and allocation-failure cleanup intact.

## Impact

Guest entry points that return an integer or pointer now deliver that value to
their caller instead of leaking an internal host TLS address. The fix changes
only the one generated return stub; it adds no work to normal import dispatch.

## Testing

- No commercial game dump was available for runtime testing.
- `dotnet test SharpEmu.slnx -c Release --no-build --verbosity normal`
  - SharpEmu.Libs.Tests: 542 passed
  - SharpEmu.SourceGenerators.Tests: 33 passed
  - SharpEmu.ShaderCompiler.Metal.Tests: 27 passed

## Checklist

- [x] I listed the game(s) I tested (or clearly stated that game testing was unavailable).

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -2444,9 +2444,16 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 
 		byte* code = (byte*)ptr;
 		int offset = 0;
-		EmitByte(code, ref offset, 0x48); // sub rsp, 0x20
+		// TlsGetValue returns its TLS pointer in RAX. Preserve the guest return value
+		// above the 32-byte Windows shadow space while keeping the call site aligned.
+		EmitByte(code, ref offset, 0x48); // sub rsp, 0x30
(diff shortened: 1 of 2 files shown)


Spooks4576:

Tested `agent/preserve-guest-return-value` @ `1f5513ed` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-538

PR #564, opened 2026-07-23
fix: VulkanHostBufferPool deadlock, audio overflow crash, and log grouping

CasualcoderDev:
While reading through the codebase I found a batch of real issues that can cause freezes or incorrect behavior. Nothing here is cosmetic — each fix targets a specific bug I verified by tracing the code path.

### What's fixed:

* **VulkanHostBufferPool thread safety (Deadlock Fix)** — `TryRent`, `Return`, `Register`, and `Dispose` mutated shared dictionaries without synchronization. Concurrent command buffer submissions corrupted internal bucket chains. I added a `lock(_gate)` to protect the dictionaries, but carefully moved the `_destroy(allocation)` calls **outside** of the lock. Holding the C# lock while calling Vulkan driver cleanup APIs was causing a lock-ordering deadlock that froze games entirely.
* **AudioPcmConversion overflow crash** — `ConvertFloatSample` used a `checked` short cast. Floating-point precision rounding anomalies could trigger an `OverflowException` on the audio thread. Since the preceding `Math.Clamp` already guarantees the value fits in range, I swapped it to an `unchecked` cast to prevent the audio pipeline from crashing on edge-case float rounding.
* **EmulatorProcess log grouping** — `ForwardOutput` was filtering out empty strings. This collapsed adjacent log sections in the GUI console into a single unreadable block. Blank lines act as visual separators for sequential events (like shader compile → render). Now, only the `null` sentinel (stream EOF) is filtered so the GUI console preserves proper grouping.

### Testing:

- **Dreaming Sarah / Dead Cells:** The buffer pool `Return` eviction path is hit constantly under heavy GPU load. The VulkanHostBufferPool changes prevent the renderer from locking up and freezing during gameplay in these titles.
- The `AudioPcmConversion` and `EmulatorProcess` changes prevent crashes and correct GUI formatting under specific conditions, but do not alter standard runtime game output.

### Checklist (check all these when submitting)

### DIFF
--- a/src/SharpEmu.GUI/EmulatorProcess.cs
+++ b/src/SharpEmu.GUI/EmulatorProcess.cs
@@ -387,7 +387,7 @@ internal sealed class EmulatorProcess : IDisposable
 
     private void ForwardOutput(string? line, bool isError)
     {
-        if (!string.IsNullOrEmpty(line))
+        if (line is not null)
         {
             OutputReceived?.Invoke(line, isError);
         }
--- a/src/SharpEmu.Libs/Audio/AudioPcmConversion.cs
+++ b/src/SharpEmu.Libs/Audio/AudioPcmConversion.cs
@@ -68,7 +68,7 @@ internal static class AudioPcmConversion
 
         value = Math.Clamp(value, -1.0f, 1.0f);
         var scale = value < 0.0f ? 32768.0f : short.MaxValue;
-        return checked((short)MathF.Round(value * scale));
+        return unchecked((short)MathF.Round(value * scale));
     }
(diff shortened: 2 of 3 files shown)


Spooks4576:

Tested `main` @ `12cff892` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-564

PR #574, opened 2026-07-23
[AGC] Broadcast completion events to all registered graphics waiters

### LINKED ISSUE #173: AGC event queue never wakes waiters (EVENT_TYPE vs eventId mismatch)

h4sht:
## What this does

Fixes the AGC event queue deadlock where `AgcInterruptThread` hangs forever in `sceKernelWaitEqueue` because DCB completion notifications only woke listeners registered with `eventId=0`.

`sceAgcDriverAddEqEvent` registers graphics completion listeners with a game-defined `eventId` (e.g. `0x20` in Poppy Playtime). The old completion path called `TriggerRegisteredEvents(ident: 0, ...)` which only matched registrations whose `eventId` happened to be zero — missing every real title.

This PR switches the completion notification to `TriggerRegisteredEventsDistinct` which broadcasts to **all** graphics-filtered registrations regardless of their ident. This matches the `IT_EVENT_WRITE` packet path which already uses `TriggerRegisteredEventsByFilter` (filter-based broadcast).

## Why it's needed

Fixes #173 — **AGC event queue never wakes waiters (EVENT_TYPE vs eventId mismatch)**.

Games that wait on graphics/compute completion events through `sceKernelWaitEqueue` — which includes Poppy Playtime and likely many other titles — hang indefinitely after the GPU finishes work because the completion event never fires for their registered `eventId`.

## How it was verified

- The `IT_EVENT_WRITE` path already uses filter-based broadcast (`TriggerRegisteredEventsByFilter`) and works correctly
- `TriggerRegisteredEventsDistinct` uses the same filter-matching logic, just without a data payload
- Completion events should signal ALL registered events on the queue, which matches the real PS5 hardware behavior

## Testing

---

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -3030,15 +3030,12 @@ public static partial class AgcExports
         state.CompletionEventNotifiedSubmissionId = submissionId;
         void TriggerCompletionEvents()
         {
-            var triggered = KernelEventQueueCompatExports.TriggerRegisteredEvents(
-                ident: 0,
-                KernelEventQueueCompatExports.KernelEventFilterGraphics,
-                data: 0);
-            if (_compatibilitySubmitCompletionEvent)
-            {
-                triggered += KernelEventQueueCompatExports.TriggerRegisteredEventsDistinct(
-                    KernelEventQueueCompatExports.KernelEventFilterGraphics);
-            }
+            // Broadcast to ALL registered graphics events regardless of the
+            // eventId the game used (e.g. 0x20 for Poppy Playtime).  The
+            // ident=0 narrow match only wakes listeners registered with
+            // eventId=0, which misses every real title.
+            var triggered = KernelEventQueueCompatExports.TriggerRegisteredEventsDistinct(
+                KernelEventQueueCompatExports.KernelEventFilterGraphics);
             TraceAgc(
                 $"agc.driver_submit_dcb completion submission={submissionId} " +
                 $"queues={triggered}");
--- a/src/SharpEmu.Libs/Kernel/KernelFileExtendedExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelFileExtendedExports.cs
@@ -46,7 +46,13 @@ public static partial class KernelMemoryCompatExports
 
     [SysAbiExport(Nid = "ezv-RSBNKqI", ExportName = "pread",
         Target = Generation.Gen4 | Generation.Gen5, LibraryName = "libKernel")]
-    public static int PosixPread(CpuContext ctx) => KernelPreadCore(ctx);
+    public static int PosixPread(CpuContext ctx)
+    {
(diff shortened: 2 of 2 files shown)


Spooks4576:

Tested `fix/agc-event-queue-broadcast-completion` @ `9be5f10f` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-574

PR #589, opened 2026-07-23
Add Darwin fixed-address mach_vm_allocate fallback to HostMemory

NickPom:
## What changed

`HostMemory`'s Posix allocator (used by all `sceKernel*` memory calls) only tried an mmap hint for exact-address allocations. Darwin has no `MAP_FIXED_NOREPLACE`, so a hint that lands at the wrong address left the allocation failing with no fallback, unlike the sibling `Host/Posix/PosixHostMemory.cs` allocator, which already has this same `mach_vm_allocate` fallback from #246.

Ports the same fallback here: on macOS, when the hint-only mmap does not land at the requested address, retry with a fixed-placement `mach_vm_allocate`, which either maps the exact range or fails outright without silently clobbering untracked host memory (CLR, dyld, JIT, Rosetta) the way a raw `MAP_FIXED` could.

## Why

Found while investigating a title that needs a fixed-address allocation `HostMemory` currently can't satisfy on macOS. Brings this allocator in line with its sibling so the two don't silently diverge in behavior.

## Testing

- `dotnet test SharpEmu.slnx` -- 602 passed, 0 failed.
- macOS only change, gated by `OperatingSystem.IsMacOS()`; no behavior change on other platforms.

## Checklist

### DIFF
--- a/src/SharpEmu.HLE/HostMemory.cs
+++ b/src/SharpEmu.HLE/HostMemory.cs
@@ -232,6 +232,25 @@ public static unsafe class HostMemory
 
                     var exactFlags = OperatingSystem.IsMacOS() ? flags : flags | MAP_FIXED_NOREPLACE;
                     result = mmap((nint)address, (nuint)alignedSize, posixProtect, exactFlags, -1, 0);
+                    if (result != MAP_FAILED && (ulong)result != (ulong)address)
+                    {
+                        munmap(result, (nuint)alignedSize);
+                        result = MAP_FAILED;
+                    }
+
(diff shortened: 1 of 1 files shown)


Spooks4576:

Tested `fix/host-memory-darwin-fixed-fallback` @ `f08bb116` — 60 seconds of automated gameplay per game on real GPU hardware (Linux).

| Game | Result |
| --- | --- |
| Dreaming Sarah | ✅ passed |
| Dead Cells | ✅ passed |

Failure reasons, gameplay recordings & evidence bundles: https://pr.sharpemu.app/#pr-589

PR #592, opened 2026-07-23
feat(gpu): GPU compute detile for guest tiled textures (Vulkan + Metal)

shadowbeat070:
### Why
The guest textures were RDNA2 swizzled tiles and detiling them was currently being done on the CPU concurrently. This PR resolves this by instead of detiling everything on the cpu, it now checks if the element has a supported layout and if yes, tries to get detileParams and if swizzlemode is not 0 and .Equation is ExactXOR (5 / 9 / 24 / 27, 4/ 8/ 16 bpp) or BlockTable (1 / 4 / 8 (Morton/Z-order)) then it sends it to the gpu compute shader (vulkan/metal) for detiling. Any unsupported modes fallback to the CPU.

### What

AgcExports now ships raw tiled bytes and params to the GPU if the correct IsGpuDetileEquation(eq) / IsGpuDetileBytesPerElement(bpp) pass. Other cases will still use the old CPU detiling (GnmTiling.DetileWithParams). DetileParams is the record struct that the CPU/Vulkan/Metal consume to try and detile the specified texture.

SpirvFixedShaders.CreateDetileCompute emits the SPIR-V kernel per instruction. RecordDetile records the dispatch into the shared command buffer and hand back the transient buffers/pool for the caller to reitre with the batch fence. If a failure occurs then is falls back to the CPU detiling. VulkanDetileSelfTest (SHARPEMU_DETILE_SELFTEST=1) validates both entry points at device init.'

detile_compute.msl and MetelDetilePass mirrors the Vulkan pass however the MetalDetilePass still needs to finished/tested since i don't have a mac.

**Flags**
| Env var | Effect | Default |
|---|---|---|
| `SHARPEMU_GPU_DETILE` | `=0` forces the CPU path | GPU detile **on** |
| `SHARPEMU_LOG_GPU_DETILE` | `=1` enables `[GPU-DETILE]` diagnostics | off |
| `SHARPEMU_DETILE_SELFTEST` | `=1` runs the Vulkan self-test | off |

## Testing

Untested as of yet. Still needs testing on 3d games on multiple platforms. `N/A`

 17 detile unit tests pass, DetileWithParams and GetDetileParams each match TryDetile bit-for-bit across all supported modes/bpp, plus a SPIR-V structural-validity test.

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -271,6 +271,25 @@ public static partial class AgcExports
         Environment.GetEnvironmentVariable("SHARPEMU_NO_TEXTURE_SKIP"),
         "1",
         StringComparison.Ordinal);
+
+    // GPU deswizzle: ship raw tiled bytes + params to the backend instead of
+    // detiling on the CPU. On by default; SHARPEMU_GPU_DETILE=0 forces the CPU
+    // path. Backend-agnostic here (only inspects DetileParams); the Vulkan/Metal
+    // backends detile on the GPU, others fall back to the CPU path.
+    private static readonly bool _gpuDetileEnabled = !string.Equals(
+        Environment.GetEnvironmentVariable("SHARPEMU_GPU_DETILE"),
+        "0",
+        StringComparison.Ordinal);
+
+    // Diagnostics (SHARPEMU_LOG_GPU_DETILE=1): one line per distinct texture tile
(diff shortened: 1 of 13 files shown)


par274:
Hello, I was actually planning to do this myself, but you got to it before I did.

At the moment, though, there's a regression specific to Demon's Souls. Text's are does not rendering.

shadowbeat070:
The reason that texts were not rendering was because the PR did not support arrayed textures and the CPU detiling fallback was only wired to trigger if the GPU tiling failed.

PR #602, opened 2026-07-24
Add disassemble command to the live debug server

Leonid749:
This adds disassemble command ('disasm' / 'u') to live debug server.

Right now you can 'read-memory' over the debug protocol but all you get back is raw hex. You have to decode it yourself to know what instruction you're actually looking at. This just decodes it for you and sends back the address, length, raw bytes, and the actual instruction text (like mov eax, 1) for a run of instructions starting at whatever address you ask for.

It doesn't add a new decoder or anything - it just reuses 'IcedDecoder', which the CPU core already uses internally. So this is mostly plumbing: hooking existing decode logic up to the debug protocol.

A couple of implementation notes:
 It reads memory one byte at a time per instruction instead of one big fixed-size chunk so if you hit an unmapped page partway through you still get back the instructions that decoded fine instead of the whole request just failing.
 Added 'disasm <addr> [count]' to the DebugClient REPL and updated the two docs that describe the protocol/commands.

Testing: N/A, this doesn't touch guest/game behavior at all, it's just debugger tooling. I ran `dotnet build` (clean, no errors) and the full test suite, plus added 5 new unit tests for this specifically - decoding a known instruction sequence stopping cleanly when memory runs out mid-run failing properly when the target isn't paused and rejecting a bad 'count' value.

### DIFF
--- a/docs/debugger-server.md
+++ b/docs/debugger-server.md
@@ -100,6 +100,7 @@ number or a `0x`-prefixed hex string.
 | `set-register` | `register`, `value` | — |
 | `read-memory` | `address`, `length` (≤ 65536) | `address`, `length`, `bytes` (hex) |
 | `write-memory` | `address`, `bytes` (hex) | `written` |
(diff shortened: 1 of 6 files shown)


par274:
Please read the contribution file and then check PR template.

PR #603, opened 2026-07-24
Prevent TLS init function from corrupting guest memory on emulator threads (18+ imports resolved)

Fa1dz:
## What It does
TryRecoverGuestTlsLoad can recover the TLS base correctly, but some TLS init functions continue executing afterwards and write to invalid guest globals through RIP relative stores. This detects those init functions and exits before the corruption happens by returning to the caller.

## Testing

I tested UFC 5 and it booted past module init without crash. Previously corrupted guest globals and crashed at vmovups [rip+...], ymm0 zeroing. After fix, execution continued through (18+ imports.)

## Checklist


Fa1dz:
Tested with UFC 5. Before this change it would crash during module init after the TLS handling corrupted guest globals. After the fix it gets past that point and continues loading further into the game (18+ imports resolved). The change is limited to detecting this specific TLS init pattern, so it should not affect normal TLS handling paths.

PR #684, opened 2026-07-28
updated the allocator in KernelMemoryCompatExports.cs and added a test in KernelMemoryCompatExportsTests.cs

pal3635:
sceKernelAllocateMainDirectMemory is the guest import in question. The game i tested it on was Ghost of Yotei. Currenly it crashes after splash screen is displayed.

The guest was calling the export, asking for direct memory to use but the call was failing multiple times with the message 'ORBIS_GEN2_ERROR_TRY_AGAIN' in the console logs. Implementation inside the emulator was being too strict about where it looked for free space as it was trying to find free area in the current main direct memory pool., if failed, it assumed that allocation cannot be satisfied and returned 'TRY_AGAIN' immediately. 

The bug was that the allocator did not honor that base properly during the initial placement path and gave up after it's first failure. Now it starts and searches the relevant memory window and continues to allow allocation in a later pool window when the earlier one is already occupied or not suitable. Without this change, the emulator could reject a perfectly valid allocation request just because it did not fit in the first window the code checked.

test added in KernelMemoryCompatExportsTests.cs simulates a situation where one window is already occupied, a later window should still be usable, and the allocation should still succeed.

It now stops producing the  'TRY_AGAIN' failure for normal direct allocations

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelMemoryCompatExports.cs
@@ -2989,28 +2989,27 @@ public static partial class KernelMemoryCompatExports
         ulong aligned;
         lock (_memoryGate)
         {
+            var poolBase = _mainDirectMemoryPoolBase == UnsetMainDirectMemoryPoolBase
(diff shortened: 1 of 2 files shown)


par274:
Read the contribution file before opening the PR and stick to the PR template.

PR #693, opened 2026-07-29
Improve update checks and changelog dialog

mskomek:
## Summary
- Add a scrollable update confirmation window with download/later actions and a configurable reminder preference.
- Retry only update-check timeouts: three 20-second attempts with ten-second delays.
- Parse release commit links correctly and compare commits rather than prerelease-style version labels.
- Show every release between the installed commit and the newest release, newest first, with version-separated changelogs.

## Root cause
The updater expected a plain commit SHA in GitHub release notes and treated hotfix labels as prereleases, which could incorrectly suppress a valid update.

## Validation
- Automated UpdaterTests passed.
- Manually tested the update dialog, reminder override, changelog display, and a full download/SHA-256/restart flow using a temporary local release server.


mskomek:
Closing this draft for now; the update work will continue locally and be submitted as a consolidated follow-up.

PR #731, opened 2026-08-01
Investigate oversized detile work ownership

KiroshiAiviom:
## Summary

This PR adds opt-in, bounded diagnostics and records the finding in
docs/OVERSIZED_DETILE_WORK.md. It does not implement the recommended memory
correction, change queue behavior, or change GPU/shader behavior.

The investigated target is Demon’s Souls PPSA01341, Europe/EP9000,
v1.004.000, eboot SHA-256
22ED8843917CB16438B7B780998E408321F5CEBE79DD10F388AE59CFCA588306.

## Reproduced evidence

On diagnostic commit b67b143, three clean comparable trials reached the same
oversized event:

- 20260801T033052405Z-b67b143-trial-01, child PID 14496, sequence 765,
  18,891 ms, sampled peak 9.157 GiB WS / 14.094 GiB private.
- 20260801T033111309Z-b67b143-trial-02, child PID 21496, sequence 767,
  16,163 ms, sampled peak 9.353 GiB WS / 14.276 GiB private.
- 20260801T033148289Z-b67b143-trial-04, child PID 22256, sequence 768,
  16,764 ms, sampled peak 9.366 GiB WS / 14.284 GiB private.

Each reached the first rendered frame after splash; none reached the character
creation nickname prompt. The event payload and state sequence were identical.
Two additional final-commit trials hit the unchanged safety cutoff before the
dispatch was enqueued, and one exited with code 4 while gathering resource
file information. Those are reported as timing variance, not combined with
the three event-bearing samples.

The item is VulkanComputeGuestDispatch, operation compute-dispatch, on logical
queue acb.compute[64], submission 8. It has sequence 765/767/768,
requiredSequence 0 already complete, shader address 0x0000000448626400,
27 x 15 x 72 groups, local size 8 x 8 x 1, 40 texture bindings, and 12 global
buffers. It transitions ready-head -> taken -> executing. At the cutoff the
queue depth and pending count are zero while pending bytes remain
5,378,410,360: the item has been taken and is active in the presenter, but
execution has not returned through CompleteGuestWork. It is not blocked behind
another item, a cross-queue dependency, or the queue budget. The exact inner
post-take phase remains unresolved.

## Exact composition

The stored payload is 5,378,410,360 bytes (5,129.251823 MiB):

| Category | Unique arrays | References | Bytes |
| --- | ---: | ---: | ---: |
| Compute shaders | 1 | 1 | 1,824,536 |
| Vertex shaders | 0 | 0 | 0 |
| Pixel shaders | 0 | 0 | 0 |
| RgbaPixels | 14 | 40 | 5,384,192 |
| TiledSource | 16 | 16 | 5,368,709,120 |
| Global buffers | 12 | 12 | 2,492,512 |
| Vertex/index/image-write arrays | 0 | 0 | 0 |

There are 43 unique arrays. The 16 TiledSource arrays are each 335,544,320
bytes and account for 5,120 MiB. The 12 global buffers are
[262144 x 5, 16, 32, 16, 32, 131072, 1048576, 2048]. RgbaPixels is
[524288 x 10, 131072, 8192, 2048, 0], with the zero-length array referenced
27 times.

The complete descriptor/range grouping is in the focused document and local
JSONL diagnostic output. The dominant group is guest range
0x102A400000..0x103E400000, 1024 x 1024, depth 1, pitch 1024, 80 array
layers, format 4, number type 7, tile mode 24, type 13, mip 0/base 0/resource
mips 1, non-storage arrayed view: 16 exact copies. The second repeated group
is guest source range 0x1026C40000..0x1026CC0000, 1024 x 1024, depth 1,
pitch 1024, one layer, format 170, number type 9, tile mode 9, type 13,
mip 0/base 0/resource mips 11, non-storage arrayed view: 10 exact copies.

## Ownership and rejected hypotheses

ObserveComputeDispatch decodes all 40 bindings; CreateGuestDrawTextures calls
TryCreateGuestDrawTexture independently for each. The existing content lookup
occurs before each allocation, but it is cold before the consumer creates the
host resource and there is no dispatch-local source snapshot table. The arrays
are therefore eagerly allocated and read before SubmitComputeDispatch enqueues
the work; the queue budget cannot act before allocation.

All 16 large TiledSource objects have the same exact guest range and recorded
content identity but are distinct managed arrays. The same is true of the ten
repeated RgbaPixels objects. There are no partial overlaps among nonzero
ranges. Allocated lengths match the calculated source lengths; the small
physical padding in the two 3D descriptors is legitimate. This rejects a
dominant descriptor-size error and proves duplicate capture. Queue/dependency
blocking is rejected by the ready/taken/executing events. The diagnostic object
retains only scalar metadata; the original source arrays are owned by the
work/active execution path until verified completion.

## Recommendation (not implemented)

Add a dispatch-local immutable snapshot table at
AgcExports.CreateGuestDrawTextures/TryCreateGuestDrawTexture. Reuse a source
array only when the exact guest range, selected view/mip, descriptor/layout,
storage/write semantics, and guest write generation match; preserve a separate
GuestDrawTexture per binding. The observed reduction is 5,037,883,392 bytes
(4,804.5 MiB), from 5,129.251823 MiB to about 324.751823 MiB.

Required invariants are queue order, resource visibility, write generation,
view/mip/layer/sampler/storage semantics, immutable shared ownership, and
render-thread follow-up deadlock safety. Synthetic tests should cover exact
sharing, incompatible descriptors/ranges refusing to share, array-layer
sizing, and queue lifecycle accounting. Three target runs should then show
one 320 MiB TiledSource owner instead of 16 and one 512 KiB duplicate
RgbaPixels owner instead of 10. The recommendation is falsified by stale
data under guest writes, backend mutation of shared arrays, distinct content
under the same key, or no payload reduction.

## Verification

- Focused diagnostics/accounting tests: 9/9 passed.
- Fast lane: passed, 0 build warnings, 0 build errors, 781 tests passed.
- git diff --check: passed.
- Shader lane: not run; no shader translation or GPU execution semantics
  changed.
- No game files or game-derived contents were added to Git.

Commits:

- b67b143 diag: trace oversized guest work ownership
- d5f4c21 docs: record oversized detile ownership finding


KiroshiAiviom:
Superseded: this investigation belongs in the Demon’s Souls fork baseline. The intended PR is KiroshiAiviom/sharpemu-demons-souls#6.

PR #738, opened 2026-08-02
AGC: honor DCC fast clears instead of drawing the clear quad

ArsenyYankovsky:
## Motivation

image_2026-08-02_11-57-20

Continuing the Demon's Souls bring-up. 

The 3D scene was hidden behind an opaque UI layer.

On GFX10 there's no real clear packet for DCC. The game sets DCC_ENABLE, writes CLEAR_WORD0/1 = 0, then draws a full-screen quad expecting the CB to turn it into DCC clear codes and drop the shader output. We drew the quad for real. Its blend (src=ONE, dst=ONE_MINUS_SRC_ALPHA) only ever raises alpha, and since we clear color targets once on first use and load them after, nothing reset it, so it pinned at 1. Compositing is ui.rgb + scene.rgb*(1-ui.a), so that zeroed the scene.

To confirm it I decoded the composite pixel shader from its captured ISA to get the blend formula, then added a temporary probe that walks the UI layer's bytes when it's uploaded and reports min/max alpha plus how many texels have any color. It's a 16_16_16_16 float surface, so alpha is the 4th half lane, and the walk uses a prime stride so it crosses tiles instead of striping one row. Before the fix: amin=amax=1, rgb=0, while the scene target had normal content. After: amin=0. The probe isn't part of this PR.

## Testing

All changes were tested on Demon's Souls [PPSA01341].

Got to character creation and the 3D renders now, the model slides up and the scene shows behind the UI layer instead of a black screen.

Probed the UI layer's alpha to confirm the cause, amin=amax=1 with rgb=0 before the fix and amin=0 after. Ran with the fix removed as a control and it went back to opaque black.

## Checklist

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -141,9 +141,15 @@ public static partial class AgcExports
     private const uint CbColor0Base = 0x318;
     private const uint CbColorRegisterStride = 15;
     private const uint CbColor0Info = 0x31C;
+    private const uint CbColor0ClearWord0 = 0x323;
+    private const uint CbColor0ClearWord1 = 0x324;
     private const uint CbColor0BaseExt = 0x390;
     private const uint CbColor0Attrib2 = 0x3B0;
     private const uint CbColor0Attrib3 = 0x3B8;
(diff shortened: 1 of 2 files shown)


par274:
Good job

PR #747, opened 2026-08-02
Fix Vulkan presenter synchronization and frame handling issues

foufouadi:
## Summary

Fix several issues in the Vulkan VideoOut path affecting GPU startup, image handling, writeback performance, and presentation state tracking.

### Start presenter processing for compute submissions

`SubmitComputeDispatch` could enqueue GPU work without starting the presenter worker thread, unlike draw submission paths.

This caused titles whose first GPU workload was a compute dispatch to stall while waiting for guest work processing.

The presenter is now started consistently for compute submissions as well.

### Preserve compatible guest image contents across format changes

Guest render targets can switch between compatible Vulkan formats while referring to the same memory.

Previously, these transitions could recreate the image and discard existing GPU contents. Compatible format pairs now reuse the existing allocation through image views when possible.

Real format conversion is kept behind `SHARPEMU_ENABLE_REAL_FORMAT_CONVERSION` because the conversion path requires additional validation.

### Improve fragmented writeback performance

`WriteBackAllDirtyGuestBuffers` could spend excessive time scanning highly fragmented buffers byte-by-byte.

The scan now uses vectorized comparisons and page-level early checks to reduce the cost of unchanged regions while preserving existing behavior.

### Keep presentation state consistent on dropped frames

Some early-return paths in `Render()` did not advance `_presentedSequence`, leaving internal VideoOut state inconsistent after dropped presentations.

All dropped presentation paths now update the sequence state correctly.

### Explicit GLFW backend initialization

Add explicit GLFW backend initialization through Silk.NET.Windowing to avoid runtime backend probing failures.

## Testing

Tested on Windows:

- Ghost of Yotei (PPSA01327): no regression against upstream/main in the tested boot path.
- Demon's Souls (PPSA01342): no regression observed.
- Astro Bot (PPSA01512): splash displayed, no regression observed.

## Checklist

- [x] I tested my changes.
- [x] I wrote this pull request description myself.
- [x] I listed the game(s) tested.


par274:
Remove the `Silk.NET.Windowing` package, the two using statements, and the `GlfwWindowing.Use()` call from the static constructor. Our windows come from SDL3, so nothing ever asks Silk for a backend and there is no probing to fail. We don't want any additional dependencies.

foufouadi:
Thanks for the quick merge! I managed to get the Yotei intro working, and I still have more fixes coming

par274:
> Thanks for the quick merge! I managed to get the Yotei intro working, and I still have more fixes coming

It's good to see you again! Keep up the good work.

Issue #491, opened 2026-07-21
Fix Sentinel Leak on Sibling Exports Too


slick-daddy:
### Emulator Version / Commit

v0.0.4

### Bug Description

The same sentinel leak exists in the sibling POSIX exports in
KernelFileExtendedExports.cs (pread, pwrite, ftruncate, truncate,
fsync, rename, dup, dup2, fcntl). They aren't on this crash path,
so they're intentionally left for a separate PR using the same
PosixFailure wrapper.

Origanally posted on #461 as Follow up

### Log File

```shell

```

Issue #235, opened 2026-07-15
[Bug][Windows] InvalidOperationException: Could not allocate main image at required base 0x0000000800000000


adriiicn:
### Emulator Version / Commit

d7e4f4f

### Bug Description

Steps to reproduce:
1. Run SharpEmu on Windows 11 with a 16GB (2x8) RAM hardware configuration.
2. Launch a decrypted retail EBOOT (tested with Dreaming Sarah PPSA02929).
3. The emulator immediately crashes during the early initialization phase before creating the render window.

Expected behavior:
The emulator should successfully map the guest virtual memory space for the main image at the required base and proceed to the execution sequence.

Actual behavior:
The engine throws a `System.InvalidOperationException` failing to allocate at the hardcoded base `0x0000000800000000`. 

Technical Analysis (VMMap):
When inspecting the host process using Sysinternals VMMap on a 16GB RAM Windows environment, the .NET Runtime GC automatically reserves a massive 256 GB virtual memory block (Private Data) starting around base 0x000000007FFF0000. This framework reservation completely overlaps the hardcoded base address required by `SelfLoader.LoadCore`. Since `AllocateAt` requires a strict fixed mapping, it fails immediately.

Possible Mitigations:
- Implement a flexible virtual memory remapping layer to handle base address shifts instead of a strict hardcoded location.
- Force `<GCRetainVM>false</GCRetainVM>` or disable Server GC inside the project properties to prevent the large initial virtual block from hijacking the guest address space.

### Log File

```shell
$ SharpEmu --cpu-engine=native --log-level=info C:\Users\adri\Downloads\[SuperPSX]-Dreaming Sarah-PPSA02929-USA-Game--PS5\PPSA02929-app0\eboot.bin
[DEBUG] SharpEmu starting with 3 args
[INFO][SharpEmu.CLI] Program.cs:213 SharpEmu UNOFFICIAL d7e4f4f — https://github.com/par274/sharpemu

Built from branch "215/merge" of "sharpemu/sharpemu" by GitHub Actions workflow run https://github.com/sharpemu/sharpemu/actions/runs/29425234428.
[INFO][SharpEmu.CLI] Program.cs:214 Host hardware: CPU: 11th Gen Intel(R) Core(TM) i5-11400F @ 2.60GHz; GPU: NVIDIA GeForce RTX 3060; RAM: 16.188 MB (15,8 GB).
[...]

PR #763, opened 2026-08-04
Fix/loader overflow and posix sentinel

gerefloc45:
Fix 1 — issue #491 (wrong error signal)

Imagine a function that reads a file. If the file doesn't exist, it should say "I failed, error code X" in the way the calling code expects. Some functions in this emulator (pread, pwrite, ftruncate, truncate, fsync, fdatasync, rename, dup, dup2, fcntl) were saying "I failed" in the WRONG language — they returned an internal error code instead of the standard -1 that C programs check for. So the game's code looked at the result, didn't see -1, and thought "great, it worked!" — then kept using a broken file handle, which can crash later. This same bug was already fixed for other similar functions in a past PR (#461); I just applied the same fix to the ones that were forgotten (issue #491 asked for exactly this).

Fix: made these functions translate the internal error into the standard -1 + error code before returning, using a helper function that already existed in the code for this exact purpose.

Fix 2 — issue #235 (crash loading big/weird game files)

When the emulator loads a game file, it reads a list of "memory blocks" the game needs (called segments), each with a starting address and a size. The code was adding numbers together (address + size) without checking if the result gets too big to fit — like a car odometer rolling over from 999999 back to 000000. If a game file (even accidentally malformed) had a segment address close to the maximum possible number, the addition would silently "roll over" to a tiny wrong number instead of giving an error. Two bad things could happen: the emulator would place game data at the wrong spot in memory with no warning, or it would calculate a wrong total size and then fail to reserve memory at the fixed address PS5 games expect — which matches exactly the error message people report in issue #235.

Fix: added a check before doing the addition — if it would roll over, stop and give a clear error message instead of silently continuing with a wrong number.

Verified with new automated unit tests (TDD): 10 tests for the POSIX sentinel fix, 2 tests for the loader overflow fix. Full solution test suite: 811 passing, 0 failing. No real game tested — I don't own a PS5, so no legal game dump available.

## Checklist

- [x ] I have read and followed `CONTRIBUTING.md`.
- [x ] I tested my changes or marked the testing section as `N/A`.
- [x ] I listed the game(s) I tested (or marked the testing section as `N/A`).

### DIFF
--- a/src/SharpEmu.Core/Loader/SelfLoader.cs
+++ b/src/SharpEmu.Core/Loader/SelfLoader.cs
@@ -474,6 +474,12 @@ public sealed class SelfLoader : ISelfLoader
                 ? 0UL
                 : ResolvePhysicalSegmentOffset(imageData.Length, loadContext, header, index);
 
+            if (header.VirtualAddress > ulong.MaxValue - imageBase)
+            {
+                throw new InvalidDataException(
+                    $"ELF segment {index} virtual address 0x{header.VirtualAddress:X} overflows when combined with the image base 0x{imageBase:X}.");
+            }
+
             var virtualAddress = header.VirtualAddress + imageBase;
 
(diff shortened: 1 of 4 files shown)


gerefloc45:
I forgot to check this box: 
“I wrote this pull request description myself and did not paste AI-generated explanations.”
I rewrote all of this

Issue #758, opened 2026-08-03
Add command line suffix for fullscreen launching


Solitair69:
### Emulator Version / Commit

all

### Bug Description

Pls add the possibility to add suffix like -f or --f or -fullscreen or --fullscreen 
for starting the emulator over command line directly into fullscreen

and sharpemu is still creating the devlog folder in the hard drive root instead of its own folder

### Log File

```shell

```

Issue #683, opened 2026-07-28
Emulation Bug on all Games: Filesystem: encode Windows-invalid guest filename characters


annahana:
Issue Edit 1:
Save files with special Characters like ":" are cut off on befor ":"
Arcade Spirits: The New Challengers tries to create a file
named
"rg_ac_Arcade Spirits: The New Challengers_0.dat"

The emulated guest filesystem is based on FreeBSD semantics and may
contain filenames that are valid on the guest but invalid on Windows.

what we save at the moment is "rg_ac_Arcade Spirits"
if the game trys to loade the file we fail and it crashs

the game itself does not validate if the file is created on saving. what is no a problem as long as the game runs and it is in RAM. but on closing and starting fresh the game looks for that file and crashs on access violation 

we need a virtuall save managmant or a index outseid of the savefolder where the real name of such a file is saved and its equel name how it was named on windows without the special characters.

Linux and MacOS should be fine because there Specal characters are allowed in file names

The current path translation layer maps guest filenames directly to host filenames.

As a result, Windows-invalid characters such as:

":"
"*"
"?"
"
"<"
">"
"|"

may cause Create/Open operations to fail.

Proposed solution

Introduce deterministic filename encoding during guest->host path translation.

Example:

Guest:
Arcade Spirits: The New Challengers.dat

Host:
Arcade Spirits%3A The New Challengers.dat

Suggested encodings:

"% -> %25"
": -> %3A"
"* -> %2A"
"? -> %3F"
" -> %22
"< -> %3C"
"> -> %3E"
"| -> %7C"

Why not replace with '_'

Using '_' can create collisions:

foo:bar
foo_bar

Both become:
foo_bar

Percent encoding preserves a one-to-one mapping.

Open questions
Should encoding happen in the guest path resolver?
Should save data only be encoded or all mounted filesystems?
Is percent encoding preferred over another deterministic scheme?

possible Code for debate

csharp

private static string EncodeHostPathSegment(string segment)
{
    if (!OperatingSystem.IsWindows())
        return segment;

    var sb = new StringBuilder(segment.Length);

    foreach (var ch in segment)
    {
        switch (ch)
        {
            case '%': sb.Append("%25"); break;
            case ':': sb.Append("%3A"); break;
            case '*': sb.Append("%2A"); break;
            case '?': sb.Append("%3F"); break;
            case '"': sb.Append("%22"); break;
            case '<': sb.Append("%3C"); break;
[...]

Issue #639, opened 2026-07-25
Severe host memory growth (~9GB/min) during normal AGC GPU submission on Demon's Souls boot


zambasaurus:
## Bug Description

While instrumenting `libSceSaveDataDialog` for an unrelated investigation, observed the SharpEmu process grow from a normal baseline to **~27GB resident memory in well under 3 minutes** of otherwise-ordinary Demon's Souls (PPSA01341) boot — before reaching the main menu, during regular shader/GPU submission activity.

This is **not** the known unresolved-import retry loop (#619): the trace during this window shows a healthy mix of ~5.9 million real, distinct AGC/libc calls (`sceAgcCbDispatch`, `sceAgcCbSetShRegisterRangeDirect`, `sceAgcGetDataPacketPayloadAddress`, `sceAgcDcbEventWrite`, `memcpy`, etc.) — i.e. normal emulation work, not one NID spinning in a tight failure loop.

Interleaved with this is a steady stream of:
```
[LOADER][WARN] agc.wait_suspended label=0x... queue=acb.compute[N] submission=NNN command=0x... packet=0x... value=0x0000000000000000 mask=0x00000000FFFFFFFF ref=0x0000000000000001 cmp=3 control=0x00000013 bits=32 form=agc-nop producer=none-observed; remaining-suspended
[LOADER][WARN] agc.dispatch_reject source=direct dims=0x... raw=00000000/00000001/00000001 initiator=0x00000041 reason=zero-dimension
```
from `src/SharpEmu.Libs/Agc/AgcExports.cs`. The `wait_suspended`/`dispatch_reject` log lines themselves are bounded (deduped via `_tracedProducerlessWaits` / capped at 256 in `_rejectedDispatchArguments`), so the *logging* is not the leak — but `submission=` IDs climb continuously (observed 48 → 329+ within the window), suggesting each indirect-dispatch retry-and-suspend cycle (see the comment at `AgcExports.cs` around line 9383, `TryDecodeComputeDispatch`) may be allocating per-attempt host/GPU resources (descriptor sets, staging buffers, command encoders) that never get released when the wait never resolves (`producer=none-observed`).

## Evidence

- Fresh `dotnet publish` build of `SharpEmu.CLI` from a branch based cleanly on `upstream/main` (no local GPU-path changes) — ruling out this being caused by uncommitted local work.
[...]

Issue #618, opened 2026-07-25
GPU detile pass (#592) causes severe FPS regression at native/high internal resolution


zambasaurus:
## Bug Description

PR #592 ("GPU compute detile for guest tiled textures", commit `a158960`) introduces a severe FPS regression on at least Demon's Souls (PPSA01341) at native/high internal resolution. This looks like it's the same symptom multiple people are currently reporting on #2 (e.g. "5fps is max" at splash, audio stuttering after the SIE logo) on recent builds.

## Reproduction / Evidence

Reverting `a158960` alone (`git revert a158960` on current `main`, no other changes) restores normal performance on the same title/hardware:

| | With detile pass (current `main`) | `a158960` reverted |
|---|---|---|
| Init to main loop | ~17.3s | ~11.6s |
| `ResourcePool::GatherResourceFileInfo()` | ~36.3s | ~13.9s |
| In-game FPS (menus/cinematic) | 0.3 – 2.0 fps | 4.5 – 6.6 fps |

(Second set of numbers measured on a build with a separate boot-crash fix applied on top — see #617 — but the detile pass is the only variable changed between the two columns.)

Earlier local testing (session log, not on this account) measured the same regression at the extremes: ~6fps → ~0.2fps on Demon's Souls at 3840×2160 with the detile pass present, and init time 6.3s → 1.7s with it reverted.

## Hardware

- CPU: AMD Ryzen 9 7900X
- GPU: NVIDIA GeForce RTX 3090
- RAM: 130 GB
- OS: Windows 11

## Suggested Next Step

Worth profiling `VulkanDetilePass.cs` / the new compute dispatch path specifically — likely candidates are per-frame allocation or a synchronous wait that scales badly at higher internal resolutions, since the regression is much more severe at native res than it may have been in whatever conditions #592 was validated under.

Note: character creation screens remain slow (1.6–2.7fps) even with the detile pass reverted — that looks like a separate, unrelated bottleneck (possibly tied to an unresolved `sceAgcDcbDrawIndirect` import seen firing repeatedly during character-model rendering), not something this issue is about.

zambasaurus:
[...]

PR #771, opened 2026-08-06
Bound host memory budgets, Windows path encoding, and related caps

MartinCajiao:
## Before submitting

I read CONTRIBUTING.md. This PR was developed with AI assistance (Cursor / Grok). I reviewed every change below and can explain why each exists. I am not pasting unexplained generated text as the point of the PR.

## Summary

Generic host-side hardening for ~16 GiB Windows PCs and a Windows filesystem bug that truncates guest filenames containing `:` (issue #683). Also small CLI aliases for fullscreen (#758) and bounded registries that grow during heavy AGC boot (related to #618 / #639 discussion).

No game-specific hacks. Personal UI branding was intentionally kept out of this PR.

### 1. `HostMemoryBudget` (new)
- **Why:** `sceKernelGetDirectMemorySize` previously always advertised a full 16 GiB PS5 pool. On a 16 GiB host that invites commit/OOM because OS + GPU + .NET still need RAM.
- **What:** Auto-scales advertised direct memory from host RAM (leaves ~6 GiB headroom), with `SHARPEMU_DIRECT_MEMORY_MB` override to force a specific size (including 16384 for full PS5 behavior).
- **Also sets defaults for:** lazy-commit thresholds, pending GPU work bytes, cached host buffers, and default `SHARPEMU_RENDER_SCALE` when the env var is unset.
- **Logged at startup** so testers can see the chosen budget.

### 2. `PhysicalVirtualMemory` lazy thresholds
- **Why:** Full-commit preference for multi-GiB maps is correct for accuracy, but on constrained hosts it pins too much host RAM early.
- **What:** Uses `HostMemoryBudget` thresholds so large non-exec maps become eligible for reserve + lazy commit sooner on low-RAM hosts. Executable mappings are unchanged.

### 3. Windows guest path percent-encoding (#683)
- **Why:** Guest FreeBSD-style names like `Arcade Spirits: The New Challengers.dat` are legal on the guest but Windows truncates/fails at `:`. Saves then cannot reload.
- **What:**
  - `HostFsPath.EncodeHostPathSegment` / `DecodeHostPathSegment` (Windows only for mount translation).
  - `NormalizeMountRelativePath` encodes each segment before `Path.Combine`.
  - `getdirentries` decodes host names back to the guest.
  - `SaveDataStorage.Sanitize` uses percent-encoding instead of `_` replacement to avoid collisions (`foo:bar` vs `foo_bar`).
- **Tests:** round-trip colon encoding + updated Sanitize expectations.

### 4. `VulkanDetilePass` pool cap (#618-related)
- **Why:** Free lists retained every size bucket forever; busy loading screens create many texture sizes and native Vulkan memory climbs.
- **What:** Cap 4 buffers per bucket and a total pooled-byte budget (128–256 MiB depending on host). Excess allocations are destroyed on return instead of cached forever.

### 5. `GuestDataPool` / presenter defaults on constrained hosts
- **Why:** Same 16 GiB pressure; keep managed caches smaller by default when the host is constrained.
- **What:** Lower max cached bytes / arrays-per-bucket; presenter pending-work and render-scale defaults read from `HostMemoryBudget` unless env overrides are set.

### 6. `GpuWaitRegistry` hard cap (#639-related)
- **Why:** Producerless waits are never removed by `CollectExpiredRetries` / `CollectDeadlockBroken`, so the waiter table can grow without bound during boot storms.
- **What:** Soft hard-cap at 4096 waiters. When exceeded, drop oldest *orphan* waiters (no recorded producer, no retry deadline, not latched) first; only if still over budget drop oldest remaining. This does **not** force-resume DCBs (still fail-closed for ordering). Tradeoff: orphaned DCBs that were already stuck may be abandoned instead of retaining unbounded host state.
- **Test:** registry stays <= 4096 after flooding 5000 producerless waits. Existing produced-value retention tests still pass.

### 7. CLI `--fullscreen` aliases (#758)
- **Why:** Users asked for `-f` / `--fullscreen`; `--window-mode=fullscreen` already mapped to exclusive.
- **What:** Accept `-f`, `--f`, `-fullscreen`, `--fullscreen` as aliases for exclusive fullscreen. Usage text updated.

## Testing

- Unit: `SaveDataStorageTests`, `HostFsPathEncodingTests`, `GpuWaitRegistryProducedRetentionTests` — **17 passed** on Windows Release.
- Runtime game testing for Demon's Souls / Arcade Spirits: **not yet done on this exact commit** (no dump attached in this session). Please treat runtime claims as unverified until I (or a reviewer) run a title and attach logs showing `Host memory budget:` and save/path behavior.

## Checklist

- [x] I tested my changes (unit tests) or marked remaining runtime testing clearly above.
- [x] I wrote / thoroughly edited this description so each change has a concrete why (AI-assisted development disclosed above).
- [x] I listed testing status above.

## Reviewer notes / possible split

If this is too broad, I can split into:
1. Windows path encoding only (#683)
2. HostMemoryBudget + detile/presenter caps
3. GpuWaitRegistry cap (most controversial correctness tradeoff)

Happy to revise based on maintainer preference.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/SharpEmu.CLI/Program.cs
+++ b/src/SharpEmu.CLI/Program.cs
@@ -266,6 +266,7 @@ internal static partial class Program
 
         Log.Info(BuildInfo.Banner);
         Log.Info(HostSystemInfo.Summary);
+        Log.Info(HostMemoryBudget.Summary);
 
         ebootPath = Path.GetFullPath(ebootPath);
         Console.Error.WriteLine($"[DEBUG] Full path: {ebootPath}");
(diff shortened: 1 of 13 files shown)


MartinCajiao:
Superseded by the split draft PRs for clearer review: path encoding, host memory budgets, and GpuWaitRegistry cap.

PR #776, opened 2026-08-06
[CPU] Recover the register forms of SSE4a EXTRQ/INSERTQ (#328)

soufian3hm:
## Before submitting

## What this fixes

SSE4a is an AMD-only extension. Every Intel host raises `#UD` on it, and so does Rosetta 2 — which is what #328 is about. `DirectExecutionBackend.Amd64Compat.cs` already recovers `EXTRQ`/`INSERTQ` from that fault and resumes the guest, but only for their **immediate** forms. The guard was:

```csharp
if (isExtrq && instruction.OpCount != 3 || isInsertq && instruction.OpCount != 4)
{
    return false;
}
```

Both instructions also have a **register** form (`66 0F 79 /r` and `F2 0F 79 /r`, the `_mm_extract_si64` / `_mm_insert_si64` intrinsics) that takes its length and index from a register rather than from immediates. Those decode to `OpCount == 2`, so the guard rejected them, the recovery declined, and the process died on the opcode. I verified the four encodings against the Iced version this repo pins:

| encoding | bytes | `Mnemonic` | `OpCount` | recovered before |
|---|---|---|---|---|
| `EXTRQ xmm1, 0x28, 0x00` | `66 0F 78 C1 28 00` | `Extrq` | 3 | yes |
| `EXTRQ xmm1, xmm2` | `66 0F 79 CA` | `Extrq` | 2 | **no** |
| `INSERTQ xmm1, xmm2, 0x28, 0x00` | `F2 0F 78 CA 28 00` | `Insertq` | 4 | yes |
| `INSERTQ xmm1, xmm2` | `F2 0F 79 CA` | `Insertq` | 2 | **no** |

The load-time patcher can't cover them either — `Sse4aExtrqBlendPatch.TryMatch` requires the literal bytes `66 0F 78` followed by the immediates `28 00`.

## What changed

`DirectExecutionBackend.Amd64Compat.cs` — the operand-count gate becomes a dispatch over the four `Code` values. The interesting part is that the two register forms disagree about where their control fields live:

- **`EXTRQ xmm1, xmm2`** takes length from `xmm2[5:0]` and index from `xmm2[13:8]` — both in the **low** quadword — and still extracts out of `xmm1` itself.
- **`INSERTQ xmm1, xmm2`** takes length from `xmm2[69:64]` and index from `xmm2[77:72]` — the **high** quadword — while the bits it inserts still come from the low one.

That asymmetry is the whole reason this needed care, so there's a test dedicated to it that fails if the halves are swapped.

`Sse4aBitFieldEmulator.cs` — adds `DecodeRegisterControl`, which splits that length/index pair out of a quadword. It's kept in the emulator rather than the backend for the reason the class comment already gives: the bit math stays testable without the unsafe CONTEXT plumbing.

The extract/insert math itself is untouched. Fields that overrun the quadword (`index + length > 64`, or `length == 0` with a non-zero index — undefined per AMD) are still declined rather than emulated, which matches both the immediate forms and the rule the BMI fallback states for itself: never half-handle an opcode. I also read the destination before writing it, so `EXTRQ xmm1, xmm1` — where the control fields and the extracted data share one quadword — comes out right.

This is strictly additive. The code path it touches previously returned `false` and let the title die, so it cannot regress a title that works today.

## Testing

**N/A** — I have no PS5 title that emits the register forms, and I want to be straight about that rather than imply otherwise. What this does have is direct coverage of the faulting path:

`Sse4aRegisterFormRecoveryTests.cs` (new, 7 tests) fabricates the state the OS hands the `#UD` handler — a Win64 `CONTEXT` carrying the XMM registers, and a RIP pointing at real instruction bytes in probe-visible host memory — and drives the production entry point `TryRecoverAmdCompatInstruction` over it, so decode, bit math and CONTEXT write-back are exercised exactly as on a live fault:

- `EXTRQ xmm1, xmm2` reads control from the source's low quadword, extracts from the destination, zeroes `[127:64]`, advances RIP by 4
- `INSERTQ xmm1, xmm2` reads control from the source's **high** quadword
- a control pair planted in `INSERTQ`'s source *low* quadword is ignored — asserts the result differs from what reading the wrong half would give
- `EXTRQ xmm1, xmm1` (same register as source and destination)
- an overrunning field declines **and** leaves the CONTEXT byte-for-byte untouched
- both immediate forms still recover, as regression guards on the refactor

`Sse4aBitFieldEmulatorTests.cs` (+4) covers `DecodeRegisterControl`: field placement, that bits above 15 are ignored, that the two bits above each 6-bit field are masked off, and that an all-zero control is the length-64 case.

I checked the new tests are load-bearing by reverting the source change and re-running them: 4 of the 7 fail, and they pass again with it restored.

```
```

`Build succeeded`, and 900 tests pass with 0 failures (889 before this branch, +11 new).

## Notes on #328

The originally reported crash — the fixed 12-byte `EXTRQ`+`VPBLENDD` pattern hard-coded against `xmm2` missing PPSA15552's `xmm1` encoding — is genuinely gone; `Sse4aExtrqBlendPatch` now reads the source register from the ModRM r/m field, and the reporter confirmed that on current `main`. This PR closes a different, still-open hole in the same issue: the generic `#UD` fallback that is meant to be the safety net for everything the patcher misses did not actually cover half the instruction's encodings.

The other remaining gap in #328 — the Darwin XMM bridge, so `_posixXmmContextBridged` can become true on macOS — is untouched here. I don't have an Apple Silicon machine, and getting a raw `mcontext` offset wrong is a memory-corruption risk rather than a non-functional patch, so I've left it alone.

## Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Emulation/Sse4aBitFieldEmulator.cs
+++ b/src/SharpEmu.Core/Cpu/Emulation/Sse4aBitFieldEmulator.cs
@@ -5,7 +5,7 @@ namespace SharpEmu.Core.Cpu.Emulation;
 
 /// <summary>
 /// Pure software implementation of the bit-field math behind AMD's SSE4a EXTRQ/INSERTQ
-/// (immediate-form) instructions.
+/// instructions, in both their immediate and register forms.
 ///
 /// The direct-execution backend runs guest PS5 code natively on the host CPU. The PS5's Zen 2
 /// cores implement AMD-only SSE4a (EXTRQ/INSERTQ), but Intel hosts - and Rosetta 2 on Apple
@@ -15,8 +15,8 @@ namespace SharpEmu.Core.Cpu.Emulation;
 /// a different register allocation, a title built with a different compiler version, and so on
 /// - still aborts the title. This class ported from Kyty's
 /// <c>Loader::X64InstructionEmulator::TryEmulateSse4a</c> provides the general bit-field
-/// extract/insert so the illegal-instruction handler can finish *any* immediate-form
-/// EXTRQ/INSERTQ in software and resume, instead of relying on a single hard-coded byte pattern.
+/// extract/insert so the illegal-instruction handler can finish *any* EXTRQ/INSERTQ in software
+/// and resume, instead of relying on a single hard-coded byte pattern.
 ///
 /// The methods operate on plain 64-bit integers rather than the OS CONTEXT record so the bit
 /// math can be unit-tested in isolation; the unsafe CONTEXT/XMM plumbing lives in the backend
@@ -24,6 +24,18 @@ namespace SharpEmu.Core.Cpu.Emulation;
 /// </summary>
 public static class Sse4aBitFieldEmulator
 {
+    /// <summary>
+    /// Splits the length/index pair that the register forms of EXTRQ and INSERTQ read out of a
+    /// source register rather than out of immediates: length in bits [5:0] of the controlling
+    /// quadword, index in bits [13:8]. Which quadword that is differs between the two
+    /// instructions, so the caller selects it - see the backend adapter.
+    /// </summary>
+    public static void DecodeRegisterControl(ulong control, out int length, out int index)
+    {
+        length = (int)(control & 0x3F);
+        index = (int)((control >> 8) & 0x3F);
+    }
+
     public static bool IsValidBitField(int length, int index)
     {
         var len = length & 0x3F;
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Amd64Compat.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Amd64Compat.cs
@@ -9,17 +9,21 @@ namespace SharpEmu.Core.Cpu.Native;
 
 // General software fallback for the AMD-only instructions PS5 titles occasionally emit that a
 // Zen 2-only host implements but Intel hosts (and Rosetta 2 on Apple Silicon) do not:
-//   - SSE4a EXTRQ/INSERTQ, immediate form
+//   - SSE4a EXTRQ/INSERTQ, both the immediate and the register forms
 //   - MONITORX/MWAITX
 //
-// This is a direct port of Kyty's Loader::X64InstructionEmulator (TryEmulateSse4a /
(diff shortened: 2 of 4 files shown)


soufian3hm:
**Update: this is now verified end-to-end on real hardware**, which the description could not claim when I opened it.

I marked Testing as N/A because I had no title emitting the register forms. I still do not — but I built a synthetic bare ELF whose payload is exactly the faulting sequence, and once #781 landed locally (a REX.B bug was stopping *any* fault from reaching the managed handler, so this recovery could never run) the whole path executes:

```asm
mov  rax, 0x123456789ABCDEF0
movq xmm1, rax          ; value to extract from
mov  rax, 0x0810        ; length = 0x10 (bits [5:0]), index = 0x08 (bits [13:8])
movq xmm2, rax          ; control register
extrq xmm1, xmm2        ; 66 0F 79 CA - SSE4a register form, #UD on Intel
movq rax, xmm1
ret
```

On an Intel i7-11370H (no SSE4a), with #776 + #781:

```
[LOADER][INFO] Patched 0 TLS loads, 0 TLS stores, 0 stack-canary accesses, 0 SSE4a EXTRQ blends
[LOADER][INFO] Host lacks SSE4a EXTRQ/INSERTQ used by the guest; emulating those instructions in software.
[LOADER][INFO] Guest returned: 48350
```

Three things worth drawing out:

1. **The load-time patcher matched nothing** (`0 SSE4a EXTRQ blends`), confirming the register form is outside what `Sse4aExtrqBlendPatch` can cover — which is the premise of this PR.
2. **The recovery fired** and announced itself.
3. **48350 == 0xBCDE**, which is the arithmetically correct answer: extracting 16 bits at index 8 from `0x123456789ABCDEF0` gives `(v >> 8) & 0xFFFF` = `0xBCDE`. So the guest resumed with the right value in rax, not merely without crashing.

That last point matters most for review: it independently confirms the encoding detail this PR turns on — that `EXTRQ xmm1, xmm2` takes its length and index from the **low** quadword of the source. Had I read that from the high quadword (INSERTQ's rule), the result would have decoded as length=0, index=0 and returned the full `0x123456789ABCDEF0` instead.

Without #781 this same guest hangs at 100% CPU with no output, so the two changes are independent but this one is only observable once that one is in. No code change here — evidence only.

Generator for the guest, if useful for reproducing: it emits the bare-ELF header shape `Load_AcceptsBareDecryptedElf` already asserts, with the payload above. Happy to contribute it as a test fixture if maintainers want it.

PR #784, opened 2026-08-07
[Shader] Decode the VOP3-encoded VOPC and VOP1 aliases

soufian3hm:
## Before submitting

## What this fixes

VOP3 re-encodes the **entire VOPC opcode space at `0x000-0x0FF`** and the **entire VOP1 space at `0x180-0x1FF`**, so those instructions can carry what the short forms have no room for: abs/neg on a source, a clamp, or a destination other than VCC. The compiler picks the `_e64` form for precisely those cases.

Neither range was decoded. The VOP3 table's lowest key is `0x101`, and it has no entries at all in `0x180-0x1FF`, so all of them fell through to:

```csharp
_ => $"Vop3Raw{opcode:X3}",
```

On current `main`, a three-instruction program:

```asm
v_cmp_lt_f32_e64 s[4:5], |v0|, v1
v_rcp_f32_e64    v2, v0
s_endpgm
```

fails with:

```
block=0x0: pc=0x0 Vop3Raw001: unsupported vector opcode Vop3Raw001
```

That is a hard fail with no fallback. `AgcExports.TryTranslateGuestDraw` leaves `state.TranslatedDraw = null` and the draw is dropped, deduplicated to one `[COMPAT][SHADER]` line per distinct failure. **One such instruction anywhere in a pixel shader deletes every draw that binds it, for the rest of the session, with no crash and no log spam.**

There is corroborating evidence inside the tree that this range is real and reachable: VOP3 opcode `0x101` (`VCndmaskB32`, the VOP2-in-VOP3 alias) was already added by hand, one opcode at a time, rather than by fixing the range.

## What changed

Four edits. They are **not separable** — see the note below.

1. **`Gen5ShaderTranslator.DecodeVop1` / `DecodeVopc`** — the name tables are hoisted into `Vop1Name` / `VopcName` so they can be shared. No behaviour change.
2. **`DecodeVop3`** — both alias ranges route through those tables. The VOP3-specific `unknown-vop3 op=0x%03X` message is kept deliberately: both tables still have holes (f64 compares, several VOP1 slots), and reporting the aliased VOPC/VOP1 number would hide *which VOP3 opcode actually needs implementing* — the one piece of information needed to fix the next title.
3. **`CreateInstruction`** — a VOP3-encoded compare writes an SGPR pair, which is the whole reason the compiler chose the long form, so `VCmp*` now takes a scalar destination from the vdst byte. `VCmpx` is excluded: GFX10 makes it EXEC-only and both emitters already route it to the wave mask. `VReadfirstlaneB32` is included because `TryEmitVectorAlu` requires a scalar destination and would otherwise start failing the moment edit 2 made the `_e64` form decodable.
4. **Both emitters** (`Gen5SpirvTranslator.Alu.cs`, `Gen5MslTranslator.Alu.cs`) — prefer the decoded destination over the hardcoded VCC fallback.

### Why this cannot be split

Landing the decoding without the emitter changes is **strictly worse than today**. It converts a loud, deduplicated hard failure into silent register corruption: the compare result goes to VCC while the named SGPR pair keeps a stale mask, producing wrong branches and wrong pixels with zero diagnostics.

The Metal edit is part of that, not an optional macOS extra — the decoder is shared, so shipping without it corrupts that backend in exactly the same way.

## Testing

**N/A** for game testing — I have no PS5 title, and I want to be straight about what that means for this PR (see the honest assessment below).

`Gen5ShaderTranslatorTests.cs` (+5):

- `v_cmp_lt_f32_e64` decodes to `VCmpLtF32` with a **ScalarRegister** destination of `s4`
- `v_rcp_f32_e64` decodes to `VRcpF32` with a **VectorRegister** destination
- a VOP3-encoded `VCmpx` keeps its vector destination, pinning edit 3's exclusion
- a theory over both alias ranges asserting nothing decodes to a `Vop3Raw` stub, and that opcodes the tables do not cover fail with an error containing `unknown-vop3` (pins the error-message contract)

`Gen5Vop3AliasSpirvTests.cs` (new, 4) compiles the real thing end to end through `TryCreateState` → `Gen5ShaderScalarEvaluator.TryEvaluate` → `Gen5SpirvTranslator.TryCompileComputeShader`:

- each alias form compiles to SPIR-V, and both together
- **two compares differing only in which SGPR pair they name produce different modules** — if the emitter still hardcoded VCC they would be byte-identical, so this test specifically catches edits 1+2 landing without 3+4

Confirmed load-bearing. Reverting all three source files and re-running gives exactly the stubs:

```
Vop3EncodedCompareDecodesThroughTheVopcTable   Actual: "Vop3Raw001"
AliasRangesNeverDecodeToAnUnnamedStub(1)       Opcode = Vop3Raw001
AliasRangesNeverDecodeToAnUnnamedStub(193)     Opcode = Vop3Raw0C1
AliasRangesNeverDecodeToAnUnnamedStub(417)     Opcode = Vop3Raw1A1
AliasRangesNeverDecodeToAnUnnamedStub(426)     Opcode = Vop3Raw1AA
Vop3EncodedCmpxKeepsItsVectorDestination       String: "Vop3Raw011"
```

```
```

`Build succeeded`, 900 tests pass, 0 failures (889 before this branch, +11 new). That includes the Metal suite (27) and the ShaderCompiler suite (30), both green.

## Honest assessment

**I am not claiming this makes any title playable, and I would rather say so than have this oversold.**

What is established: these encodings hard-fail today, the failure silently deletes draws, and after this they compile. That is measured, not argued.

What is **not** established: that any specific title emits them. The structural case is reasonable — AMD's backend selects `_e64` when a compare carries source modifiers, a clamp, or a non-VCC SGPR destination, and non-VCC destinations come from nested or short-circuited conditionals, which trivial UI shaders do not have — and the hand-added `0x101` entry suggests someone hit this range already. But "structurally motivated" is not "measured on a title I ran".

The realistic shape of success is that some previously-vanished geometry appears and the `[COMPAT][SHADER]` log starts naming a *different* missing opcode. `v_alignbit_b32` (VOP3 `0x14E`) is the strongest candidate for what shows up next — grepping the tree for `Alignbit` returns nothing, so it is missing from the decoder table *and* both emitters and needs real emitter work rather than a table entry. I have deliberately left that out of scope.

If someone with a title can run this and paste the resulting `[COMPAT][SHADER]` lines, that would be far more valuable than my speculation about which opcode is next.

## Checklist

### DIFF
--- a/src/SharpEmu.ShaderCompiler.Metal/Gen5MslTranslator.Alu.cs
+++ b/src/SharpEmu.ShaderCompiler.Metal/Gen5MslTranslator.Alu.cs
@@ -664,10 +664,17 @@ public static partial class Gen5MslTranslator
             }
             else
             {
-                var target = instruction.Control is Gen5SdwaControl
-                    { ScalarDestination: { } scalarDestination }
-                    ? scalarDestination
-                    : VccLoRegister;
+                // Same rule as the SPIR-V emitter: the decoder shares this instruction stream, so
+                // a VOP3-encoded compare must reach its named SGPR pair here too. Defaulting to
+                // VCC would silently produce wrong branches with no diagnostic.
+                var target =
+                    instruction.Destinations.Count > 0 &&
+                    instruction.Destinations[0].Kind == Gen5OperandKind.ScalarRegister
+                        ? instruction.Destinations[0].Value
+                        : instruction.Control is Gen5SdwaControl
+                            { ScalarDestination: { } scalarDestination }
+                            ? scalarDestination
+                            : VccLoRegister;
                 StoreMaskBit(target, active);
             }
 
--- a/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
+++ b/src/SharpEmu.ShaderCompiler.Vulkan/Gen5SpirvTranslator.Alu.cs
@@ -1662,10 +1662,19 @@ public static partial class Gen5SpirvTranslator
             }
             else
             {
-                var compareDestination = instruction.Control is Gen5SdwaControl
-                    { ScalarDestination: { } scalarDestination }
-                    ? scalarDestination
-                    : 106u;
+                // Prefer the decoded destination: a VOP3-encoded compare names an SGPR pair, and
+                // writing VCC instead would leave that pair stale while corrupting a register the
+                // shader may still be relying on. Plain VOPC leaves Destinations empty and keeps
+                // the VCC default; SDWA is subsumed because it already records a scalar
+                // destination when it is not VCC.
+                var compareDestination =
+                    instruction.Destinations.Count > 0 &&
+                    instruction.Destinations[0].Kind == Gen5OperandKind.ScalarRegister
+                        ? instruction.Destinations[0].Value
+                        : instruction.Control is Gen5SdwaControl
+                            { ScalarDestination: { } scalarDestination }
+                            ? scalarDestination
+                            : 106u;
                 StoreWaveMask(compareDestination, activeCondition);
             }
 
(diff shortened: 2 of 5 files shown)


foufouadi:
i went through the changes and they look good to me.The decoder/emitter updates are consistent, and the tests cover the important cases.I tested Astro Bot, Ghost of Tsushima, and Ghost of Yōtei, and none of them seem to hit these VOP3 alias opcodes.
Out of curiosity, how did you come across this issue? Was it from someone else's logs or while auditing the decoder? If there's a game known to use these opcodes, I'd be happy to test it and see the impact.

PR #793, opened 2026-08-07
[Kernel] Write the full 64-bit handle in sceKernelCreateSema

soufian3hm:
## Problem

`SceKernelSema` is a 64-bit opaque pointer, so `sceKernelCreateSema`'s out-parameter is eight bytes. It writes four:

```csharp
if (!TryWriteUInt32(ctx, semaphoreAddress, handle))
```

The guest's upper half keeps whatever was in that variable beforehand — in practice, uninitialised stack or struct padding.

It is the odd one out among the kernel-object creators in this tree:

| export | out-parameter write |
|---|---|
| `sceKernelCreateEventFlag` | `ctx.TryWriteUInt64(outAddress, handle)` |
| `sceKernelCreateEqueue` | `ctx.TryWriteUInt64(outAddress, handle)` |
| `scePthreadMutexInit` | `TryWriteUInt64Compat(ctx, mutexAddress, handle)` |
| **`sceKernelCreateSema`** | **`TryWriteUInt32(ctx, semaphoreAddress, handle)`** |

## Severity, honestly

**No title is known to break on this today.** Every semaphore export here takes the handle in RDI and masks it back to 32 bits, so the value round-trips through our own HLE regardless of what the upper half holds. I checked each consumer before writing this.

So the deviation is from the ABI, not from our internal consistency. What it leaves behind is a guest variable that is half-initialised: a guest that copies it, compares two of them, or null-checks the full 64 bits is reading uninitialised data, and the resulting bug would be nondeterministic and extremely unpleasant to trace back to here. I would rather not leave that for whoever hits it.

I am flagging the severity plainly rather than dressing this up as a fix for a specific title, because it isn't one.

## Change

One line, plus a comment recording why the width matters and which siblings it now matches.

**The POSIX side is deliberately unchanged.** `sem_init` writes a 32-bit cookie into an opaque `sem_t` and `TryGetPosixSemaphoreHandle` reads it back at the same width — that is a consistent pair, and a legitimate design for a struct the guest never inspects. Only the Orbis entry point is wrong.

## Tests

`KernelCreateSemaOutParameterTests`, three cases:

- the out-parameter is pre-filled with `0xDEADBEEFDEADBEEF` before the call, so a partial write shows up as surviving garbage rather than hiding behind zeroed memory
- the handle written out is one `sceKernelWaitSema` and `sceKernelDeleteSema` actually accept — a wider write is no good if it writes the wrong thing
- two semaphores get distinct handles

Verified load-bearing: restoring the 32-bit write fails `WritesTheWholeSixtyFourBitHandleVariable` and leaves the other two passing, which is the correct signature for a change that adds width without altering the handle.

`dotnet build SharpEmu.slnx -c Release` clean, `dotnet test SharpEmu.slnx -c Release --no-build` → **892 passed, 0 failed**.

### DIFF
--- a/src/SharpEmu.Libs/Kernel/KernelSemaphoreCompatExports.cs
+++ b/src/SharpEmu.Libs/Kernel/KernelSemaphoreCompatExports.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2026 SharpEmu Emulator Project
+﻿// Copyright (C) 2026 SharpEmu Emulator Project
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 using System.Buffers.Binary;
@@ -71,7 +71,11 @@ public static class KernelSemaphoreCompatExports
             Count = initialCount,
         };
 
-        if (!TryWriteUInt32(ctx, semaphoreAddress, handle))
+        // SceKernelSema is a 64-bit opaque pointer, so the whole variable has to be written.
+        // Writing only the low half leaves the guest's upper four bytes holding whatever was
+        // there, which every other kernel-object creator here avoids: sceKernelCreateEventFlag
+        // and sceKernelCreateEqueue both write 64 bits.
+        if (!ctx.TryWriteUInt64(semaphoreAddress, handle))
         {
             _semaphores.TryRemove(handle, out _);
             return SetReturn(ctx, OrbisGen2Result.ORBIS_GEN2_ERROR_MEMORY_FAULT);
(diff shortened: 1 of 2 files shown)


par274:
I see that some of your PRs don't adhere to the contributing guidelines; please read them first and stick to the PR template.

Also, none of these changes can be proven without game testing, meaning I can't merge them into main; this is how you can truly prove that something has been fixed or added.

soufian3hm:
ngl bro i have claude max so i told it to contribute if they aint helpful my bad g

PR #798, opened 2026-08-07
Fix VEH trampoline lock register encoding

### LINKED ISSUE #779: Any guest fault hangs the emulator at 100% CPU with no vectored handler entered (minimal repro, no game needed)

Gabresixx:
Fixes #779.

I tracked the hang down to the exception trampoline. It was supposed to emit `lock cmpxchg [r9], r10`, but the REX prefix was `0x4C`, so the memory operand ended up using `rcx` instead of `r9`.

During a guest fault, `rcx` could be something invalid like `0x8`, so the VEH would fault again inside itself before reaching the managed handler and keep looping.

The fix is just changing the two affected prefixes from `0x4C` to `0x4D`, which sets `REX.B` and makes the operand use `r9` as intended.

I also added regression tests for the emitted encoding and small synthetic RET, UD2, and access-violation guests on Windows x64.

## Testing

Synthetic RET/UD2/access-violation reproducers on Windows x64: passed.  
Full Release suite: 893 passed, 0 failed.  
Release build: 0 warnings, 0 errors.  
No commercial games were used; this is a synthetic reproducer for #779.

## Checklist

- [x] I tested my changes.
- [x] I listed the game(s) I tested or marked testing as N/A.

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -2929,7 +2929,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 		EmitByte(code, ref offset, 0x0F); EmitByte(code, ref offset, 0x85);
 		int hostPauseJump = offset;
 		EmitUInt32(code, ref offset, 0u);
-		EmitByte(code, ref offset, 0xF0); EmitByte(code, ref offset, 0x4C);
+		EmitByte(code, ref offset, 0xF0); EmitByte(code, ref offset, 0x4D);
 		EmitByte(code, ref offset, 0x0F); EmitByte(code, ref offset, 0xB1); EmitByte(code, ref offset, 0x11); // lock cmpxchg [r9], r10
 		EmitByte(code, ref offset, 0x0F); EmitByte(code, ref offset, 0x85);
 		int hostRetryJump = offset;
@@ -3012,7 +3012,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 		EmitByte(code, ref offset, 0x0F); EmitByte(code, ref offset, 0x85);
 		int guestPauseJump = offset;
 		EmitUInt32(code, ref offset, 0u);
-		EmitByte(code, ref offset, 0xF0); EmitByte(code, ref offset, 0x4C);
+		EmitByte(code, ref offset, 0xF0); EmitByte(code, ref offset, 0x4D);
 		EmitByte(code, ref offset, 0x0F); EmitByte(code, ref offset, 0xB1); EmitByte(code, ref offset, 0x11);
 		EmitByte(code, ref offset, 0x0F); EmitByte(code, ref offset, 0x85);
 		int guestRetryJump = offset;
--- a/tests/SharpEmu.Libs.Tests/Cpu/ImportTrampolineAbiTests.cs
+++ b/tests/SharpEmu.Libs.Tests/Cpu/ImportTrampolineAbiTests.cs
@@ -60,6 +60,22 @@ public sealed class ImportTrampolineAbiTests
         }
     }
 
+    [Fact]
+    public unsafe void ExceptionHandlerTrampoline_UsesR9ForBothManagedEntryLocks()
+    {
+        if (!OperatingSystem.IsWindows() || RuntimeInformation.ProcessArchitecture != Architecture.X64)
+        {
+            return;
+        }
+
+        var code = CreateExceptionHandlerTrampolineBytes();
+        var expected = new byte[] { 0xF0, 0x4D, 0x0F, 0xB1, 0x11 };
+        var incorrect = new byte[] { 0xF0, 0x4C, 0x0F, 0xB1, 0x11 };
+
+        Assert.Equal(2, CountOccurrences(code, expected));
+        Assert.Equal(0, CountOccurrences(code, incorrect));
+    }
+
     private static unsafe byte[] CreateTrampolineBytes()
     {
         var backend = (DirectExecutionBackend)RuntimeHelpers.GetUninitializedObject(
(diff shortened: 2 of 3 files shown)


foufouadi:
Tested on latest main. It fixes Silent Hill, which now reaches Vulkan and presents frames.
However, it reproducibly crashes:
- GTA V: 0xC0000005 at 0x801561D30
- Minecraft: 0xC0000005 at 0x80D104AA0
Without #798, GTA reaches the cinematic and Minecraft stalls instead of crashing. The trampoline fix likely needs adjustment before merging.
image

PR #801, opened 2026-08-08
fix(logging): cap FileLogSink size and roll over via SHARPEMU_LOG_MAX_BYTES

ajaysaagar-dev:
## Summary
`FileLogSink` (used when `SHARPEMU_LOG_FILE` is set) opened its target file
in Append mode with no size limit or rotation. Long-running sessions, or
repeated runs pointed at the same fixed log path, could grow the file
unbounded.

This does **not** affect the default logging path documented in the README
(`--log-file` / piped via `Tee-Object` or `tee`), which already creates a
fresh, timestamped log file per session
(`{titleId}-{yyyyMMdd-HHmmss}.log`) and had no unbounded growth risk.

### `FileLogSink.cs` (+64 lines)
- Added an optional `maxBytes` constructor parameter (default `0` = no cap,
  preserving existing behavior for all current call sites).
- Stores the file path so it can be renamed on rollover; `_writer` is no
  longer `readonly` so it can be replaced.
- `Write()` checks `_writer.BaseStream.Position >= _maxBytes` after the
  existing error-level eager flush, and triggers rollover when crossed —
  using the stream's actual position rather than a manual byte counter.
- `TryRollover()` flushes, disposes, moves `_path` → `_path + ".1"`
  (overwriting any previous `.1`), and opens a fresh writer. If the
  move/create fails, it falls back to re-opening in append mode; if that
  also fails, the sink marks itself disposed so subsequent `Write()` calls
  no-op instead of throwing.

### `SharpEmuLog.cs` (+14 lines)
- `ResolveMaxBytesFromEnvironment()` reads `SHARPEMU_LOG_MAX_BYTES`, parses
  it as a byte count, and returns `0` (no cap) if absent or invalid.
- `ResolveSinkFromEnvironment()` now passes `maxBytes:` through to the
  `FileLogSink` constructor.

## Scope
- Only affects file output policy — no verbosity, sink topology, or
  console-path changes.
- `--log-file` / `TeeTextWriter` path is unchanged (already per-session by
  design).
- `SHARPEMU_LOG_FILE` path now respects the cap.
- Any other `FileLogSink` call site (e.g. from the GUI) only gets capping
  if it explicitly passes `maxBytes`; existing call sites with no arg are
  unaffected.
- Retention is a single rolled backup (`foo.log` → `foo.1.log`), not a
  multi-generation history — kept intentionally simple for this fix.

## Usage
```powershell
# Cap the log at 50 MiB (52,428,800 bytes); rolls over to SharpEmu.log.1 once crossed
$env:SHARPEMU_LOG_FILE = "SharpEmu.log"
$env:SHARPEMU_LOG_MAX_BYTES = "52428800"
.\SharpEmu.exe "C:\path\to\eboot.bin"
```
Without `SHARPEMU_LOG_MAX_BYTES` set, behavior is identical to before this
change.

## Testing
[Fill in what was actually verified — e.g. "manually set
SHARPEMU_LOG_MAX_BYTES to a small value and confirmed rollover to
SharpEmu.log.1 after the threshold was crossed; confirmed default
(unset) behavior is unchanged."]

## Out of scope
- No change to what gets logged or at what verbosity.
- `SHARPEMU_LOG_FILE` itself isn't currently documented in the
  README/docs — happy to open a small follow-up to document both env vars
  if maintainers want them public-facing.

### DIFF
--- a/src/SharpEmu.Logging/FileLogSink.cs
+++ b/src/SharpEmu.Logging/FileLogSink.cs
@@ -11,13 +11,18 @@ public sealed class FileLogSink : ISharpEmuLogSink, IDisposable
     private static readonly TimeSpan FlushInterval = TimeSpan.FromMilliseconds(500);
 
     private readonly object _sync = new();
-    private readonly StreamWriter _writer;
+    private readonly string _path;
+    private readonly long _maxBytes;
+    private StreamWriter _writer;
     private readonly Timer _flushTimer;
     private bool _disposed;
 
-    public FileLogSink(string path, bool append = true, bool includeTimestamp = true)
+    public FileLogSink(string path, bool append = true, bool includeTimestamp = true, long maxBytes = 0)
     {
         ArgumentException.ThrowIfNullOrWhiteSpace(path);
+
+        _path = path;
+        _maxBytes = maxBytes > 0 ? maxBytes : 0;
 
         var directory = Path.GetDirectoryName(path);
         if (!string.IsNullOrEmpty(directory) && !Directory.Exists(directory))
@@ -91,6 +96,11 @@ public sealed class FileLogSink : ISharpEmuLogSink, IDisposable
             {
                 _writer.Flush();
             }
+
+            if (_maxBytes > 0 && _writer.BaseStream.Position >= _maxBytes)
+            {
+                TryRollover();
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


ajaysaagar-dev (commented):
Summary comment for maintainers, since I made a few inline decisions worth flagging explicitly:

**Design choice — single-generation rollover:** `TryRollover()` only keeps
one backup (`foo.log` → `foo.1.log`), not a multi-generation history. I kept
this intentionally minimal for the fix's scope — happy to extend to N backups
if that's preferred, just say the word.

**Fallback behavior:** if the file move/create during rollover fails, the
sink falls back to re-opening in append mode rather than crashing the
emulator over a logging failure. If that also fails, the sink marks itself
disposed and subsequent writes no-op silently rather than throwing. Wanted
to flag this tradeoff explicitly — logging failures degrade gracefully
instead of taking down the session, but that also means a broken log sink
fails silently. Open to feedback if you'd prefer this surfaced differently
(e.g. a one-time console warning).

**Backward compatibility:** `maxBytes` defaults to `0` (no cap) at every
existing call site, so this is opt-in only via the new
`SHARPEMU_LOG_MAX_BYTES` env var — no behavior change for anyone not using it.

**Testing done:**
- Verified default (unset `SHARPEMU_LOG_MAX_BYTES`) behavior is unchanged.
- Set a small `SHARPEMU_LOG_MAX_BYTES` value and confirmed rollover to
  `SharpEmu.log.1` once the threshold was crossed, with logging continuing
  uninterrupted in the new file.
- Did not add an automated test for `FileLogSink` — I didn't find existing
  test coverage for this class to extend. Happy to add one if there's a
  preferred pattern/harness to follow.

Issue #256, opened 2026-07-16
Unimplemented: sceSaveDataSetParam + sceSaveDataDelete + NpUniversalDataSystem (Dreaming Sarah crashes on new game)


jackby03:
## Symptoms
- Dreaming Sarah (PPSA02929) boots to menu (video + audio + input working)
- Crashes when starting a new game

## Log evidence

```
[LOADER][WARN] Import#542 unresolved: nid=85zul--eGXs ret=0x00000008002DFC38 rdi=0x00000008004C6AD0 rsi=0x0000000000000000 rdx=0x00007FFFE01FF448 rcx=0x0000000000000530
[LOADER][WARN] Import#546 unresolved: nid=S1GkePI17zQ ret=0x00000008002DFDEA rdi=0x00007FFFE01F7420 rsi=0x0000000804110480 rdx=0x00000008002F5CFF rcx=0x000000000000001E
[LOADER][WARN] Import#1070 unresolved: nid=4llLk7YJRTE ret=0x00000008002DA980 rdi=0x0000000000000000 rsi=0x00000008002F797D rdx=0x0000000000000000
```

NIDs identified in aerolib.bin:
- 85zul--eGXs -> sceSaveDataSetParam
- S1GkePI17zQ -> sceSaveDataDelete
- 4llLk7YJRTE -> sceNpUniversalDataSystemEventPropertyArraySetString

## Context
Dreaming Sarah is the most successful test case: 49K imports, 31 threads, menu fully functional. Implementing these would likely allow the game to progress past the menu into gameplay.

PR #824, opened 2026-08-17
[Codec/Native] Real H.264 decode for sceVideodec2, fix TLS loader missing the main module

foufouadi:
## Summary

This PR fixes two unrelated issues: sceVideodec2 never produced a real picture, and the TLS load patcher could miss the main game module entirely on boot.

sceVideodec2 (the hardware compute-based H.264 decoder) was capability-only: sceVideodec2CreateDecoder/Decode/Flush all resolved and returned success, so the game's video pipeline ran end-to-end with no errors, but Decode never actually decoded anything and no picture was ever reported ready. The result is a video player that "completes" playback normally while the screen stays black  confirmed on Ghost of Yotei's intro cinematic, which reaches every Videodec2 call correctly but never gets a frame back.

The fix opens a real FFmpeg H.264 session per decoder handle (Videodec2Decoder) and feeds it the same Annex-B access units the game already hands to sceVideodec2Decode. Decode and presentation pacing run on their own threads, never the guest thread, so a whole clip isn't decoded faster than it can be displayed . Decoded frames convert to BGRA and go straight to VulkanVideoPresenter, bypassing guest memory the same way the existing Bink2 path does. If FFmpeg is unavailable or a given decoder fails to open, every export falls back to the original no-picture stub for that handle instead of crashing.

While wiring this up, VulkanVideoPresenter turned out to already be dropping decoded frames on the floor: Submit() overwrote a single "latest wins" slot, and the render loop didn't always poll it before the next call overwrote it again  a pre-existing bug the video path never exercised before now. Presentations are now queued the same bounded way guest-image flips already are, so a render tick that falls a frame or two behind still catches up instead of silently losing frames.

Separately, DirectExecutionBackend's TLS load patcher does a single scan starting at _entryPoint for FS:[0] loads to patch. On titles where the entry point resolves to a separate bootstrap allocation rather than the main module, that scan never reaches the main module's TLS loads at all, and lazily-committed pages patched in after the scan were never re-scanned either  both leave real FS:[0] loads unpatched, which show up as an early mutex-spin boot stall (reproduced on Demon's Souls, stuck at import #256 with 0 TLS loads patched).

The fix scans both the entry point's own allocation and the standard PS5/PS4 main image base unconditionally, and re-scans each newly-committed executable range as it's touched instead of only at the initial one-shot scan.

## Testing

Ghost of Yotei (PPSA26344), compared against upstream/main, same recipe to reach the intro cinematic on both:

https://github.com/user-attachments/assets/84ed7335-a1e3-41f8-ab15-43f0eb56ff7a

- sceVideodec2Decode calls producing a real picture: 0 vs 1166
- Intro cinematic visible on screen: no vs yes (confirmed visually)
- New decode errors/exceptions: 0 vs 0

Demon's Souls (PPSA01341), 90s run compared against upstream/main:

- TLS loads patched: 0 vs 4544
- Last import reached before stall/continuing: #256 vs #11,512,955+
- Stall watchdog timeout (exit code 4): yes vs no
- TLS fix isolated from the video decode change (separate build, video changes reverted): same result, confirms it's the TLS patcher and not an interaction with the Videodec2 work

Ghost of Tsushima (PPSA02225), Astro Bot (PPSA21564), and Quake (PPSA01880) were also tested; behavior is byte-for-byte identical to upstream/main on all three (same known unresolved-import warnings, same stall points, no new errors). Tsushima's import table does reference libSceVideodec2 (its stubs are set up) but the game never reaches a call to it before an unrelated, pre-existing stall; Astro Bot and Quake never reference the library at all, so none of the exports modified by this PR were exercised during those runs.

## Checklist


foufouadi:
Added a small follow-up commit exposing SHARPEMU_FORCE_SUBMIT_ORPHAN_PREAMBLES in the GUI. The Ghost of Yotei intro in this PR currently works with this flag enabled, so I added a checkbox for it instead of requiring manual environment variable setup. It's available in both the global and per-game settings.

par274:
Actually, in BK2, the game engine itself was supposed to decode it. This is the correct.

foufouadi:
ah, that makes sense,thanks for the clarification

PR #836, opened 2026-08-22
fix(agc,video): CMASK/DCC metadata state machine  , remove heavy heavy motion trails in DeadCell

angleyanalbedo:
## Before submitting

I have read and follow the contribution guidelines.

## Summary

Fixes for Dead Cells (PPSA15552): heavy motion trails in the dungeon view. The changes form one closely related set of AGC / video-out fixes developed against this game, growing into a general CMASK/DCC metadata state machine that mirrors the hardware's fast-clear semantics:

1. **CMASK/DCC meta-state machine**: a per-surface ledger keyed by colour-buffer base address, tracking the colour buffer -> metadata mapping (CMASK base/EXT and DCC base/EXT registers) together with the CB CLEAR_WORD0/1 registers. Registration alone does not clear anything.
2. **Mark-clean on real clear events**: a surface is marked "all clear" only when something actually clears its metadata — a DMA constant fill into the registered metadata window, a compute write into it, or an EliminateFastClear draw. This mirrors shadPS4's meta handling in vk_rasterizer.cpp.
3. **Consumption as LoadOp.Clear**: the first render pass that binds a marked surface starts from LoadOp.Clear with the clear colour decoded from CLEAR_WORD0/1 (format-aware unpack for R8G8B8A8_UNORM and R16G16B16A16_FLOAT), then marks the surface dirty so later passes in the same frame load existing contents. CPU-backed images are never cleared this way.
4. **Frame-boundary mark at guest flip**: at every guest flip (the frame boundary inside the submission stream) all registered surfaces are marked "all clear". This is the per-surface successor of the earlier heuristic that reset only the first multi-attachment group per frame, which has been removed. On hardware, per-frame surfaces are re-cleared through fast-clear metadata every frame; Dead Cells in particular never programs the CMASK/DCC registers yet its scene faces are cleared per frame on real hardware, so register tracking alone was insufficient and the flip-time mark restores that per-frame clear. If a game's own clear (EFC draw or full-screen clear) touches a surface first, consumption happens on that pass, so nothing is double-cleared.
5. **Thread safety**: the ledger is guarded against concurrent access from the GPU command parse thread and the render thread; clear-word decoding and consumption are consistent under one lock.

## Testing

- Dead Cells (PPSA15552) - dungeon-view trails eliminated with no flicker. the character still miss need more invest
- SharpEmu.Libs unit test suite passes (834 tests).

## Checklist

- [x] I tested my changes use `dotnet test`
- [x] I check this pull request description myself .
- [x] I listed the game(s) I tested 'DeadCell`.

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcExports.cs
+++ b/src/SharpEmu.Libs/Agc/AgcExports.cs
@@ -118,6 +118,9 @@ public static partial class AgcExports
     // Multiple producers can share one target label; last-writer-wins would
     // starve waits on the others.
     private static readonly Dictionary<ulong, List<ulong>> _cbReleaseMemTargets = new();
+    // CMASK state tracking: tracks which CMASK addresses are "all clear".
+    // Keyed by CMASK address, value is true if marked as cleared.
+    private static readonly Dictionary<ulong, bool> _cmaskClearedState = new();
     // header -> {ring base, write cursor} of the last submitted slice.
     // Submissions stay cursor-bounded since rings aren't zeroed. Lap
     // distinguishes a stale cursor from a previous pass over the same base.
@@ -1095,6 +1098,7 @@ public static partial class AgcExports
     private const uint CbColor0Base = 0x318;
     private const uint CbColorRegisterStride = 15;
     private const uint CbColor0Info = 0x31C;
+    private const uint CbColor0Cmask = 0x320;
     private const uint CbColor0ClearWord0 = 0x323;
     private const uint CbColor0ClearWord1 = 0x324;
     private const uint CbColor0BaseExt = 0x390;
@@ -8072,6 +8076,7 @@ public static partial class AgcExports
         var hasPsInputAddr = state.CxRegisters.TryGetValue(SpiPsInputAddr, out var psInputAddr);
         state.UcRegisters.TryGetValue(VgtPrimitiveType, out var primitiveType);
         var renderTargets = GetRenderTargets(state.CxRegisters);
+        TrackCmaskAddresses(state.CxRegisters, renderTargets);
         var drawSequence = ++gpuState.WorkSequence;
         if (state.PendingTargetlessDraw is { } stalePendingDraw)
         {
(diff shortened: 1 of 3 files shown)


par274:
Hi, I was test this PR.

resim

Some graphical errors have been fixed, but the character is still not appearing as you mentioned.

angleyanalbedo:
Sorry for the confusion! I used AI to help draft the PR summary and skimmed through it too quickly, missing the claim that characters were fully fixed. To clarify: this PR only targets the motion trails/ghosting . The missing character issue is indeed still present and will need separate investigation. I'll update the PR description.

angleyanalbedo:
sorry about that now i rewite title and i check the pr decription myself

angleyanalbedo:
it jus clear textture problem still

angleyanalbedo:
I captured and verified the frames in NVIDIA Nsight to trace the ghosting and G-Buffer rendering behavior:

Nsight Profiling Findings:

In the pipeline state, every render target pass was defaulting to VK_ATTACHMENT_LOAD_OP_LOAD.

The Geometry view confirmed that full-screen vertex meshes were being submitted, but the actual G-Buffer targets (Color 0, 1, 2) were only rasterized within a smaller sub-rect (642x362 viewport/scissor) across multiple draw calls rather than covering the whole framebuffer in a single pass.

Because uncovered regions retain stale pixels from previous frames across consecutive draws, ghosting and visual artifacts occur on reused VkImage allocations.

Analysis on Clearing Behavior & Dead Cells:

Forcing an unconditional VK_ATTACHMENT_LOAD_OP_CLEAR at the start of every frame technically hides the ghosting, but doing so unconditionally across all titles is unsafe. Titles relying on temporal accumulation (TAA, motion blur) or incremental/dirty-rect rendering would be broken by naive per-frame clears.

Dead Cells does not configure or clear hardware metadata like CMASK/DCC. Instead, it relies on guest CP DMA fills / compute routines to zero-initialize buffer memory. Because these DMA fills directly alter guest RAM without invalidating or clearing the underlying VkImage, the host Vulkan render targets stay dirty.

Why Implement the CMASK/Meta Surface State Machine:

While this specific issue could be temporarily masked with a few lines of heuristics, implementing a dedicated AMD metadata state machine (MetaSurfaceState) matches how GCN/RDNA hardware tracks fast-clears per surface.

This PR lays the proper architectural foundation for both CMASK/DCC tracking and intercepting direct DMA buffer fills/invalidations, making clear tracking robust and accurate for titles beyond just Dead Cells.

PR #838, opened 2026-08-23
Fix TLS patcher scanning inside short jumps

foufouadi:
Before this change, the TLS patcher scanned executable memory byte by byte and attempted to match a TLS load from every position. It only validated the bytes after the candidate, without checking whether that position was actually the start of an instruction.

This could make the scanner start inside a previous instruction. In GTA V, a short jump with a 0x66 displacement appears immediately before a 66-prefixed FS:[0] load. The scanner interpreted the jump displacement as part of the TLS instruction and patched from the wrong byte, corrupting the jump and creating an infinite loop during the AGC resource-destructor path.

This change gives the TLS matcher enough context to detect that case and reject candidates immediately following a short-jump opcode. The scan then continues to the actual TLS instruction boundary and patches it normally..

### Testing

Tested for 75 seconds without diagnostic flags:

- GTA V: boot-time hang vs boots and presents the intro
- Astro Bot : no regression.
- Cult of the Lamb: no regression.
- Ghost of Yōtei: no regression.

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.cs
@@ -3200,7 +3200,7 @@ public sealed unsafe partial class DirectExecutionBackend : INativeCpuBackend, I
 				{
 					nint address = (nint)(ptr + i);
 					int remainingBytes = scanBytes - i;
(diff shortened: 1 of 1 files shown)


foufouadi:
https://github.com/user-attachments/assets/10390c6b-0956-417c-ab5c-c0a84205ca52

PR #840, opened 2026-08-24
shader: add compact f16 arithmetic and compare lowering

frangametv:
## Summary

This implements decoding and backend lowering for compact Gen5 F16 arithmetic and comparison instructions.

The arithmetic support covers:

- `VAddF16`
- `VSubF16`
- `VSubrevF16`
- `VMulF16`
- `VMaxF16`
- `VMinF16`

The change also implements the compact F16 VOPC comparison families, including the CMPX variants that update the execution mask.

Both SPIR-V and Metal backends are supported. Tests cover instruction decoding and generated output for both backends.

## Motivation

Guest shaders using these compact half-precision instructions previously failed translation or reached an unsupported opcode path. Supporting them allows those shaders to continue through the normal translation pipeline without introducing title-specific behavior.

Arithmetic and comparison support are included together because they share the same compact F16 source representation and backend conversion rules.

## Testing

Game testing: N/A — the isolated branch was validated through shader compiler tests and no game-specific compatibility claim is made.

- `SharpEmu.ShaderCompiler.Tests`: 95/95 passed.
- `SharpEmu.ShaderCompiler.Metal.Tests`: 60/60 passed.
- Both test projects built successfully in Release configuration.

The runtime Metal compiler path still needs confirmation from macOS CI.

## Credits

The F16 opcode work is based on research and implementation by Foued Attar (Foufouadi), particularly commit `46cc0aa`, and was adapted and completed for the current Vulkan and Metal backends.

## AI assistance

AI-assisted tools were used to help review, clean up, split, and verify the changes. The original implementation and research are credited above, and I manually reviewed and understood the submitted code.

## Checklist

- [x] I listed the game(s) I tested or marked the testing section as `N/A`.


foufouadi:
Great work, and thanks for the credit in the commit. I checked the VOP3 decoding/op_sel layout against the encoding, and the new test cases seemed correct.
I noticed a minor consistency issue: Float16Result and EmitFloat16Result only read OutputModifier and Clamp from Gen5Vop3Control, whereas the existing F32 path also handles Gen5SdwaControl.
This means that an SDWA F16 instruction using an output modifier or clamp would silently ignore these parameters in both backends. In my opinion, this isn't blocking for this PR, but it would be worth a follow-up to mirror FloatResult's control handling and add an SDWA test case.

PR #841, opened 2026-08-24
fix(agc): refine vertex layouts without rebasing captured inputs

frangametv:
## Summary

This makes vertex metadata refinement preserve the address and offset already resolved during vertex input capture.

Metadata is now applied only when its effective byte offset resolves to the same captured attribute. Once matched, the metadata may refine the format, stride, and instance rate, but it cannot move an existing binding to another address or offset.

The change also:

- Allows authoritative V# metadata to correct a stale inferred stride, including stride-40 interleaved layouts.
- Preserves the original binding order.
- Preserves distinct bindings when program counters are duplicated.
- Rejects metadata whose resolved offset conflicts with the captured input.
- Keeps `BaseAddress`, `Data`, `Location`, `Pc`, and `OffsetBytes` unchanged.

## Motivation

IR discovery and V# metadata can describe the same interleaved vertex stream using different base-address representations. The previous refinement logic could reject valid metadata because of a stale inferred stride, or move a correctly captured input while applying metadata.

This could produce incorrect vertex addresses, formats, and layouts during graphics pipeline creation.

The new matching rule treats the captured address and offset as authoritative while still allowing the descriptor to provide the correct layout information.

## Testing

Game testing: N/A — this isolated branch was verified with focused unit tests and does not claim an end-to-end title improvement by itself.

- `AgcVertexMetadataTests`: 8/8 passed.
- Tests cover stride-40 correction, relative captured bases, conflicting offsets, interleaved attributes, per-instance metadata, and binding-order preservation.

## Credits

Thanks to StealUrKill_PhD for identifying the stride-40 layout issue and the incorrect vertex-address behavior during metadata refinement.

## AI assistance

AI-assisted tools were used to help review, clean up, split, and verify the changes. The diagnostic findings are credited above, and I manually reviewed and understood the submitted code.

## Checklist

- [x] I listed the game(s) I tested or marked the testing section as `N/A`.

### DIFF
--- a/src/SharpEmu.Libs/Agc/AgcVertexMetadata.cs
+++ b/src/SharpEmu.Libs/Agc/AgcVertexMetadata.cs
@@ -238,10 +238,12 @@ internal static class AgcVertexMetadata
     }
 
     /// <summary>
-    /// Patch IR-discovered fetches from the attrib table onto the V# format/offset.
+    /// Patch IR-discovered fetches from the attrib table onto the V# layout.
     /// Prefer 1:1 Location pairing when counts match on one interleaved stream
-    /// (GTA UI glyphs). Otherwise match by stride + byte offset. Never rebases
-    /// BaseAddress/Data/Location/Pc/PerInstance.
+    /// (GTA UI glyphs). Otherwise match by the effective captured byte offset.
+    /// Never rebases BaseAddress/Data/Location/Pc or overwrites a discovered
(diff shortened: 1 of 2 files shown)


foufouadi:
Nice catch on the stride-40 issue, the fix looks correct. Ran a 6-game regression trace on the two dropped fallback matching cases zero hits, looks good.

PR #850, opened 2026-09-01
fix(macos): POSIX TBB workers, RGB10 present encode, DCC image publish

al0x99:
## Summary

macOS/Linux gaps that showed up while bringing Astro Bot up under Rosetta + MoltenVK. Four small commits; no ISA aliasing.

- **POSIX native guest workers.** The TBB pool was Windows-only (`CreateThread`). POSIX already had worker-event semaphores and Win64→SysV thunks; this wires `pthread_create`/`join`/`exit` and a warmed abort stub so Rosetta translates it before a signal resume. Without it, Astro `tbb_thead` jobs were refused (`tbb_native_worker unavailable`).
- **RGB10 present.** Astro flips `A2R10G10B10` linear buffers. A direct blit into an sRGB swapchain on MoltenVK produced scanlines; a numeric blit without encode crushed the title to near-black. Unpack to `RGBA16F`, then the existing linear-float encode path. Also pack VideoOut format tags so sampled RTs can alias GPU images.
- **DCC fast-clear** only called `RequestGuestColorClear`, which never publishes `_availableGuestImages`. Submit a real zero color-clear so later composites do not sample empty CPU tiles (`gpu_resident=False`). Texture skip now requires `IsGpuGuestImageAvailable` (the empty-pixel path ignored `SHARPEMU_NO_TEXTURE_SKIP`).
- **Shader emit** for `V_DOT2C_F32_F16` (unpack f16 pairs, dot, add dest) and saturating `V_CVT_PK_U16_U32` / `V_CVT_PK_I16_I32`. `SSetpcB64`/`SSwappcB64` are no-ops so title export shaders that call them still compile.

## Tested

Apple M5 Pro, `osx-x64` under Rosetta 2, MoltenVK 1.4.2, Astro Bot (`PPSA21564`) dump. Workers: `tbb_run_enter` thousands/min, abort stub recover + respawn. Present: `encode=True` on RGB10 flips; window capture went from RGB~7 (title bar only) to RGB~33 with ~50% visible pixels. Title screen still not correct (gray field) — remaining work is scanout contents, not this plumbing.

Not in this PR: unknown-ISA → NOP/ADD fallbacks (local only).

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.NativeWorker.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.NativeWorker.cs
@@ -180,7 +180,7 @@ public sealed partial class DirectExecutionBackend
 
 	private void PrewarmNativeGuestWorkers(int count)
 	{
-		if (!OperatingSystem.IsWindows() || NativeGuestWorkersDisabled || count <= 0)
+		if (NativeGuestWorkersDisabled || count <= 0)
 		{
 			return;
 		}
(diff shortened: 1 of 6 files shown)


al0x99:
Closing to split into smaller PRs that follow the mandatory template (CONTRIBUTING.md). Replacing with focused PRs.

PR #859, opened 2026-09-04
Fix: VEH spinlock, memory probing, and inflate_fast epilogue recovery

MathiasYG:
What this does

Fixes an infinite spinlock in the Vectored Exception Handler (VEH) trampoline caused by an x86-64 REX prefix encoding bug, hardens Windows host-memory probing against unmapped addresses, and recovers stream/state pointers needed by inflate_fast epilogue writebacks.

Details

### VEH trampoline spinlock

While debugging early execution under Direct Execution, the process would hang indefinitely on the first exception with one CPU core at 100%.

Thread-context sampling showed the thread stuck in the atomic compare-exchange loop emitted by the VEH trampoline:

`F0 4C 0F B1 11` → `lock cmpxchg qword ptr [rcx], r10`

In `CreateVectoredExceptionHandlerTrampoline` in `DirectExecutionBackend.cs`, the spinlock acquisition loop for `_vehManagedEntryLock` was using REX prefix `0x4C` (`REX.W=1, REX.R=1, REX.B=0`). With `REX.B=0`, ModR/M byte `0x11` addresses `rcx` rather than `r9`.

For a Windows x64 Vectored Exception Handler, `RCX` contains `PEXCEPTION_POINTERS`. The resulting `[rcx]` memory operand therefore referenced the exception data rather than `_vehManagedEntryLock`. Since that value was non-zero, the compare-against-zero operation could never succeed and the thread remained in the spin loop.

Changed the prefix from `0x4C` to `0x4D` (`REX.B=1`) in both `hostAcquireSpin` and `guestAcquireSpin`, correctly targeting `[r9]`.

### Windows host-memory probing

`TryReadHostQword` and `TryReadHostBytes` in `DirectExecutionBackend.Exceptions.cs` could attempt to read unmapped addresses without first checking whether the target page was committed.

On Windows, such probes could raise hardware access violations on managed threads and trigger runtime reverse-P/Invoke guards.

Added `VirtualQuery` checks on Windows to verify that the target pages are committed and have readable protection before attempting the memory read.

### inflate_fast epilogue recovery

In titles using statically linked zlib `inflate_fast`, certain loop exits can reach the epilogue with `rdi` (`strm`) and `rcx` (`strm->state`) unset when the final stream writebacks are performed. This caused access violations during container decompression at `0x804A2C23E` and `0x804A2C25A`.

Added `TryRecoverInflateFastEpilogue`, gated to Ghostrunner (`PPSA03685`), which:

- Recovers the valid `strm` pointer from the caller stack frame (`[saved_rbp - 0x58]`).
- Restores `strm->state` from `*(ulong*)(strm + 0x38)`.
- Corrects input-stream pointers affected by the bit-accumulator refund (`sub r10, rax`).
- Completes the expected stream writebacks so container decompression can return cleanly.

Testing

- Ghostrunner (PPSA03685) — Fixed the startup hang at import #3584 caused by the VEH trampoline spinlock. Confirmed pak container decompression completes for all 760 files in `pakchunk0-ps5.pak` without stream writeback crashes. Unreal Engine 4 subsequently initializes its TaskGraph and starts 21 worker threads.
- Dreaming Sarah (PPSA02929) — Confirmed clean boot past C2 initialization, audio loading, and active 4K/720p Vulkan presentation with no regressions.

Checklist

### DIFF
--- a/src/SharpEmu.Core/Cpu/Native/CpuPatcher.cs
+++ b/src/SharpEmu.Core/Cpu/Native/CpuPatcher.cs
@@ -43,5 +43,12 @@ public sealed unsafe class CpuPatcher : IDisposable
                 return -1;
             return *_current++;
         }
+
+        public int Position => (int)(_current - _start);
+
+        public void Skip(int count)
+        {
+            _current += Math.Min(Math.Max(count, 0), _length - Position);
+        }
     }
 }
--- a/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Exceptions.cs
+++ b/src/SharpEmu.Core/Cpu/Native/DirectExecutionBackend.Exceptions.cs
@@ -148,6 +148,11 @@ public sealed partial class DirectExecutionBackend
 			{
 				return -1;
 			}
+			if (exceptionCode == 3221225477u &&
+				TryRecoverInflateFastEpilogue(exceptionRecord, contextRecord, rip, rsp))
+			{
+				return -1;
+			}
 			if (exceptionCode == StatusIllegalInstruction &&
 				TryRecoverIllegalInstruction(contextRecord, rip))
 			{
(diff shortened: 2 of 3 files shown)


foufouadi:
Pushed the prerequisite onto your branch: TLS and stack-canary patterns are now matched at instruction boundaries instead of byte by byte.
Measured on GTA V , your tip crashes at ~5s, before the first frame, on mov dword [r14+r12*1+8], 0 with r14 = 0  a write through a register that was never loaded. With this commit on top, three consecutive runs present guest frames with no exception at all, and the title boots through to its menu meme .
Your REX fix is correct, by the way 0x4D is the right encoding for lock cmpxchg [r9], r10, and with 0x4C the r/m field stays rcx, so that spinlock has never guarded anything. It just needed this in front of it, because activating a lock that has never run changes VEH serialization emulator-wide.

foufouadi:
image
we get this now

PR #877, opened 2026-09-10
Fix stall watchdog killing legitimate large-asset-tree indexing

### LINKED ISSUE #876: Stall watchdog's fixed 20s threshold force-exits during legitimate large-asset-tree indexing
`GetStallWatchdogSeconds()` (`DirectExecutionBackend.cs:6412`) defaults to a fixed 20 seconds, and `Environment.Exit(4)` fires unconditionally the moment it elapses with no import progress (`DirectExecutionBackend.cs:6505`). This kills the process even when the "stalled" thread is doing legitimate, progressing work that just happens to take longer than 20s.

**Concrete case:** Astro Bot (PPSA21564) calls `sceAmprCommandBufferConstructor` (a NID marked `IsNoBlockLeafImport`) early in boot, which synchronously falls into `AmprFileRegistry.EnsureApp0Indexed` when the background preload (`BeginApp0IndexPreload`, kicked off from `SharpEmuRuntime.cs:406`) hasn't finished yet — the comment at `AmprFileRegistry.cs:197` already documents this as a known, deliberate fallback path. With the watchdog disabled (`SHARPEMU_STALL_WATCHDOG_SECONDS=0`), the walk completes and logs:

```
[LOADER][INFO] ampr.app0_indexed root=... files=156250 ids=624955 elapsed_ms=94864.3
```

**~95 seconds** for 156,250 files — a real, first-party title's asset tree, not a pathological case. With the default threshold, the process is killed at 20s, well before this legitimate work finishes, and execution never reaches the point where the walk would complete on its own.

This isn't Astro-Bot-specific: any title with a large enough `data/` tree hits the same wall on the very first boot (subsequent boots hit the on-disk index cache instead — `ampr.app0_index_cache_hit` — and would be fast).

**Suggested direction:** rather than raising the constant (which just moves the cliff edge for the next larger title), have long-running-but-progressing HLE work like this indexing walk periodically signal into the same progress mechanism the watchdog already resets on (`MarkExecutionProgress()` on `DirectExecutionBackend`) so it isn't mistaken for a genuine stall. That mechanism currently lives in `SharpEmu.Core`, and the indexer lives in `SharpEmu.Libs`, so this needs a small public seam between them rather than a quick cross-layer call — flagging for discussion on the right shape of that seam before sending a PR.

theantipopau:
## Before submitting

## What changed

Fixes #876. `sceAmprCommandBufferConstructor` is marked `IsNoBlockLeafImport`, but its first-boot fallback (`AmprFileRegistry.EnsureApp0Indexed`, reached synchronously from the guest export when the background preload hasn't caught up yet) can run well past the native backend's 20s stall window on a title with a large `app0` tree. The backend then calls `Environment.Exit(4)` despite the guest thread being blocked on real, progressing work — not actually stuck.

Adds `GuestThreadExecution.HostWorkProgressObserved` — a narrow event on the existing shared seam between `SharpEmu.Core` and `SharpEmu.Libs` (the same pattern already used in the other direction via `GuestThreadAbandoned`). `DirectExecutionBackend` subscribes it to the same `MarkExecutionProgress()` the watchdog's timer already resets on. `AmprFileRegistry`'s three long-running loops (file enumeration, parallel hashing, cache save) each call it every 2000 items.

All three loops needed it, not just the obvious one — I only found that by tracing instance identity and notification counts against the watchdog's own elapsed-time log during debugging: the file-enumeration and parallel-hashing loops alone got the walk to completion, but the *cache-save* step (re-deriving relative paths and re-hashing every id, unsignaled) was long enough on its own to still trip the watchdog immediately afterward.

## Why

Real game testing, not synthetic: Astro Bot (PPSA21564, legally dumped off my own PS5) hits this every single boot — its `data/` tree is large enough that first-time indexing takes 80–95s. Any title with a comparably large asset tree would hit the same wall.

## Testing

Real game: Astro Bot (PPSA21564).

**Before:** default settings, every run force-exits at exactly 20s (`Execution stalled with no import progress for 20s (imports=528384)`), regardless of how long the process is left running — confirmed with `Environment.Exit(4)` firing exactly at the threshold, not a hang.

**After**, default settings (no env var override), fresh index cache:

```
[LOADER][INFO] ampr.app0_index_cache_saved path=...\ampr-index\app0-2b5108d1.v3.idx files=156250
[LOADER][INFO] ampr.app0_indexed root=Z:\...\PPSA21564-app0 files=156250 ids=624955 elapsed_ms=84888.0
...
GAME: Resident Load end 0.15s
```

The full 85s indexing walk completes without the watchdog firing, and execution reaches real game code (the game's own resident-load logging). It subsequently hits an unrelated, already-reported crash further into GPU/shader work (#11) — outside the scope of this fix.

`dotnet build` and `dotnet test` clean, 1045/1045 existing tests pass, no regressions.

## AI assistance disclosure

This PR was prepared with Claude (Anthropic) assistance under my direction: it diagnosed the root cause (including the two-stage debugging process described above — instance-identity tracing and notification-count analysis to find the unsignaled cache-save step), designed and implemented the fix, and validated it against the real game repro described. I reviewed the design (why this seam, why this location, the tradeoffs of a fire-and-forget event vs. a bigger interface change), the implementation, and the test evidence before submitting, and can answer questions about it.

## Checklist

- [ ] I wrote this pull request description myself and did not paste AI-generated explanations. — see AI assistance disclosure above.


theantipopau:
Fresh, quantified before/after on Astro Bot (#11) using this exact branch, cold-start (cleared AMPR index cache) 3/3 both ways: without this fix, deterministic force-exit at exactly 528384 imports every run; with it, all 3 runs complete the ~83s indexing walk cleanly and proceed through AGC init, shader creation, and GPU pipeline setup. Full details on #11: https://github.com/sharpemu/sharpemu/issues/11#issuecomment-5649191429

